PR #392, opened 2020-09-01
fix(checkout): INT-3116 Hide Bolt on checkout

FranciscoOnt:
## What?
Remove Bolt form payment methods if showInChekout is false , to prevent it from being rendered as payment method.

## Why?
To prevent Bolt's payment method from being rendered, but still usable from the Bolt app and manual orders.

## Sibling PRs
https://github.com/bigcommerce/bigcommerce/pull/36855

## Testing / Proof
Demo video recorded using this Draft PR as base https://github.com/bigcommerce/bigcommerce/pull/36925, actual behavior depends on the sibling PR
[Demo Video](https://drive.google.com/file/d/1u5Y3AxqKN32uYfx-cflInQjLgd23iCAB/view?usp=sharing)

@bigcommerce/checkout

### DIFF
--- a/src/app/payment/Payment.tsx
+++ b/src/app/payment/Payment.tsx
@@ -1,6 +1,6 @@
 import { CheckoutSelectors, CheckoutSettings, OrderRequestBody, PaymentMethod } from '@bigcommerce/checkout-sdk';
 import { memoizeOne } from '@bigcommerce/memoize';
-import { compact, find, isEmpty, noop } from 'lodash';
+import { compact, find, isEmpty, noop, remove } from 'lodash';
 import React, { Component, ReactNode } from 'react';
 import { ObjectSchema } from 'yup';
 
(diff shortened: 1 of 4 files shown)


src/app/payment/Payment.tsx
```diff
@@ -462,6 +462,16 @@ export function mapToPaymentProps({
 
     let selectedPaymentMethod;
     let filteredMethods;
+
+    // Prevent a payment method from being rendered
+    remove(methods, method => {
```
capsula4, line 467: I wonder if instead of using remove is better to use `filteredMethods` and use filter instead? Not sure if modifying the original value can cause issues.

FranciscoOnt:
Having issues with the fork being private and the PR no longer updating, will reopen in the new public fork https://github.com/bigcommerce/checkout-js/pull/410

PR #414, opened 2020-09-17
CHECKOUT-5173 Fix server commands

capsula4:
## What?
Fix server commands

## Why?
`npm run dev` creates `build` folder,  so `npm run dev:server` should serve `build` folder.
Also provides server command for `npm run build:server`

## Testing / Proof
manual

@bigcommerce/checkout


davidchin (approved):
Nice one 👍

PR #432, opened 2020-10-08
fix(payment): INT-3308 added style to avoid overlapping cards fields with Postcode label on it on stripeV3

BC-AxelMalagon:
## What? [INT-3308](https://jira.bigcommerce.com/browse/INT-3308)
Added style to avoid overlapping Cards fields with Postcode label on it on stripeV3

## Why?
In order to avoid overlapping fields when Postcode label is showed on StripeV3.

## Testing / Proof
[Testing Proof video INT-2902](https://drive.google.com/file/d/1QylTdxD2BPOTW-0bpcjVI6silm1R919e/view?usp=sharing)

Screen Shot 2020-10-08 at 6 35 26 PM

### DIFF
--- a/src/scss/components/checkout/widget/_widget.scss
+++ b/src/scss/components/checkout/widget/_widget.scss
@@ -72,15 +72,21 @@
     margin-top: remCalc(20px);
     padding: 1rem;
 
-    @media screen and (max-width: 375px) {
-        margin: remCalc(20px) -2rem remCalc(20px) -2rem;
-        padding: 1rem;
+    @media screen and (max-width: $screen-large-iPhone) {
+        margin: remCalc(20px) 0 remCalc(20px) 0;
+        padding: 1rem 0 1rem 1rem;
         width: 20rem;
     }
 
-    @media screen and (max-width: 320px) {
+    @media screen and (max-width: $screen-medium-iPhone) {
+        margin: remCalc(20px) -2rem remCalc(20px) -2rem;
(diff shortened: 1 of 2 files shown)


cmedinilla (approved):
LGTM!

src/scss/components/checkout/widget/_widget.scss
```diff
@@ -72,15 +72,21 @@
     margin-top: remCalc(20px);
     padding: 1rem;
 
+    @media screen and (max-width: 414px) {
```
ankurbigcomm, line 75: This seems like a magic number. Should be looking at using a more standard screen/viewport size? 

Thoughts?

BC-AxelMalagon: This size was taken from iphone 6/7/8 plus. It was added because between  375px and 414px, expiration date and card fields were still very close and overlapping.

capsula4: is there any possibility of creating a variable for this magic number? so this way it's self documented

BC-AxelMalagon: Yes. It is already added. Thanks a lot.

capsula4 (approved):

BC-AxelMalagon:
hi @capsula4, i made some changes in order to follow the same structure, and avoid having magic number in the others Media queries. Could you review them again, please?

capsula4 (approved):

PR #433, opened 2020-10-12
fix(checkout): CHECKOUT-5201 fix safari date range issue

ankurbigcomm:
## What?
Update date field in safari to use ISO8601 date format for max and min input

## Why?
Safari is unable to call new Date with other date formats apart from ISO8601. As a result the date range functionality was broken in safari. 

ISO 8601 is supported for most browsers and they are able to convert the date correctly.

## Testing / Proof
Screen Shot 2020-10-12 at 11 40 43 am

@bigcommerce/checkout


capsula4 (approved):
👏  great finding

PR #440, opened 2020-10-20
fix(checkout): BCTHEME-287 Add padding between wallet buttons

bdav87:
## What?
Adding padding between wallet buttons that appear on the first step of checkout.

## Why?
There's no space between the buttons currently.

## Testing / Proof
Before:
image

After:
image

@bigcommerce/checkout

### DIFF
--- a/src/scss/components/checkout/checkoutRemote/_checkoutRemote.scss
+++ b/src/scss/components/checkout/checkoutRemote/_checkoutRemote.scss
@@ -7,6 +7,10 @@
     flex-wrap: wrap;
     margin-left: -(spacing("third"));
     margin-right: -(spacing("third"));
+
+    > div {
+        margin-right: spacing("single");
+    }
 }
 
 .checkoutRemote-button {


src/scss/components/checkout/checkoutRemote/_checkoutRemote.scss
```diff
@@ -7,6 +7,10 @@
     flex-wrap: wrap;
     margin-left: -(spacing("third"));
     margin-right: -(spacing("third"));
+
+    > div {
+        padding-right: spacing("single");
```
capsula4, line 12: I think `margin-right` would be more appropriate for this.

bdav87: Updated to use margin instead. Confirmed button spacing appears the same in VM.

capsula4 (approved):

capsula4:
Thanks for this!

PR #443, opened 2020-10-20
CHECKOUT-5263 Default to empty array if no payments are set

capsula4:
## What?
Default to empty array if no payments are set

## Why?
Because some merchants are testing out installing a custom checkout without having any payment set up and it fails:
Screen Shot 2020-10-21 at 9 47 46 am

## Testing / Proof
manual + unit

@bigcommerce/checkout

### DIFF
--- a/src/app/customer/CheckoutButtonList.tsx
+++ b/src/app/customer/CheckoutButtonList.tsx
@@ -20,7 +20,7 @@ export const SUPPORTED_METHODS: string[] = [
 ];
 
 export interface CheckoutButtonListProps {
-    methodIds: string[];
+    methodIds?: string[];
     checkEmbeddedSupport?(methodIds: string[]): void;
     deinitialize(options: CustomerRequestOptions): void;
     initialize(options: CustomerInitializeOptions): void;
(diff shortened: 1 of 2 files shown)


ankurbigcomm (approved):
Thanks for addressing this. 🙇‍♂️

src/app/customer/CheckoutButtonList.tsx
```diff
@@ -33,7 +33,7 @@ const CheckoutButtonList: FunctionComponent<CheckoutButtonListProps> = ({
     methodIds,
     ...rest
 }) => {
-    const supportedMethodIds = methodIds
+    const supportedMethodIds = (methodIds || [])
```
icatalina, line 36: Might be better defined in the deconstruct:

```
methodIds = [],
```

or at least use the  nullish coalescing operator:

```suggestion
    const supportedMethodIds = (methodIds ?? [])
```

icatalina (approved):
Nice 👍

src/app/customer/CheckoutButtonList.tsx
capsula4: I think the constructor wouldn't work if `null` is explicitly passed?

icatalina: True, I'm not sure `null` is allowed as a replacement of `undefined` though.

capsula4: that's true, but we usually define types as `supportedMethodIds?: string[]` which is the same as `supportedMethodIds: string[] | undefined`. However, this value comes from the API and it can actually be `null`. For simplicty, we haven't really explicitly modelled `null` in the types (e.g. `supportedMethodIds?: string[] | null`).

icatalina: That can be an issue though, if we think the API might return `null` we might be better off typing it explicitly, otherwise someone can make the wrong assumption.

PR #447, opened 2020-11-04
CHECKOUT-5295 Hide AmazonPay when multi-shipping

capsula4:
## What?
Filter out `amazon` and `amazonpay` from payment methods when there are multiple consignments.

## Why?
Because those payment providers only support one single consignment.

## Testing / Proof
unit

Screen Shot 2020-11-04 at 5 40 31 pm
Screen Shot 2020-11-04 at 5 40 46 pm

@bigcommerce/checkout

### DIFF
--- a/src/app/payment/Payment.spec.tsx
+++ b/src/app/payment/Payment.spec.tsx
@@ -12,6 +12,7 @@ import { getStoreConfig } from '../config/config.mock';
 import { getCustomer } from '../customer/customers.mock';
 import { createLocaleContext, LocaleContext, LocaleContextType } from '../locale';
 import { getOrder } from '../order/orders.mock';
+import { getConsignment } from '../shipping/consignment.mock';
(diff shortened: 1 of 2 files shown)


PascalZajac:
I don't think we should be hard-coding logic to specific provider IDs here. If Amazon doesn't support multiple-address shipping, that should be some sort of capability that the provider implementation declares and Checkout reads from Payments.

capsula4:
@PascalZajac  I agree, we should be including a boolean at the API level, just thought this would be faster to solve the CSMED.

ankurbigcomm (approved):
Changes look good to me but as per @PascalZajac comment we should in the future create a more API centric solution.

Should we create a separate ticket to may be track that work.

capsula4:
Yeah I'll create a separate ticket and start the work straight away as per @PascalZajac suggestion: CHECKOUT-5336

PR #450, opened 2020-11-09
CHECKOUT-5346 Hide multishipping if not enabled

capsula4:
## What?
Hide multi-shipping UI if it is not enabled. If there are multiple consignments, then an empty shipping form will be shown.

## Why?
Checkout UI should honour the Checkout Settings. If it's not enabled, it shouldn't render the multi-shipping UI.
In some cases, some shoppers have been presented with a multi-shipping UI despite the setting being disabled.

Still not sure how they ended up with more than one shipping address in the first place, but this at least will prevent the shopper paying for shipping twice.

## Testing / Proof
Screen Shot 2020-11-09 at 3 24 49 pm

@bigcommerce/checkout

### DIFF
--- a/src/app/checkout/mapToCheckoutProps.ts
+++ b/src/app/checkout/mapToCheckoutProps.ts
@@ -14,7 +14,10 @@ export default function mapToCheckoutProps(
     const { promotions = EMPTY_ARRAY } = data.getCheckout() || {};
     const submitOrderError = errors.getSubmitOrderError() as CustomError;
     const {
-        checkoutSettings: { guestCheckoutEnabled: isGuestEnabled = false } = {},
+        checkoutSettings: {
+            guestCheckoutEnabled: isGuestEnabled = false,
+            hasMultiShippingEnabled = false,
+        } = {},
         links: { loginLink: loginUrl = '' } = {},
     } = data.getConfig() || {};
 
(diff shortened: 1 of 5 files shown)


src/app/shipping/Shipping.tsx
```diff
@@ -318,6 +318,8 @@ export function mapToShippingProps({
         countriesWithAutocomplete.push('GB');
     }
 
+    const shippingAddress = !shouldShowMultiShipping && consignments.length > 1 ? undefined : getShippingAddress();
```
davidchin, line 321: Just curious, what's the reasoning behind showing an empty address form? Is it to force the shopper to re-enter their shipping details? What would happen to the second consignment (and others) once they do so? Would they still stick around but not visible to the shopper?

src/app/checkout/mapToCheckoutProps.ts
```diff
@@ -14,7 +14,10 @@ export default function mapToCheckoutProps(
     const { promotions = EMPTY_ARRAY } = data.getCheckout() || {};
     const submitOrderError = errors.getSubmitOrderError() as CustomError;
     const {
-        checkoutSettings: { guestCheckoutEnabled: isGuestEnabled = false } = {},
+        checkoutSettings: {
+            guestCheckoutEnabled: isGuestEnabled = false,
+            hasMultiShippingEnabled = false,
```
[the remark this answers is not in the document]
capsula4: That's correct, it only controls the UI.

src/app/shipping/Shipping.tsx
capsula4: The empty address is to force the shopper to enter one. Since there are two consignments and both could have valid addresses, the shipping step cannot pick one, so instead it asks the shopper to re enter the information.

davidchin: Ok, thank you, that's good - they get combined and the shopper can resume their checkout in an expected state once they re-enter their address details 👍

davidchin (approved):
All good, LGTM 👍

PR #451, opened 2020-11-09
CHECKOUT-5327 Validate checkout safe values 

capsula4:
## What?
Validate that provided values are considered "safe", that means, does not contain any of the following:
```
< >
```

## Why?
To avoid fraudulent/suspicious activity.

## Testing / Proof
Experiment ON:
Screen Shot 2020-11-09 at 5 14 13 pm

Experiment OFF:
Screen Shot 2020-11-09 at 5 15 02 pm

@bigcommerce/checkout


davidchin (approved):
The UI changes LGTM 👍

PR #466, opened 2020-11-23
fix(common): CHECKOUT-5332 Only send in-app errors to Sentry

davidchin:
## What?
Only send errors that originated from application code and its dependencies.

## Why?
We're not responsible for errors that are caused by code that is external to the application.

At the moment, we are seeing some errors in our log that are caused by third party scripts, such as [Klaviyo](https://sentry.io/organizations/bigcommerce/issues/1904396678) and [Poptin](https://sentry.io/organizations/bigcommerce/issues/2050108407). We realised that `whitelistUrls` and `blacklistUrls` config provided by Sentry client only checks the topmost frame in the stack trace. This means that it won't filter out those exception events if they are caused by problems deeper in the stack trace. i.e.:

Screen Shot 2020-11-23 at 1 47 13 pm

We can't filter them using Sentry's server-side config because it can only filter out exceptions by error messages, not file path. File paths can only be used for grouping of events. So the filtering has to be done on the client side.

### Before
There's a call to Sentry, even though the error is caused by an external script (https://127.0.0.1:8080/index.js):

Screen Shot 2020-11-23 at 6 03 27 pm

### After
There's no call to Sentry:

Screen Shot 2020-11-23 at 6 06 37 pm

Exceptions thrown by the application still get logged:

Screen Shot 2020-11-23 at 6 06 37 pm

@bigcommerce/checkout

### DIFF
--- a/src/app/common/error/SentryErrorLogger.spec.ts
+++ b/src/app/common/error/SentryErrorLogger.spec.ts
@@ -3,7 +3,6 @@ import { RewriteFrames } from '@sentry/integrations';
 import { Integration } from '@sentry/types';
 
 import computeErrorCode from './computeErrorCode';
-import DEFAULT_ERROR_TYPES from './defaultErrorTypes';
 import ConsoleErrorLogger from './ConsoleErrorLogger';
 import { ErrorLevelType } from './ErrorLogger';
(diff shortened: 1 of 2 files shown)


capsula4 (approved):
:clap: awesome stuff :shipit:

PR #463, opened 2020-11-18
feat(checkout): INT-2544 Create checkout.com APM  custom payment forms

FranciscoOnt:
## What? [INT-2544](https://jira.bigcommerce.com/browse/INT-2544)
Create a custom payment form for checkout.com's APMs 

## Why?
Certain Checkout.com APMs require a field called document in order for the payment to proceed.

## Testing / Proof
This PR has only been tested on development environment to be shown as expected, and on integrations to keep checkout.com working as is until the new functionality is developed and tested
Screenshot taken in Development environment
Screen Shot 2021-01-12 at 2 48 47 PM

bigcommerce/checkout

PR #467, opened 2020-11-23
feat(payments): INT-2546 Handle checkout.com multi-options redirection

cmedinilla:
## What?
Add checkout.com in the redirect whitelist 

## Why?
We are changing checkout.com to be a multi-options instead a cc only paymet provider

## Testing / Proof
WIP

@bigcommerce/checkout


cmedinilla:
Added in https://github.com/bigcommerce/checkout-js/pull/463

PR #469, opened 2020-11-27
Create account

capsula4:
## What?
Allow create customer account inside checkout flow

## Why?
To minimize friction. Instead of opening the regular account signup form, presents the minimum required fields in checkout.

## Testing / Proof
Screen Shot 2020-12-09 at 12 32 28 pm

Screen Shot 2020-12-09 at 12 33 43 pm

@bigcommerce/checkout


src/app/customer/Customer.tsx
```diff
@@ -40,13 +47,15 @@ export interface WithCheckoutCustomerProps {
...
     signInEmailError?: Error;
+    createAccountError?: Error;
     signInError?: Error;
     clearError(error: Error): Promise<CheckoutSelectors>;
     continueAsGuest(credentials: GuestCredentials): Promise<CheckoutSelectors>;
     deinitializeCustomer(options: CustomerRequestOptions): Promise<CheckoutSelectors>;
     initializeCustomer(options: CustomerInitializeOptions): Promise<CheckoutSelectors>;
     sendLoginEmail(params: { email: string }): Promise<CheckoutSelectors>;
     signIn(credentials: CustomerCredentials): Promise<CheckoutSelectors>;
+    createAccount(values: any): Promise<CheckoutSelectors>;
```
davidchin, line 58: Can you provide a type for `values` instead of `any`?

src/app/customer/Customer.tsx
```diff
@@ -82,7 +91,9 @@ class Customer extends Component<CustomerProps & WithCheckoutCustomerProps, Cust
                 { isEmailLoginFormOpen && this.renderEmailLoginFormForm() }
                 { (viewType === CustomerViewType.Guest) ?
                     this.renderGuestForm() :
-                    this.renderLoginForm() }
+                    viewType === CustomerViewType.CreateAccount ?
```
davidchin, line 94: Can we pull this `viewType` check out into a private method or something? It feels odd to have two ternary conditions chained together.

src/app/customer/Customer.tsx
```diff
@@ -24,14 +28,17 @@ export interface CustomerProps {
 }
 
 export interface WithCheckoutCustomerProps {
+    canCreateAccountOnCheckout: boolean;
```
davidchin, line 31:
```suggestion
    canCreateAccountInCheckout: boolean;
```

src/app/customer/Customer.tsx
```diff
@@ -159,9 +170,35 @@ class Customer extends Component<CustomerProps & WithCheckoutCustomerProps, Cust
...
+            customerAccountFields,
+            isCreatingAccount,
+            createAccountError,
+            canCreateAccountOnCheckout,
+            createAccountLink,
+            isEmbedded,
+        } = this.props;
+
+        if (isEmbedded || !canCreateAccountOnCheckout) {
+            window.top.location.assign(createAccountLink);
```
davidchin, line 184: Do you want to render a link which redirects the shopper to the sign up page when it is clicked? Or simply exit early and not render anything. At the moment, it seems you're directly doing a redirect in a `render` method.

src/app/customer/getCreateCustomerValidationSchema.ts
```diff
@@ -0,0 +1,70 @@
...
+}
+
+function getTranslateCreateCustomerError(language?: LanguageService): TranslateValidationErrorFunction {
+    return (type, { label, min, max }) => {
+        if (!language) {
+            return;
+        }
+
+        if (type === 'required') {
+            return language.translate(`customer.required_error`, { label });
```
davidchin, line 31: I think you can just use `''` instead of back ticks because you're not interpolating anything. 🍹

src/app/customer/getCreateCustomerValidationSchema.ts
```diff
@@ -0,0 +1,70 @@
...
+}: CreateCustomerValidationSchema): ObjectSchema<CreateAccountFormValues> {
+    return object({
+            firstName: string().required(language.translate('address.first_name_required_error')),
+            lastName: string().required(language.translate('address.last_name_required_error')),
+            password: string()
+                .required(description || language.translate('customer.password_required_error'))
+                .matches(numeric, description || language.translate('customer.password_number_required_error'))
+                .matches(alpha, description || language.translate('customer.password_letter_required_error'))
+                .min(minLength, description || language.translate('customer.password_under_minimum_length_error'))
+                .max(100, language.translate('customer.password_over_maximum_length_error')),
```
davidchin, line 63: How come it is just a hardcoded number (`100`)? Does `passwordRequirements` contain the max length as well or not really?

src/app/customer/getPasswordRequirements.ts
```diff
@@ -0,0 +1,42 @@
...
+    description,
+    alpha,
+    numeric,
+}: CustomerPasswordRequirements): PasswordRequirements {
+    const allSlashes = new RegExp('/', 'g');
+
+    return {
+        minLength: minlength,
+        description,
+        alpha: new RegExp(alpha.replace(allSlashes, '')),
```
davidchin, line 39: Do you just want to remove the start and end of `/`, or all occurrences of `/`?

src/app/ui/form/CheckboxGroupFormField.tsx
```diff
@@ -83,11 +83,11 @@ const MultiCheckboxFormField: FunctionComponent<MultiCheckboxFormFieldProps> = (
     return <FormFieldContainer hasError={ getIn(errors, name) && getIn(errors, name).length }>
         { label }
 
-        <MultiCheckboxControl
+        { options.length > 1 && <MultiCheckboxControl
```
davidchin, line 86: What do we display if `options.length` is `1`?

src/app/customer/Customer.tsx
```diff
@@ -329,26 +399,37 @@ export function mapToWithCheckoutCustomerProps(
...
+        canCreateAccountOnCheckout: features['CHECKOUT-4941.account_creation_in_checkout'],
+        customerAccountFields: getCustomerAccountFields(),
+        createAccountLink,
         canSubscribe: config.shopperConfig.showNewsletterSignup,
         checkoutButtonIds: config.checkoutSettings.remoteCheckoutProviders,
         clearError: checkoutService.clearError,
+        createAccount: checkoutService.createCustomerAccount,
         continueAsGuest: checkoutService.continueAsGuest,
         // todo: remove casting when method is properly exposed.
         sendLoginEmail: (checkoutService as any).sendSignInEmail as (params: { email: string }) => Promise<CheckoutSelectors>,
```
davidchin, line 424: While you have the file opened, can you clean up this comment?

src/app/customer/Customer.tsx
```diff
@@ -329,26 +399,37 @@ export function mapToWithCheckoutCustomerProps(
...
+        checkoutSettings: {
+            privacyPolicyUrl,
+            requiresMarketingConsent,
+            isSignInEmailEnabled,
+            features,
+        },
+        links: {
+            createAccountLink,
+        },
+    } = config as any;
```
davidchin, line 412: Do we still need the casting to `any`?

davidchin (commented):
Looking good! 🙂

src/app/customer/getCreateCustomerValidationSchema.ts
capsula4: I guess this is just code that we have been moving around for a while. I copied this password validation from the current one, which I guess it comes from ng-checkout. The API does not set any max length.

src/app/ui/form/CheckboxGroupFormField.tsx
capsula4: We display a control that allows you select all or none with one click `Select All | None`
Doesn't make sense for a single checkbox.

src/app/customer/getPasswordRequirements.ts
capsula4: I think so. I think this is another copy-pasted code from ng-checkout https://github.com/bigcommerce-labs/ng-checkout/blob/master/src/app/directives/customer-password/customer-password.controller.js#L49

src/app/customer/Customer.tsx
capsula4: I'm after the latter. If account creation in the checkout is not supported, the user has already clicked a link. This is for keeping the logic centralized in one place, rather than updating "create account" links to behave differently depending on an experiment.

davidchin: Latter as in "not render anything" or you want to actually do a redirect here. If it's "not render anything", just do `return`. If it's "do a redirect here", I recommend you doing it elsewhere because it's good practice for `render` to not result in any side effect as the method can be called multiple times by React.

capsula4:
@davidchin this is ready for a final review now that the SDK has been properly bumped with the new functionality

src/app/customer/getCreateCustomerValidationSchema.ts
davidchin: Ok cool, that's fine if it's existing code

src/app/customer/CreateAccountForm.tsx
```diff
@@ -0,0 +1,119 @@
...
+    formFields,
+    createAccountError,
+    isCreatingAccount,
+    onCancel,
+    shouldRender,
+    createAccountLink,
+}) => {
+
+    if (!shouldRender) {
+        window.top.location.assign(createAccountLink);
```
davidchin, line 36: We should avoid performing any side effect in a `render` function. Doing a redirect is considered to be a side effect as it causes a direct change external to the function itself. You can use `useEffect` to perform a side effect in a `render` function but I wouldn't recommend in this case because you'll end up writing something like "do a redirect every time `shouldRender` is changed to `true` or the component is mounted". A better place to perform a redirect is inside an event handler or whatever callback that's triggered when the user clicks on "create account" link (e.g.: `onCreateAccount`). In there, you can determine whether to show a form or do a render based on the config value.

capsula4: Sorry I thought this was different to doing a redirect inside a render function because this is an FC, but I get your point, will readdress.

capsula4: @davidchin pushed some changes earlier,  hopefully I got it right this time

davidchin (approved):
Cool, thank you LGTM! 👍

PR #471, opened 2020-11-30
refactor(common): CHECKOUT-4941 Create FormField module

capsula4:
## What?
Creates FormField module and moves form field related logic to it.

## Why?
Because Account Creation will make use of it. Currently it's coupled with the Address module and they should be independent.

## Testing / Proof
unit 

https://app.circleci.com/jobs/github/bigcommerce/bigcommerce/392272

Screen Shot 2020-11-30 at 1 59 09 pm

@bigcommerce/checkout


src/app/formFields/DynamicFormField.tsx
```diff
@@ -1,86 +1,57 @@
...
 
 import { TranslatedString } from '../locale';
 import { FormField, Label } from '../ui/form';
 
-import { getFormFieldInputId, getFormFieldLegacyName } from './getFormFieldInputId';
 import CheckboxGroupFormField from './CheckboxGroupFormField';
 import DynamicFormFieldType from './DynamicFormFieldType';
 import DynamicInput from './DynamicInput';
 
-export interface AddressKeyMap<T = string> {
```
davidchin, line 13: I suggest that instead of having a top-level `formFields` module, maybe a better place to place the `DynamicFormField` component is `ui/form`. Otherwise, we'll end up with two quite similar modules as both contain generic form components. It can cause confusion in the future in terms of where to put things.

src/app/formFields/getAddressCustomFieldsValidationSchema.spec.tsx
```diff
@@ -1,14 +1,13 @@
...
 
-import { getFormFields } from './formField.mock';
-import getAddressCustomFieldsValidationSchema from './getAddressCustomFieldsValidationSchema';
-import getAddressValidationSchema from './getAddressValidationSchema';
-import { AddressFormValues } from './mapAddressToFormValues';
+import getCustomFormFieldsValidationSchema from './getCustomFormFieldsValidationSchema';
+import getFormFieldsValidationSchema, { FormFieldValues } from './getFormFieldsValidationSchema';
 
-describe('getAddressCustomFieldsValidationSchema', () => {
+describe('getCustomFormFieldsValidationSchema', () => {
```
davidchin, line 10: I think the name of the spec file is still called `getAddressCustomFieldsValidationSchema`. Should probably rename that

src/app/formFields/getAddressValidationSchema.spec.tsx
```diff
@@ -1,13 +1,12 @@
...
+import { getFormFields } from '../address/formField.mock';
 import { getShippingAddress } from '../shipping/shipping-addresses.mock';
 
-import { getFormFields } from './formField.mock';
-import getAddressValidationSchema from './getAddressValidationSchema';
-import { AddressFormValues } from './mapAddressToFormValues';
+import { default as getFormFieldsValidationSchema, FormFieldValues } from './getFormFieldsValidationSchema';
 
-describe('getAddressValidationSchema', () => {
+describe('getFormFielsValidationSchema', () => {
```
davidchin, line 9: Same with this spec file. It is still called `getAddressValidationSchema.spec.tsx`.

src/app/formFields/getCustomFormFieldsValidationSchema.ts
```diff
@@ -29,7 +28,7 @@ export default memoize(function getAddressCustomFieldsValidationSchema({
             formFields
                 .filter(({ custom }) => !!custom)
                 .reduce((schema, { name, label, required, fieldType, type, min, max }) => {
-                    const requiredErrorMessage = translate(`${ERROR_KEYS.custom}_required_error`, { label });
+                    const requiredErrorMessage = translate('address.custom_required_error', { label });
```
davidchin, line 31: The translation keys are still tied to the address namespace. If we want this to be a generic function, we should consider lifting the translation keys up and passing them into the function instead. For example, `SingleShippingForm` (where it is used at the moment) can pass in a map of translation keys specific to addresses. Alternatively, you can pass in a decorated `translate` function that returns address-specific error messages.

src/app/formFields/getFormFieldsValidationSchema.ts
```diff
@@ -20,11 +19,15 @@ const ERROR_KEYS: { [fieldName: string]: string } = {
     phone: 'address.phone_number',
 };
 
-export default memoize(function getAddressValidationSchema({
+export interface FormFieldValues {
+    [key: string]: string | { [id: string]: any };
+}
+
+export default memoize(function getFormFieldsValidationSchema({
```
davidchin, line 26: I think it is the same situation here. The translation keys are still tied to address fields (see `ERROR_KEYS`). We need to find a way to pull them out so they are not statically coupled with this function.

src/app/formFields/getCustomFormFieldsValidationSchema.ts
capsula4: I understand that currently, the translation key is `address.custom_required_error` however,I don't it's desirable to customize "custom form field required" in different ways for the address form and the account creation form. I opted to leave it in the address namespace, instead of moving to something more generic like `form.custom_required_error` as that would break existing translations. Thoughts?

src/app/formFields/getFormFieldsValidationSchema.ts
capsula4: similar to what I described above, some of the form fields are shared like `first_name` and `last_name`, so don't think it's worth moving the translations key a level above. I think for now it's okay to leave the mapping in this class and if it gets more complex in the future we could always refactor it.

src/app/formFields/getCustomFormFieldsValidationSchema.ts
davidchin: I actually thought about suggesting to take the simple route of just renaming the namespace but it'd definitely break existing translations. That's why I suggested to just lift the keys out of the function and pass them in instead. I made the suggestion because this component is no longer for a custom field for addresses. Instead it is for a custom field that can be applied to various contexts (address, account creation etc...), even if ultimately they end up with the same text. Therefore I believe it's best if `address` reference is removed from it.

Separately, we have to consider if we need to define a new key, e.g.: `customer.custom_required_error`, thereby giving giving merchants the ability to specify a different label for account creation. I agree it's probably not _strictly_ necessary because most likely they end up with the same text. But on the other hand, it does make it much clearer that the label belongs to a different section and describes a different thing.

capsula4: After thinking about how moving the keys up would end up looking like, I'm not convinced in would be a very clean design, as I would have to expose address-specific keys in the signature of `getFormFieldsValidationSchema`:

```js
// passed translation keys
{
  firstNameRequired,
  lastNameRequired,
  companyRequired,
  // rest of address specific fields
  // ...,
  customRequired,
  customMin
  customMax
  invalidCharacters
}
```

or even something a bit more complex but generic like 2 new params, `keys` and `overrides`
```js
// base translation keys
{
  required,
  min
  max
  invalidCharacters
}
// override keys
{
 firstNameRequired,
  lastNameRequired,
  companyRequired,
  // ... rest of address specific fields
}
```

I had instead created a new namespace `fields` that if present, takes precedence over the "legacy" way of translating form field error messages. Thoughts?

davidchin: I guess we have to take a step back a little bit. The interface of `ERROR_KEYS` (in master) is `{ [fieldName: string]: string }`. If you extract the translation keys out and pass them in, the interface is still `{ [fieldName: string]: string }`. You won't be passing translation keys specifically for `address`.

If you don't want to pass in translation keys, what you can do is to pass a `translate` function into the component, whose interface can look something like this `translate(key: string): string`. Similarly, it's not tied to `address`. So in both cases, you won't hit the problem you're describing and we can achieve the generalisation we want to achieve.

capsula4: I ended up moving address-specific keys up, but still left what I believe are general errors that should be shared. Let me know what you think. For compatibility with existing translations, I had to leave the old key. Or I could keep the general keys under the "address" namespace. Thoughts?

davidchin: Cool, I think you're almost there. Have you tried passing in a `translate` function instead, it seems to me that it's more fitting for your situation, e.g.:

```ts
const addressKeys = {
    address1: 'address.address_line_1',
    // etc...
};

// `address/translateAddressFieldError.ts`
function translateAddressFieldError(language: Language): string {
  return (fieldName: string, errorType: string) => {
    if (errorType === 'required') {
      return language.translate(`${addressKeys[fieldName]}_required_error`);
    }

    if (errorType === 'max') {
      return language.translate('address.max_error');
    }

    // etc...
  }
}
```

Then you can pass it into `getFormFieldsValidationSchema`

```ts
getFormFieldsValidationSchema({ translate: translateAddressFieldError })

function getFormFieldsValidationSchema({ translate }) {
  // etc...

  if (maxValue !== undefined) {
    schema[name] = schema[name].max(maxValue, translate(name, 'max'))
  }

  if (required) {
    schema[name] = schema[name].required(translate(name, 'required'))
  }

  // etc...
}
```

davidchin (approved):
Nice one, LGTM! 👍 :shipit:

Issue #475, opened 2020-12-03
Documentation contains incorrect instructions to reference hosted checkout /dist auto-loader in Checkout (Advanced Options)


Mikhail-MM:
Excited to work with the Optimized One Page Checkout source code finally - but had some trouble getting the package built & hosted correctly.

To test that my build/hosting process is correct, I made a few changes to `Checkout.tsx` - just a few custom console.logs and classNames to the React components as a sanity check to make sure that my changes were being applied.

I'm following the instructions from the BigCommerce Documentation:
https://developer.bigcommerce.com/stencil-docs/customizing-checkout/installing-custom-checkouts
```
To install a custom checkout on a store, follow these steps:

Navigate to Advanced Settings > Checkout in your store’s control panel.

Under Checkout Type, select Custom Checkouts.

Type the following in the Script URL field, replacing with the latest version number generated by running the npm run release command:

**webdav:checkout/auto-loader-<version>.js**
```

First thing to mention is that using `npm release` in this repo as the docs suggest does not work, and returns this error:
```
>echo 'Please do not relase locally, use CircleCi'
```

So, after looking at the dist folder, there was a file named `auto-loader-1.104.1`, so I used the following URL:

```
webdav:checkout/auto-loader-1.1014.1.js
```

When I logged into my live store, I saw the checkout, however my custom changes were missing completely. Now, it **was** still pulling _something_ from WebDAV because when I changed the version number to something that did not exist, i.e. `auto-loader-12345.js`, the checkout page was blank and failed to load completely.

So, how I solved this was noticing that there is a file called `auto-loader.js` without any version #.

Using the following URL, my changes were able to load successfully:

```
webdav:checkout/auto-loader.js
```

So really, no version number is required when pointing to the auto-loader in the BigCommerce Control Panel, and that's quite confusing. 

In addition, you guys may want to mention something about the build process not working at all on Windows, this may save people some headaches. I was running into cryptic npm errors and half-built dist/build folders until I switched to my MacOS laptop to give it a try.

capsula4:
Hi Mikhail, thanks for raising this issue. 

While we look into it, we advise that you run instead `npm run release:alpha` to test your changes.

The version number is not required, but highly recommended. If you make changes and you don't use the new version number, a cached version could be served to the users.

capsula4:
https://github.com/bigcommerce/checkout-js/pull/476

Mikhail-MM:
[...]

PR #476, opened 2020-12-04
docs(common): ISSUE-475 Update OS requirement

capsula4:
## What?
Update OS requirement, as per https://github.com/bigcommerce/checkout-js/issues/475

## Why?
So devs know that Unix-like is supported. In order to use Windows, they might need a Unix simulator tool.

## Testing / Proof
n/a

@bigcommerce/checkout

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@ In order to build from the source code, you must have the following set up in yo
 
 * Node >= v10.
 * NPM >= v3.
+* Unix-based operating system.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,6 +8,7 @@ In order to build from the source code, you must have the following set up in yo
 
 * Node >= v10.
 * NPM >= v3.
+* Unix-based operative system
```
TomA-R, line 11:
```suggestion
* Unix-based operating system
```

PR #484, opened 2020-12-21
CHECKOUT-5477 Check if multi-shipping is enabled once config is loaded

capsula4:
## What?
Check if multi-shipping is enabled once checkout has loaded.

## Why?
At the moment we check if the setting is enabled before the config is loaded, which will result in always false, causing single-shipping mode to always be loaded after a new address is added via the account settings page.

## Testing / Proof
multishipping

@bigcommerce/checkout

### DIFF
--- a/src/app/checkout/Checkout.spec.tsx
+++ b/src/app/checkout/Checkout.spec.tsx
@@ -62,7 +62,18 @@ describe('Checkout', () => {
         };
 
         jest.spyOn(checkoutService, 'loadCheckout')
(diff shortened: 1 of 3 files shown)


ankurbigcomm (approved):
Changes look good. 

Just wanted to understand why we had to move the spy on checkoutState.data inside the mocked implementation of loadCheckout.. Not sure why it could be failing when it was outside.

capsula4:
The spy was moved to better reflect what happens in the real world. getConfig() only returns data after loadCheckout() has been called.

PR #493, opened 2021-01-15
CHECKOUT-5529 Render gift wrapping in cart summary

capsula4:
## What?
Render gift wrapping as product description

## Why?
Because currently we do not render this information, and totals won't match

## Testing / Proof
Screen Shot 2021-01-15 at 12 19 14 pm

@bigcommerce/checkout

### DIFF
--- a/src/app/order/OrderSummaryItem.spec.tsx
+++ b/src/app/order/OrderSummaryItem.spec.tsx
@@ -69,4 +69,21 @@ describe('OrderSummaryItems', () => {
                 .not.toContain('product-price--beforeDiscount');
         });
     });
+
+    describe('when description is present', () => {
+        beforeEach(() => {
+            orderSummaryItem = shallow(<OrderSummaryItem
+                amount={ 10 }
+                description="Description"
+                id="foo"
+                name="Product"
+                quantity={ 2 }
+            />);
+        });
+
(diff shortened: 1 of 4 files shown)


RobinOng (approved):

theromulans:
I'm surprised to see that the money value isn't right aligned here?

To me this reads as if the $10.00 is included in the line item price, which it clearly is not.

capsula4:
@theromulans this was discussed with Valter and Charles. We took this approach as it's less likely to cause problems with potential checkout customizations. This adds a new element to the DOM, while right aligning it would require a bigger refactor around how we display line items, potentially breaking things with customizations.

https://bigcommerce.slack.com/archives/C0WC9CSJX/p1610670045059300

theromulans:
Thanks for the explanation @capsula4 - understood.

davidchin (approved):
Yes, I agree with @theromulans, the amounts don't seem to visually add up. But I can't come up with a change without changing the existing DOM. So maybe it's good enough for now. Perhaps we can do some CSS magic or change the design if it turns out that people are unhappy.

capsula4:
Yeah I agree, I think it would look nicer if the numbers added up, but this seems to be lower risk and acceptable for now.

PR #494, opened 2021-01-18
 CHECKOUT-2322 Show gift wrapping cost

capsula4:
## What?
Show gift wrapping cost in order summary. Also "tag" items that contains gift wrapping
Note: depends on SDK Changes and API changes
https://github.com/bigcommerce/bigcommerce/pull/39048
https://github.com/bigcommerce/checkout-sdk-js/pull/1058

## Why?
So totals add up and shoppers can understand the difference between the subtotal and the grand total.

## Testing / Proof
unit
Tested in conjunction with the API changes, see screenshots in https://github.com/bigcommerce/bigcommerce/pull/39048

@bigcommerce/checkout

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1608,9 +1608,9 @@
       }
     },
     "@bigcommerce/checkout-sdk": {
-      "version": "1.121.1",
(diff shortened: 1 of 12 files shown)


davidchin (approved):
🚢 👍

PR #497, opened 2021-01-26
feat(payment): PAYPAL-876 added onError handler

andriiVitvitskyi1990:
## What?
PAYPAL-876 added onError handler

## Why?
According to task PAYPAL-876. Added onError handler to display error popup when polling timer expired

## Testing / Proof
tested on dev

@bigcommerce/checkout @davidchin

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -383,24 +383,24 @@
       },
       "dependencies": {
         "@babel/helper-plugin-utils": {
-          "version": "7.10.4",
-          "resolved": "https://registry.npmjs.org/@babel/helper-plugin-utils/-/helper-plugin-utils-7.10.4.tgz",
-          "integrity": "sha512-O4KCvQA6lLiMU9l2eawBPMf1xPP8xPfB3iEQw150hOVTqj/rfXz0ThTb4HEzqQfs2Bmo5Ay8BzxfzVtBrr9dVg=="
+          "version": "7.12.13",
+          "resolved": "https://registry.npmjs.org/@babel/helper-plugin-utils/-/helper-plugin-utils-7.12.13.tgz",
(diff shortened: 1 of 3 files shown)


BC-ACherednichenko (approved):
LGTM, 

only one concern that we need to check before the merge

`onError` should be an optional param in `checkout-sdk`,  so we could keep compatibility with earlier checkout versions or custom checkout implementations

src/app/payment/paymentMethod/PaypalCommercePaymentMethod.tsx
```diff
@@ -29,6 +29,9 @@ const PaypalCommercePaymentMethod: FunctionComponent<PaypalCommercePaymentMethod
                 setSubmitted(true);
                 submitForm();
             },
+            onError: (error: any) => {
```
davidchin, line 32: You don't need `any` as the parameter should already be typed.

davidchin (approved):

src/app/payment/paymentMethod/PaypalCommercePaymentMethod.tsx
andriiVitvitskyi1990: fixed

package.json
```diff
@@ -26,7 +26,7 @@
   },
   "homepage": "https://github.com/bigcommerce/checkout-js#readme",
   "dependencies": {
-    "@bigcommerce/checkout-sdk": "^1.124.1",
+    "@bigcommerce/checkout-sdk": "^1.125.0",
```
BC-ACherednichenko, line 29: Can you update `package-lock.json` as well?

andriiVitvitskyi1990: fixed

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@bigcommerce/checkout",
-  "version": "1.110.3",
+  "version": "1.110.0",
```
BC-ACherednichenko, line 3: It should not change the version on the module, it is managed by circle-ci

andriiVitvitskyi1990: fixed

PR #501, opened 2021-01-29
feat(shipping): CHECKOUT-4726 Allow adding address in multishipping UI

capsula4:
## What?
Allow adding address in multishipping UI

## Why?
To improve UX. This way, shoppers do not have to be taken out of Checkout.

## Testing / Proof
multishipping2

#### Validation
Screen Shot 2021-01-29 at 4 31 09 pm

@bigcommerce/checkout


src/app/address/AddressFormModal.tsx
```diff
@@ -0,0 +1,116 @@
...
+    countriesWithAutocomplete,
+    countries,
+    values,
+    setFieldValue,
+    isLoading,
+}) => (
+    <Modal
+        additionalModalClassName="modal--medium"
+        header={
+            <ModalHeader>Add Address</ModalHeader>
```
davidchin, line 46: The title "Add Address" needs to be translatable if possible.

src/app/shipping/MultiShippingForm.tsx
```diff
@@ -87,32 +103,92 @@ class MultiShippingForm extends PureComponent<MultiShippingFormProps & WithLangu
...
+                    getFields={ getFields }
+                    googleMapsApiKey={ googleMapsApiKey }
                     isLoading={ isLoading }
-                    isMultiShippingMode={ true }
-                    shouldDisableSubmit={ this.shouldDisableSubmit() }
-                    shouldShowOrderComments={ shouldShowOrderComments }
-                    shouldShowShippingOptions={ !hasUnassignedLineItems(consignments, cart.lineItems) }
-                />
-            </Form>
+                    isOpen={ true }
```
davidchin, line 114: Should you just pass `itemAddingAddress` to `isOpen`? It seems to be me that it's a variable to control the visibility of the modal?

src/app/shipping/MultiShippingForm.tsx
```diff
@@ -87,32 +103,92 @@ class MultiShippingForm extends PureComponent<MultiShippingFormProps & WithLangu
...
+        if (!itemAddingAddress) {
+            return;
+        }
+
+        const shippingAddress = mapAddressFromFormValues(address);
+
+        try {
+            createCustomerAddress(shippingAddress);
+        } catch (_) {
+            // Fail silently if customer address was not saved.
```
davidchin, line 159: What do you mean by fail silently? Why don't we show an alert or something to tell the shopper that something went wrong? It feels odd if we don't give any visual feedback and nothing is saved after clicking on the save button?

src/app/shipping/updateShippableItems.ts
```diff
@@ -29,14 +29,17 @@ export default function updateShippableItems(
...
-        const id = newId && (i === updatedItemIndex || !cartItemIds.includes(item.id)) ?
-            newId : item.id;
-
-        updatedItems[i] = {
-            ...item,
-            id,
-            consignment: findConsignment(consignments || [], id  as string),
-        };
+        if (newId && !cartItemIds.includes(item.id) || i === updatedItemIndex) {
+            const itemId = newId || item.id;
```
davidchin, line 33: Prefer to use `??` unless there's a good reason not to?

src/app/shipping/updateShippableItems.ts
```diff
@@ -29,14 +29,17 @@ export default function updateShippableItems(
...
+
+            updatedItems[i] = {
+                ...item,
+                id: itemId,
+                consignment: findConsignment(consignments || [], itemId as string),
+            };
+        } else {
+            updatedItems[i] = item;
```
davidchin, line 41: This `foreach` loop can probably be rewritten as a `map` 🍹

src/app/address/AddressFormModal.tsx
```diff
@@ -0,0 +1,116 @@
...
+    shouldShowSaveAddress?: boolean;
+    isOpen: boolean;
+    defaultCountryCode?: string;
+    getFields(countryCode?: string): FormField[];
+    onSaveAddress(address: AddressFormValues): void;
+    onRequestClose?(): void;
+    onAfterOpen?(): void;
+}
+
+const AddressModal: FunctionComponent<AddressFormModalProps & WithLanguageProps & FormikProps<AddressFormValues>> = ({
```
davidchin, line 31: It'd be nice if the name of the component is the same as the file (`AddressFormModal`).

src/app/address/AddressFormModal.tsx
```diff
@@ -0,0 +1,116 @@
...
+                    <a
+                        className="button optimizedCheckout-buttonSecondary"
+                        href="#"
+                        onClick={ preventDefault(onRequestClose) }
+                    >
+                        <TranslatedString id="common.cancel_action" />
+                    </a>
+
+                    <Button
+                        disabled={ isLoading }
```
davidchin, line 74: Will `isLoading` be true when in the process of creating an address? I can't seem to find where the pending status is passed.

src/app/shipping/MultiShippingForm.spec.tsx
```diff
@@ -92,5 +96,23 @@ describe('MultiShippingForm Component', () => {
                 })
             );
         });
+
+        it('renders address form modal when shoppers choose new address', async () => {
+            component.find('.consignmentList > li')
```
davidchin, line 101: Prefer to use `data-test` to select elements in tests rather than CSS class names.

davidchin (commented):
Nice one! I would love to see a screenshot of the mobile view. I'm just curious how it looks and whether scrolling still works well. 🙂

src/app/shipping/MultiShippingForm.tsx
capsula4: I could do that, but I actually prefer to completely destroy the modal. The reason is that if I don't destroy the modal, the form values are remembered and the second time you open the modal to add a second address, it will remember the values from the first address.

src/app/shipping/MultiShippingForm.tsx
capsula4: The consignment would have got the address updates, but not saved in the customer address book. The intention was to handle more gracefully the scenario where customer service is down.

src/app/address/AddressFormModal.tsx
capsula4: yeah, it's currently being passed. It's the generic `isLoading` status from the SDK

capsula4:
Sure can attach a mobile view. The modal already supports scrolling. Even for desktop it relies on the scrolling, as it's quite long (as it can be seen in the screenshot).

src/app/address/AddressFormModal.tsx
davidchin: Oh okay. Are you sure though? 🤔  I tried to trace back the code and I found that `isLoading` is passed from `Shipping` component (https://github.com/bigcommerce/checkout-js/blob/master/src/app/shipping/Shipping.tsx#L300) and it only covers certain loading statuses, such as updating consignment etc... And creating an address is not in the list? But I might be reading the code wrong 😅

davidchin (commented):
Thank you for the mobile screenshot. I just wanted to double check because I know with a touch device, we need to do some extra work (i.e.: apply `overflow: hidden` to the `body`) to ensure that when a user scroll a modal, they don't scroll the page along as well. It seems we've already done the work before, which is good!

snaderiBC:
Hey @capsula4 Do you know where we define the `AddressType`? I can't see it here (or in other UIs) but it is there in the codebase.

src/app/shipping/MultiShippingForm.tsx
capsula4: I didn't think about address creation for other reasons as we are supposed to validate the input in the frontend anyway. Only submit a form when the values are actually valid.

If the service was down, no address would be saved so I thought it could be quite disruptive to show that error continuously for every new address, but I guess it just takes a single click to close a modal.

src/app/address/AddressFormModal.tsx
capsula4: I think I got confused with the loading spinner being showed up. I'm not actually checking for the isCreatingAddress flag as I'm currently doing that in the "background". I guess I could wait for both the consignment update and the address creation.

capsula4:
Added error handling for when customer address fails to be saved
Screen Shot 2021-02-04 at 12 46 51 pm

capsula4:
@davidchin  ready for a re-review. Most checks failing due to missing Checkout SDK types

src/app/shipping/MultiShippingForm.tsx
```diff
@@ -87,32 +107,101 @@ class MultiShippingForm extends PureComponent<MultiShippingFormProps & WithLangu
...
+        }
+
+        const shippingAddress = mapAddressFromFormValues(address);
+
+        await this.handleSelectAddress(shippingAddress, itemAddingAddress.itemId, itemAddingAddress.key);
+
+        try {
+            await createCustomerAddress(shippingAddress);
+        } catch (e) {
+            onUnhandledError(new CreateCustomerAddressError(e));
```
davidchin, line 174: To answer your concern regarding displaying the error as a modal. What about if we display the error as an `Alert` inside the modal, considering the error already happens inside a modal. The UX pattern is similar to manage payment instruments modal. And the user doesn't need to reopen the modal and retype everything to retry.

Another question that I have is, what happens if you refresh the page after receiving such error. I assume the dropdown won't be displaying an address because it's designed to display a list of "saved" addresses? So what would the experience be like?

Finally the error message doesn't adequately communicate that the shopper can continue to check out despite the fact that their address is not saved to their address book. Imagine you're a shopper who doesn't know the implementation, when I see that error message, would I still think that I can proceed? Or would I feel something's broken.

src/app/customer/error/CreateCustomerAddressError.ts
```diff
@@ -0,0 +1,14 @@
+import { setPrototypeOf, CustomError } from '../../common/error';
+import { getLanguageService } from '../../locale';
+
+export default class CreateCustomerAddressError extends CustomError {
+    constructor(data: Error) {
+        super({
+            name: 'CREATE_CUSTOMER_ADDRESS_FAILED',
+            message: getLanguageService().translate('customer.create_address_error'),
```
davidchin, line 8: Generally speaking it is not recommended to have translation string as the message of `Error` instance. This is because if the error bubbles up and get logged by the Sentry logger. The translated error message will be logged. If the error message is translated into a different language (or re-worded in English), it'll create a separate Sentry error report than the ones that are not translated. So it'd make our logs noisy. Another reason is that we probably don't know how to read the message in another language, and it'd take us time to figure out what it actually means.

capsula4:
Screen Shot 2021-02-08 at 11 04 47 am

capsula4:
@davidchin I've addressed your feedback, let me know what you think, thanks!

src/app/shipping/MultiShippingForm.tsx
```diff
@@ -87,32 +105,108 @@ class MultiShippingForm extends PureComponent<MultiShippingFormProps & WithLangu
...
-                    cartHasChanged={ cartHasChanged }
-                    isLoading={ isLoading }
-                    isMultiShippingMode={ true }
-                    shouldDisableSubmit={ this.shouldDisableSubmit() }
-                    shouldShowOrderComments={ shouldShowOrderComments }
-                    shouldShowShippingOptions={ !hasUnassignedLineItems(consignments, cart.lineItems) }
+            <Fragment>
+                <ErrorModal
+                    error={ createCustomerAddressError }
+                    message={
```
davidchin, line 111: There's only one error at the moment so what you have is good. But if in the future we have more than one error, we could do something like https://github.com/bigcommerce/checkout-js/blob/0d34c09f6b324921e2a9c4958bbdaab6382043de/src/app/payment/Payment.tsx#L206 where we can map to a different error message depending on its type.

davidchin (approved):
LGTM!! 👍

PR #521, opened 2021-02-23
feat(payment): PAYMENTS-6576 remove the label "Payment Methods" in checkout

ngodinhloc:
## What?
- remove the label "Payment Methods" in the checkout 

## Why?
- it not needed

## Testing / Proof
image

@bigcommerce/checkout

### DIFF
--- a/src/app/payment/PaymentForm.tsx
+++ b/src/app/payment/PaymentForm.tsx
@@ -4,9 +4,9 @@ import { isNil, noop, omitBy } from 'lodash';
 import React, { memo, useCallback, useContext, useMemo, FunctionComponent } from 'react';
 import { ObjectSchema } from 'yup';
 
-import { withLanguage, TranslatedString, WithLanguageProps } from '../locale';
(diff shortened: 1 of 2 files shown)


RobinOng (approved):

src/app/payment/PaymentForm.tsx
```diff
@@ -215,14 +215,8 @@ const PaymentMethodListFieldset: FunctionComponent<PaymentMethodListFieldsetProp
         setSubmitted,
     ]);
 
-    const legend = useMemo(() => (
-        <Legend>
-            <TranslatedString id="payment.payment_method_label" />
```
TomA-R, line 220: I think we could remove this from [the translations file](https://github.com/bigcommerce/checkout-js/blob/2fd2e57016e00ec77eded07759af7720749eafd6/src/app/locale/translations/en.json#L223) too? cc @bigcommerce/checkout

ngodinhloc: good catch @TomA-R

PR #523, opened 2021-02-26
fix(checkout): CHECKOUT-4936 Display shipping step with custom item c…

bdav87:
…arts

## What?
Displaying the shipping step if a cart only has custom items.

## Why?
Custom items are pseudo-physical items per our existing functionality in manual orders and LCO. 

Currently draft orders with only custom items can't adjust a shipping address adjusted in checkout, and S2S carts consisting of only custom items cannot complete a checkout due to a missing shipping method error.

By surfacing the shipping step when custom items are present, draft orders can have their shipping address adjusted by shoppers, and checkout can be completed when carts created via S2S API only have custom items.

## Testing / Proof
Before:
image

After:
image

@bigcommerce/checkout

### DIFF
--- a/src/app/checkout/getCheckoutStepStatuses.ts
+++ b/src/app/checkout/getCheckoutStepStatuses.ts
@@ -5,7 +5,7 @@ import { createSelector } from 'reselect';
 import { isValidAddress } from '../address';
 import { EMPTY_ARRAY } from '../common/utility';
 import { SUPPORTED_METHODS } from '../customer';
-import { hasSelectedShippingOptions, hasUnassignedLineItems } from '../shipping';
+import { hasSelectedShippingOptions, hasUnassignedLineItems, itemsRequireShipping } from '../shipping';
 
 import CheckoutStepType from './CheckoutStepType';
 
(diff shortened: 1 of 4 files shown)


src/app/checkout/getCheckoutStepStatuses.ts
```diff
@@ -77,14 +77,26 @@ const getShippingStepStatus = createSelector(
...
+                const physicalItems = cart.lineItems.physicalItems.some(lineItem => lineItem.isShippingRequired);
+                if (config && config.checkoutSettings.features['CHECKOUT-4936.enable_custom_item_shipping'] && cart.lineItems.customItems) {
+                    return cart.lineItems.customItems.length > 0 || physicalItems;
+                }
+
+                return physicalItems;
+            }
+
+            return false;
+        })();
```
capsula4, line 99: rather than creating a function that calls itself, maybe create an actual function in a new file? Or even re-use the function `getShippableItemsCount` and modify so that it can take a second param

`getShippableItemsCount(cart, includeCustomItems: bool = false)`

bdav87: I opted to write a new function in a separate file, since the intent is to check if anything in cart requires shipping, versus counting the shippable items.

src/app/shipping/itemsRequireShipping.ts
```diff
@@ -0,0 +1,16 @@
+import { Cart, StoreConfig } from '@bigcommerce/checkout-sdk';
+
+const itemsRequireShipping = (cart: Cart | undefined, config: StoreConfig | undefined) => {
```
capsula4, line 3: rather than `Cart | undefined` or `StoreConfig | undefined` you can use the optional param `?`

Also, would be good if you can add unit tests for this function.

src/app/shipping/itemsRequireShipping.ts
```diff
@@ -0,0 +1,16 @@
...
+    if (cart) {
+        const physicalItems = cart.lineItems.physicalItems.some(lineItem => lineItem.isShippingRequired);
+        if (config && config.checkoutSettings.features['CHECKOUT-4936.enable_custom_item_shipping'] && cart.lineItems.customItems) {
+            return cart.lineItems.customItems.length > 0 || physicalItems;
```
capsula4, line 7: instead of having `|| physicalItems;` you can simplify the logic by returning early

```js
if (!cart) {
  return false;
}

if (cart.lineItems.physicalItems.some(lineItem => lineItem.isShippingRequired)) {
  return true;
}

if (config && config.checkoutSettings.features['CHECKOUT-4936.enable_custom_item_shipping'] && cart.lineItems.customItems) {
  return cart.lineItems.customItems.length > 0
}

return false;

capsula4 (approved):

bdav87:
@bigcommerce/cpoe-pod Can this be merged or is there something else that has to be done first?

PR #526, opened 2021-03-03
CHECKOUT-5619 Update text when account has been created

capsula4:
## What?
Update rendered text when account has been created

## Why?
Because it's misleading, as the account has been created and they actually set a password when filling in the form

## Testing / Proof
Screen Shot 2021-03-03 at 2 05 52 pm
Screen Shot 2021-03-03 at 2 07 01 pm

@bigcommerce/checkout


mjhmatt:
The text is slightly different from the ticket? I thought we wanted this:

An account has been created for you, but you'll need to setup your password to login.  

With an account:

- You can proceed through check out faster
- Save shipping addresses
- Access your order history
- Track new orders

?

capsula4:
@mjhmatt I checked with @bc-charlesho and he is okay with it. I could add the extra text but IMO it's too verbose and distracts the user from the main actions: create an account + continue shopping

bc-charlesho:
@capsula4 @mjhmatt the copy looks good. We decided to revise the copy to make it less wordy. The key point being  to fix the original copy to let the shopper know that an account as been created for them.

RobinOng (approved):

PR #530, opened 2021-03-09
fix(checkout): ADYEN-38 Fix Adyen icons on checkout page

vneutrino:
## What?
Made logos for Klarna methods on checkout consistent.

## Why?
According to task https://jira.bigcommerce.com/browse/ADYEN-38
Logos for Klarna methods on checkout seem inconsistent. 

#### BEFORE:
image

### AFTER:
image

@bigcommerce/checkout

### DIFF
--- a/src/app/payment/paymentMethod/PaymentMethodTitle.tsx
+++ b/src/app/payment/paymentMethod/PaymentMethodTitle.tsx
@@ -123,8 +123,8 @@ function getPaymentMethodTitle(
         }
 
         return (
-            customTitles[method.id] ||
             customTitles[method.gateway || ''] ||
+            customTitles[method.id] ||
             customTitles[method.method] ||
             customTitles[PaymentMethodType.CreditCard]
         );


BC-ACherednichenko (approved):
Nicely done!

davidchin (commented):
Why don't you just fix the image itself instead of changing the code (i.e.: point to a different image other than `/img/payment-providers/klarna-header.png`)?

vneutrino:
@davidchin because AdyenV2 payment methods icons have smaller size than icons from our CDN.
If I change Klarna icon at our CDN, we will have inconsistency in icons with another payment gateway.
That's why I've changed an order in title selection, the `method.gateway` has a higher priority than `method.id`.

davidchin:
I'm not entirely sure why Adyen has anything to do with Klarna 🤔

vneutrino:
@davidchin Adyen provides Klarna as payment method for EU countries

davidchin (approved):

PR #531, opened 2021-03-09
feat(payment): INT-3969 Use APMs logos instead names in checkout accordion for Checkout.com

MLoyo:
[INT-3969](https://jira.bigcommerce.com/browse/INT-3969)

## What?
Use logos for multiple APMs of checkout.com payment method instead of names

## Why?
In order to replace text and use the images for the multiple APMs that checkout.com let use

## Testing / Proof
Screen Shot 2021-03-09 at 3 19 42 PM
Screen Shot 2021-03-09 at 3 20 09 PM
Screen Shot 2021-03-09 at 3 20 25 PM

## Depends on

- bigcommerce/bigcommerce#39812

@bigcommerce/checkout

### DIFF
--- a/src/app/payment/paymentMethod/PaymentMethodTitle.spec.tsx
+++ b/src/app/payment/paymentMethod/PaymentMethodTitle.spec.tsx
@@ -272,4 +272,34 @@ describe('PaymentMethodTitle', () => {
         expect(component.find(CreditCardIconList).prop('selectedCardType'))
             .toEqual('visa');
     });
+
+    it('renders only Checkout.com APMs logos based on their gateway id', () => {
+        const checkoutcomTitleComponent = (id: string) => mount(
+            <PaymentMethodTitleTest
+                { ...defaultProps }
+                method={ {
+                    ...defaultProps.method,
+                    method: PaymentMethodId.Checkoutcom,
+                    id,
+                } }
+            />
+        );
+        const baseURL = (id: string) => `/img/payment-providers/checkoutcom_${id}.png`;
+
+        let component = checkoutcomTitleComponent('sepa');
+        expect(component.find('[data-test="payment-method-logo"]').prop('src'))
+            .toEqual(`${config.cdnPath}${baseURL('sepa')}`);
+
+        component = checkoutcomTitleComponent('oxxo');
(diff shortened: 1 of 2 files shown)


vmparra (changes requested):
Perhaps some test cases?

MLoyo:
> Perhaps some test cases?

♻️

MLoyo:
@vmparra ♻️

vmparra (approved):
LGTM 👍🏻

src/app/payment/paymentMethod/PaymentMethodTitle.spec.tsx
```diff
@@ -272,4 +272,30 @@ describe('PaymentMethodTitle', () => {
...
+
+    it('renders only Checkout.com APMs logos based on their gateway id', () => {
+        const checkoutcomAPMs = [
+            'sepa',
+            'oxxo',
+            'boleto',
+            'qpay',
+        ];
+
+        checkoutcomAPMs.forEach(apm => {
```
capsula4, line 284: rather than doing a foreach, you could create a testing component instead?
```js
const TestComponent = id => (<PaymentMethodTitleTest
                { ...defaultProps }
                method={ {
                    ...defaultProps.method,
                    method: PaymentMethodId.Checkoutcom,
                    id,
                } }
            />)

let component = <TestComponent id="sepa" />;
expect(component.find('[data-test="payment-method-logo"]'))
                .toEqual('sepaLogoUrl');
component = <TestComponent id="oxxo" />;
expect(component.find('[data-test="payment-method-logo"]'))
                .toEqual('oxxoLogoUrl');

// ...

MLoyo:
@capsula4 Thanks for your suggestion, I just applied  it in the PR ♻️

capsula4 (approved):
thanks! looks good 👍

PascalZajac:
This seems worse from an accessibility perspective?

cmedinilla:
@PascalZajac AFAIK we are using this approach for all the APMs across gateways, we need start a discussion about if this is the best approach for all the APMs across the platform

PascalZajac:
@cmedinilla understood, thanks for the context.

PR #535, opened 2021-03-16
CHECKOUT-5618 Check also `stateOrProvinceCode` for address equality

capsula4:
## What?
Check state code for address equality, and any combination between `stateOrProvince` and `stateOrProvinceCode`

## Why?
Sometimes, when comparing if two addresses are equal, we might be comparing a request payload and a response payload. Request payloads may only contain `stateOrProvince` or `stateOrProvinceCode`, and for consignments, the `stateOrProvince` also accepts state codes as values.

## Testing / Proof
Unit

@bigcommerce/checkout

### DIFF
--- a/src/app/address/isEqualAddress.ts
+++ b/src/app/address/isEqualAddress.ts
@@ -12,13 +12,27 @@ export default function isEqualAddress(address1?: ComparableAddress, address2?: 
     return isEqual(
         normalizeAddress(address1),
         normalizeAddress(address2)
-    );
+    ) && isSameState(address1, address2);
+}
+
+function isSameState(address1: ComparableAddress, address2: ComparableAddress): boolean {
+    if (address1.stateOrProvince && address1.stateOrProvince === address2.stateOrProvince) {
+        return true;
+    }
+
+    if (address1.stateOrProvinceCode && address1.stateOrProvinceCode === address2.stateOrProvinceCode) {
+        return true;
+    }
+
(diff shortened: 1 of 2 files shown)


src/app/address/isEqualAddress.ts
```diff
@@ -12,13 +12,26 @@ export default function isEqualAddress(address1?: ComparableAddress, address2?:
...
         normalizeAddress(address2)
-    );
+    ) && isSameState(address1, address2);
+}
+
+function isSameState(address1: ComparableAddress, address2: ComparableAddress): boolean {
+    const addressValues1 = [address1.stateOrProvince, address1.stateOrProvinceCode];
+    const addressValues2 = [address2.stateOrProvince, address2.stateOrProvinceCode];
+
+    if (addressValues1.some(value => addressValues2.includes(value) && !!value)) {
```
davidchin, line 22: I think you should only compare state code against state code and state name against state name. This line compares state code with state name and return true if they are the same.

> ...and for consignments, the stateOrProvince also accepts state codes as values

I don't think that is true. We automatically set the state code (`stateOrProvinceCode`) using the state name (`stateOrProvince`) if only the latter is provided. But we don't accept it as a state code if it is passed in as state name.

capsula4: I had the impression that we allowed that but after testing I found we actually don't. I have refactored accordingly. Thanks!

davidchin (approved):
👍

PR #549, opened 2021-04-08
feat(payment): PAYPAL-976 Order placement v2

vneutrino:
## What?
Added APM fields to checkout

## Why?
Due to https://jira.bigcommerce.com/browse/PAYPAL-976

## Testing / Proof
Tested manually
image

@bigcommerce/checkout

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1655,9 +1655,9 @@
       }
     },
     "@bigcommerce/checkout-sdk": {
-      "version": "1.147.1",
-      "resolved": "https://registry.npmjs.org/@bigcommerce/checkout-sdk/-/checkout-sdk-1.147.1.tgz",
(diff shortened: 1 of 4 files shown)


andriiVitvitskyi1990 (approved):

src/app/payment/PaymentForm.tsx
```diff
@@ -139,6 +139,8 @@ const PaymentForm: FunctionComponent<PaymentFormProps & FormikProps<PaymentFormV
                 values={ values }
             />
 
+            <div className="fieldsContainer" id="checkout-payment-fields" />
```
BC-ACherednichenko, line 142: per my comment here https://github.com/bigcommerce/checkout-sdk-js/pull/1105#discussion_r616808766, there is no need to create a separate container in the payment form, each payment method renderer can handle this part next to radio-buttons, with such widget as https://github.com/bigcommerce/checkout-js/blob/051363765fb15c03d818dcb09d0668be574f490a/src/app/payment/paymentMethod/HostedWidgetPaymentMethod.tsx#L269

BC-ACherednichenko (commented):
I think we have to implement this feature consistently with other providers

capsula4 (approved):

PR #565, opened 2021-04-27
feat(payment): INT-4071 Clearpay - Create payment method

roberbnd:
## What? [INT-4071](https://jira.bigcommerce.com/browse/INT-4071)
Create the Clearpay form that will be displayed on the checkout page.

## Why?
We need Clearpay to be displayed as payment option when merchants have enabled it

## Testing / Proof
https://drive.google.com/file/d/1zoG-uAWQ0LSMogJGCi1vBSpNPXh180KZ/view?usp=sharing


capsula4 (approved):

josesonora:
💚 Tested:

- Clearpay is displayed as payment method only when order amount is between limits

PR #581, opened 2021-05-12
[DRAFT] PAYMENTS-6796 How we will localise payment error messages

Tharaae:
## What?
Currently, error messages coming from BigPay are displayed without being translated. Instead, we want to use translated messages based on the error codes.

**Open question**: How shall we handle non-existing error code?

## Why?
So shoppers read payment error messages in their own languages.

## Testing / Proof
Screen Shot 2021-05-12 at 10 59 45 am

Screen Shot 2021-05-12 at 10 57 30 am

### DIFF
--- a/src/app/locale/translations/en.json
+++ b/src/app/locale/translations/en.json
@@ -282,7 +282,11 @@
             "visa_checkout_continue_action": "Continue with Visa Checkout",
             "zip_continue_action": "Continue with Zip",
             "zip_name_text": "Zip",
-            "zip_display_name_text": "Own it now, pay later"
+            "zip_display_name_text": "Own it now, pay later",
+            "errors": {
+                "incorrect_address": "Unable to process the payment because invalid data was supplied with the transaction. (translated)",
+                "transaction_declined": "Payment was declined. Please try again. (translated)"
+            }
         },
         "redeemable": {
             "applied_text": "Applied",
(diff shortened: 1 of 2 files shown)


capsula4 (commented):
I wonder if it's better to perform the translation at the API level. Not sure what's our approach as a platform, but most of our Checkout Storefront APIs are localized.

The problem with this approach is that if a new error code is added in BigPay, it will require a checkout-js release as well. This will make the whole cycle lengthier in order to provide with a translation.

The advantage of having it checkout-js is that merchants could decide to customize the error message in whatever way they wanted.

In saying so, I think both approaches are compatible and complementary. However, if we had to choose one, I think localizing the error messages at the API level might be a better start.

Tharaae:
> I wonder if it's better to perform the translation at the API level. Not sure what's our approach as a platform, but most of our Checkout Storefront APIs are localized.
> 
> The problem with this approach is that if a new error code is added in BigPay, it will require a checkout-js release as well. This will make the whole cycle lengthier in order to provide with a translation.
> 
> The advantage of having it checkout-js is that merchants could decide to customize the error message in whatever way they wanted.
> 
> In saying so, I think both approaches are compatible and complementary. However, if we had to choose one, I think localizing the error messages at the API level might be a better start.

Thanks for the review @capsula4
My understanding in that translations cannot be done in BigPay (I guess it's unaware of the store language). Moreover, we do want  merchants to be able to customise the error messages as well.

TomA-R:
We'll probably want to put this behind an experiment to control the rollout (I know, it's still a draft)

PR #586, opened 2021-05-17
feat(payment): PAYMENTS-6813 add wip ppsdk payment method

leeBigCommerce:
## What?

- Add a new `PPSDKPaymentMethod`
- Add a map of initialisation strategies to Functional Components which will render the UI
- Add mounting logic to call init/deinit strategy methods upon rendering a valid initialisation strategy
- Have the `PPSDKPaymentMethod` return no visuals and call the error handler on an initialisation strategy mismatch
- Add a `NoUI` component for PPSDK initialisation strategies that require no initial UI (e.g. payment methods which redirect offsite on a user click of "place order")
- Guard all the above with a experiment feature toggle which will not allow the selection of `PPSDKPaymentMethod` when off

## Why?

- To allow for correct rendering of new PPSDK Payment Methods (at first, ones with no visuals)

## Testing / Proof
- New tests

@bigcommerce/checkout

### DIFF
--- a/src/app/payment/paymentMethod/PPSDKPaymentMethod/Wrapper.tsx
+++ b/src/app/payment/paymentMethod/PPSDKPaymentMethod/Wrapper.tsx
@@ -1,0 +1,13 @@
+import React, { useEffect, FunctionComponent } from 'react';
+
+interface Props {
+    onMount(): () => void;
+}
+
+export const Wrapper: FunctionComponent<Props> = props => {
+    const { children, onMount } = props;
+
+    useEffect(onMount, [onMount]);
+
+    return <>{ children }</>;
+};
--- a/package-lock.json
+++ b/package-lock.json
@@ -142,9 +142,9 @@
           "integrity": "sha512-QQmLOGJ3DEgokHbMSA8cj2a+geXqmnpyOFT0lhQV6P3/YOJvGDEwoedcwxEQ30gJIwIIunHIicunJ2rzK5gB2A=="
         },
         "electron-to-chromium": {
-          "version": "1.3.728",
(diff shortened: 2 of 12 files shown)


leeBigCommerce:
Awaiting a bumped SDK version following merge of: https://github.com/bigcommerce/checkout-sdk-js/pull/1139

leeBigCommerce:
> Awaiting a bumped SDK version following merge of: [bigcommerce/checkout-sdk-js#1139](https://github.com/bigcommerce/checkout-sdk-js/pull/1139)

Done

src/app/payment/paymentMethod/PPSDKPaymentMethod/Wrapper.tsx
```diff
@@ -0,0 +1,13 @@
+import React, { useEffect, FunctionComponent } from 'react';
+
+interface Props {
+    onMount(): () => void;
+}
+
+export const Wrapper: FunctionComponent<Props> = props => {
```
capsula4, line 7: might be worth renaming to something more specific, like `PPSDKPaymetWrapper`

capsula4 (approved):
Code looks good. One thing regarding the payment methods that renders no UI, is that they usually modify the text in the submit button. They usually show "Continue with X" rather than "Place Order". More of a UX thing but might be worth keeping it consistent.

leeBigCommerce:
> Code looks good. One thing regarding the payment methods that renders no UI, is that they usually modify the text in the submit button. They usually show "Continue with X" rather than "Place Order". More of a UX thing but might be worth keeping it consistent.

Good point! I've made an additional ticket to cover that off in future work: https://jira.bigcommerce.com/browse/PAYMENTS-6899

PR #590, opened 2021-05-18
feat(payment): INT-4170 added hosted fields on Mollie verification field

JoseAlvarezHdz:
What? INT-4170
Added hosted fields on verification fields on Mollie.

Why?
...

## Testing / Proof
...

@bigcommerce/checkout


JoseBobadilla20:
💚 Tested
QA Details: I tested TSV for Mollie to check works correctly.

mauricio-sg (approved):
LGTM

src/app/payment/paymentMethod/MolliePaymentMethod.tsx
```diff
@@ -66,21 +75,30 @@ const MolliePaymentMethod: FunctionComponent<MolliePaymentMethodsProps> = ({ ini
     }
 
     function isCreditCard(): boolean {
+
         return method.method === MolliePaymentMethodType.creditcard;
     }
 
+    function validateInstrument(_shouldShowNumber: boolean, selectedInstrument: CardInstrument) {
```
capsula4, line 82: the first param doesn't seem to be used? maybe remove?

src/app/payment/paymentMethod/PaymentMethodTitle.tsx
```diff
@@ -123,7 +123,7 @@ function getPaymentMethodTitle(
             },
             [PaymentMethodId.Mollie]: {
                 logoUrl: method.method === 'credit_card' ? '' : cdnPath(`/img/payment-providers/${method.method}.svg`),
-                titleText: startCase(methodName),
+                titleText: method.method === 'credit_card' ? startCase(methodName) : methodName,
```
capsula4, line 126: if the method is not `credit_card`, is it actually wrong to also upper case the first letter?

capsula4 (approved):
Code LGTM, just one minor comment around an unused param

src/app/payment/paymentMethod/MolliePaymentMethod.tsx
JoseAlvarezHdz: its used by Adyen, so thats why we can't remove it!

src/app/payment/paymentMethod/PaymentMethodTitle.tsx
JoseAlvarezHdz: Yes, for example `iDeal` should appear with the `i` as lowercase instead of `Ideal` or `IDeal`

PR #591, opened 2021-05-18
feat(payment): INT-4310 Create HostedDropInPaymentMethod

clopezh:
## What?
Proposal to create a HostedDropInPaymentMethod to support the Digital River Drop In component.

This new PaymentMethod allow us to hide the checkboxes to save cards since the drop in component have them in their own logic.

Also allow us to hide/show the submitPaymentButton depends on the element that is rendered (Drop In or CardInstrumentFieldset) using the paymentContext.

## Why?
Right now there isn't a payment method that supports this type of component and behavior.

## Testing / Proof
Demo: https://drive.google.com/file/d/1ws428vHXhPTeaIqy4PlDdiaxM3_sfLRr/view?usp=sharing

Note: Avoid Circle CI, I didn't complete the unit testing until get approval

@bigcommerce/checkout

### DIFF
--- a/src/app/payment/paymentMethod/HostedDropInPaymentMethod.tsx
+++ b/src/app/payment/paymentMethod/HostedDropInPaymentMethod.tsx
@@ -1,0 +1,388 @@
+import { CardInstrument, CheckoutSelectors, CustomerInitializeOptions,
+    CustomerRequestOptions,
+    Instrument, PaymentInitializeOptions,
+    PaymentInstrument,
+    PaymentMethod, PaymentRequestOptions } from '@bigcommerce/checkout-sdk';
+import { memoizeOne } from '@bigcommerce/memoize';
+import classNames from 'classnames';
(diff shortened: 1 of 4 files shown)


src/app/payment/paymentMethod/HostedDropInPaymentMethod.tsx
```diff
@@ -0,0 +1,385 @@
...
+        const {
+            deinitializeCustomer = noop,
+            deinitializePayment = noop,
+            method,
+            onUnhandledError = noop,
+            setSubmit,
+            setValidationSchema,
+        } = this.props;
+
+        setValidationSchema(method, null);
```
capsula4, line 123: might be worth explaining why we are doing this.

capsula4 (approved):
Looks good to me! 🚢  it

AlanMunozN (approved):
LGTM 👍

PR #593, opened 2021-05-20
feat(payment): INT-4289 Suppress payment methods out of scope on Digital River

juanjosedeleon:
## What?
[INT-4289](https://jira.bigcommerce.com/browse/INT-4289): Suppress payment methods out of scope on Digital River.

## Why?
On BigCommerce we will not be supporting all the alternative [payment methods provided by Digital River](https://jira.bigcommerce.com/browse/INT-4289) by default.

## Testing / Proof
Checkout screen without unsupported methods as Alipay and Wire Transfer:
image


HadesDX (approved):
LGTM

capsula4 (approved):

PR #594, opened 2021-05-24
feat(payment): PAYMENTS-6899 add ppsdk payment method submit button text

leeBigCommerce:
## What?

- Add logic to set the Submit button label to `Continue with {methodName}` when a no UI PPSDK payment method is selected
- Guard the above with a experiment feature toggle which maintains current functionality when off

## Why?
- To convey to consumers that more steps are required with that payment method prior to submitting the order

## Testing / Proof
- New tests

https://user-images.githubusercontent.com/56807262/119301209-7300f380-bca5-11eb-8753-a1604cd8194b.mov

@bigcommerce/checkout

### DIFF
--- a/src/app/payment/PaymentSubmitButton.tsx
+++ b/src/app/payment/PaymentSubmitButton.tsx
@@ -10,9 +10,16 @@ interface PaymentSubmitButtonTextProps {
     methodGateway?: string;
     methodId?: string;
     methodType?: string;
+    continueWith?: string;
+    ppsdkFeatureOn?: boolean;
 }
 
-const PaymentSubmitButtonText: FunctionComponent<PaymentSubmitButtonTextProps> = memo(({ methodId, methodType, methodGateway }) => {
+const PaymentSubmitButtonText: FunctionComponent<PaymentSubmitButtonTextProps> = memo(({ methodId, continueWith, methodType, methodGateway, ppsdkFeatureOn = false }) => {
+
+    if (ppsdkFeatureOn && continueWith) {
+        return <TranslatedString data={ { methodName: continueWith } } id="payment.continue_with_action" />;
+    }
+
     if (methodId === PaymentMethodId.Amazon) {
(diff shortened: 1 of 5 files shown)


src/app/payment/PaymentSubmitButton.tsx
```diff
@@ -98,8 +111,15 @@ export default withCheckout(({ checkoutState }) => {
...
+    const ppsdkFeatureOn = Boolean(
+        checkoutService.getState()
+            .data.getConfig()
+            ?.checkoutSettings.features['PAYMENTS-6806.enable_ppsdk_strategy']
+    );
+
     return {
         isInitializing: isInitializingCustomer() || isInitializingPayment(),
         isSubmitting: isSubmittingOrder(),
+        ppsdkFeatureOn,
```
capsula4, line 123: minor, for consistency with other variables, maybe name it `isPpsdkFeatureEnabled`

src/app/payment/PaymentSubmitButton.tsx
```diff
@@ -10,9 +10,16 @@ interface PaymentSubmitButtonTextProps {
     methodGateway?: string;
     methodId?: string;
     methodType?: string;
+    continueWith?: string;
```
capsula4, line 13: should this be `paymentMethodName` rather than `continueWith` ?

leeBigCommerce: I think I'm mixing existing props that pass only data, with a new prop which is trying to portray behaviour. 

Would we be happier with something more like: https://github.com/bigcommerce/checkout-js/pull/598 ?

leeBigCommerce:
Superseded by: https://github.com/bigcommerce/checkout-js/pull/598

src/app/payment/PaymentSubmitButton.tsx
leeBigCommerce: Moved to: https://github.com/bigcommerce/checkout-js/pull/598

PR #602, opened 2021-05-28
fix(customer): Disable continue as guest button when loading

FranciscoOnt:
## What? [INT-3604](https://jira.bigcommerce.com/browse/INT-3604) & [INT-3605](https://jira.bigcommerce.com/browse/INT-3605)
Pass the `isLoading` variable to the guest component and use it to validate the disabling of 'continue as Guest' button and also the "login" link to prevent it from being clickable when the checkout is loading

## Why?
As mentioned in the Jira Issues, when a customer clicks on the continue as guest button and the google pay / amazon pay scrips haven't finished loading it triggers an error only visible in the developer console which get logged into sentry.

By disabling the buttons until the scripts have finished loading this error is prevented.

## Testing / Proof
[Demo Video](https://drive.google.com/file/d/1dTN99MLiCYULoxVTJvuUQ5M4JeonLnCC/view?usp=sharing)

@bigcommerce/checkout @bigcommerce/payments @bigcommerce/apex-integrations

### DIFF
--- a/src/app/customer/Customer.tsx
+++ b/src/app/customer/Customer.tsx
@@ -38,6 +38,7 @@ export interface WithCheckoutCustomerProps {
     isContinuingAsGuest: boolean;
     isCreatingAccount: boolean;
     isGuestEnabled: boolean;
+    isInitializing: boolean;
     isSendingSignInEmail: boolean;
     isSignInEmailEnabled: boolean;
     isSigningIn: boolean;
@@ -108,6 +109,7 @@ class Customer extends Component<CustomerProps & WithCheckoutCustomerProps, Cust
             email,
             initializeCustomer,
             isContinuingAsGuest = false,
+            isInitializing = false,
             privacyPolicyUrl,
             requiresMarketingConsent,
             onUnhandledError = noop,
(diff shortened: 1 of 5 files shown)


JoseAlvarezHdz (approved):
LGTM!

icatalina (approved):
Is this not a problem when we logging as a registered user? I would assume if we do so quickly (eg. using an autocomplete in a slow connection) we might get the same error.

Not sure if it is possible, but a nicer approach will be to handle the errors in the _deinitialize_ methods correctly 🤷

icatalina (changes requested):
Hi @cmedinilla, I asked a question in my last response. Could you please confirm we've check whether this is an issue for registered users?

FranciscoOnt:
@icatalina I was checking  the registered user scenario and found that this error is also present when the user clicks on the `Already have an account? Sign in now` button, the container disappears and the error is triggered. Let's place this PR on hold while the  new approach is discussed.

icatalina:
thanks for looking into it.

FranciscoOnt:
@icatalina , I've made changes as suggested on the[Jira issue](https://jira.bigcommerce.com/browse/INT-3604), can you please re-review? ♻️  , also is there any other concern with his approach that I should look into?

src/app/customer/GuestForm.tsx
```diff
@@ -32,6 +33,7 @@ const GuestForm: FunctionComponent<GuestFormProps & WithLanguageProps & FormikPr
     canSubscribe,
     checkoutButtons,
     isContinuingAsGuest,
+    isLoading,
```
icatalina, line 36: I wonder if we should just collapse these 2 into `isLoading`. The form doesn't really care why it is loading, just that it is... 😕 what do you think?

cc: @davidchin

src/app/customer/Customer.tsx
```diff
@@ -121,13 +123,15 @@ class Customer extends Component<CustomerProps & WithCheckoutCustomerProps, Cust
...
                         initialize={ initializeCustomer }
+                        isInitializing={ isInitializing }
                         methodIds={ checkoutButtonIds }
                         onError={ onUnhandledError }
                     />
                 }
                 defaultShouldSubscribe={ defaultShouldSubscribe }
                 email={ this.draftEmail || email }
                 isContinuingAsGuest={ isContinuingAsGuest }
+                isLoading={ isInitializing }
```
icatalina, line 134: If we go ahead with the other suggestion and collapse these 2, we can just make the `||` here:

```suggestion
                isLoading={ isContinuingAsGuest || isInitializing }
```

src/app/customer/GuestForm.tsx
FranciscoOnt: sounds good to me, but I would like to hear @davidchin 's opinion.

FranciscoOnt:
@icatalina I made the changes you suggested on the latests commit

icatalina (approved):

PR #608, opened 2021-06-02
feat(payment): INT-3792 Disabled payment method due to settings

clopezh:
## What?
To disable the payment method depending from the settings we need to update the disabledPaymentMethods array that is sending as configuration to the Digital River JS in the strategy.

## Why?
To support the checkboxes in Digital River settings page.

## Testing / Proof
image

## Depends on: https://github.com/bigcommerce/bigcommerce/pull/41232

### DIFF
--- a/src/app/payment/paymentMethod/DigitalRiverPaymentMethod.spec.tsx
+++ b/src/app/payment/paymentMethod/DigitalRiverPaymentMethod.spec.tsx
@@ -73,16 +73,6 @@ describe('when using Digital River payment', () => {
                 disabledPaymentMethods: [
                     'googlePay',
                     'alipay',
-                    'bPay',
-                    'codJapan',
-                    'klarnaCredit',
-                    'konbini',
-                    'msts',
-                    'payco',
-                    'payPal',
-                    'payPalCredit',
-                    'payPalBilling',
-                    'wireTransfer',
                 ],
             },
         };
@@ -107,16 +97,6 @@ describe('when using Digital River payment', () => {
                             disabledPaymentMethods: [
                                 'googlePay',
                                 'alipay',
-                                'bPay',
-                                'codJapan',
-                                'klarnaCredit',
-                                'konbini',
-                                'msts',
-                                'payco',
-                                'payPal',
-                                'payPalCredit',
-                                'payPalBilling',
-                                'wireTransfer',
                             ],
                             classes: {
                                 base: 'form-input optimizedCheckout-form-input',
--- a/src/app/payment/paymentMethod/DigitalRiverPaymentMethod.spec.tsx
+++ b/src/app/payment/paymentMethod/DigitalRiverPaymentMethod.spec.tsx
@@ -60,13 +60,72 @@ describe('when using Digital River payment', () => {
 
         expect(component.props())
             .toEqual(expect.objectContaining({
-                containerId: `${method}-component-field`,
+                containerId: `${method.id}-component-field`,
                 initializePayment: expect.any(Function),
                 method,
             }));
     });
 
     it('initializes method with required config', () => {
+        const methodWithInitializationData = {
+            ...method,
+            initializationData: {
+                disabledPaymentMethods: [
+                    'googlePay',
+                ],
+            },
+        };
+
+        const container = mount(<PaymentMethodTest { ...defaultProps } method={ methodWithInitializationData } />);
+        const component: ReactWrapper<HostedDropInPaymentMethodProps> = container.find(HostedDropInPaymentMethod);
+
+        component.prop('initializePayment')({
+            methodId: method.id,
+            gatewayId: method.gateway,
+        });
+
+        expect(checkoutService.initializePayment)
+            .toHaveBeenCalledWith({
+                digitalriver: {
+                    configuration: {
+                        button: {
+                            type: 'submitOrder',
+                        },
+                        flow: 'checkout',
+                        paymentMethodConfiguration: {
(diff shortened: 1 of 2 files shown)


vmparra (approved):
LGTM 👍🏻

src/app/payment/paymentMethod/DigitalRiverPaymentMethod.tsx
```diff
@@ -21,7 +21,31 @@ const DigitalRiverPaymentMethod: FunctionComponent<DigitalRiverPaymentMethodProp
...
+        const { disabledPaymentMethods } = initializationData;
+
+        if (disabledPaymentMethods) {
+            disabledPaymentMethodsFromSettings = disabledPaymentMethods;
+        }
+    }
+
+    const disabledPaymentMethodsData = [
+        ...disabledPaymentMethodsFromSettings,
+        'alipay',
```
leeBigCommerce, line 37: Why are some disabled methods coming from the API (`rest.method.initializationData.disabledPaymentMethods`) 

while others are hardcoded in here in the view? (`disabledPaymentMethodsData`)

I.e. Why don't they all come from the API?

src/app/payment/paymentMethod/DigitalRiverPaymentMethod.tsx
```diff
@@ -36,19 +60,7 @@ const DigitalRiverPaymentMethod: FunctionComponent<DigitalRiverPaymentMethodProp
...
-                        'klarnaCredit',
-                        'konbini',
-                        'msts',
-                        'payco',
-                        'payPal',
-                        'payPalCredit',
-                        'payPalBilling',
-                        'wireTransfer',
-                    ],
+                    disabledPaymentMethods: disabledPaymentMethodsData,
```
leeBigCommerce, line 63: Do you have a link to some Digital River documentation which explains this `disabledPaymentMethods` option?

I couldn't find it and am interested how it's meant to be used

src/app/payment/paymentMethod/DigitalRiverPaymentMethod.tsx
clopezh: The reason is: 
The list that is hardcoded is the apms those are out of scope. 
The `rest.method.initializationData.disabledPaymentMethods` are the apms that are supported and have checkboxes in DR settings page: 
image

src/app/payment/paymentMethod/DigitalRiverPaymentMethod.tsx
clopezh: There isn't in DR documentation portal, but they told me that you can use this site: https://tools.drapi.io/cm/drop-in/drop-in-builder and it's an option where you can enable or disable payment methods, then the page show you the payload.

src/app/payment/paymentMethod/DigitalRiverPaymentMethod.tsx
leeBigCommerce: Why not hardcode them in the BE?

It would help us keep a more singular source of truth for settings.
 
It would also allow us to simplify this code and stop settings from being kept forever in branches of checkout.

clopezh: Ok I'll move this settings to the BE

clopezh:
@leeBigCommerce  I already updated the array to come from BE, in this PR is the change: https://github.com/bigcommerce/bigcommerce/pull/41232, Could you please take another look? ♻️

src/app/payment/paymentMethod/DigitalRiverPaymentMethod.tsx
```diff
@@ -21,7 +21,17 @@ const DigitalRiverPaymentMethod: FunctionComponent<DigitalRiverPaymentMethodProp
     ...rest
 }) => {
     const { setSubmitted } = useContext(FormContext);
-    const containerId = `${rest.method}-component-field`;
+    const containerId = `${rest.method.id}-component-field`;
+    let disabledPaymentMethodsFromSettings: string[] = [];
```
leeBigCommerce, line 25: Could this now be something like:

`const disabledPaymentMethods = rest.method.initializationData?.disabledPaymentMethods;` ?

The value should end up as your array of strings from the API or undefined, a simpler way of writing lines 26 - 33.

And then you should be able to just pass `disabledPaymentMethods` directly as the useCallback requirement.

src/app/payment/paymentMethod/DigitalRiverPaymentMethod.spec.tsx
```diff
@@ -60,14 +60,34 @@ describe('when using Digital River payment', () => {
...
             }));
     });
 
-    it('initializes method with required config', () => {
-        const container = mount(<PaymentMethodTest { ...defaultProps } method={ method } />);
+    it('initializes method with required config including initializationData', () => {
+        const methodWithInitializationData = {
+            ...method,
+            initializationData: {
+                disabledPaymentMethods: [
```
leeBigCommerce, line 73: Now this data is coming from the BE, the front-end only knows that this is an array of strings - it doesn't care what those strings are. 

So we can simplify our tests to just have shorter representations of the real data.

I.e. this array could just be something like `['some-method', 'another-method']`

leeBigCommerce (approved):
👍

leeBigCommerce:
Could an @bigcommerce/checkout member take a look for us?

icatalina (approved):
Nice 🙇

PR #616, opened 2021-06-18
[draft, wip] feat(payment): BOLT-19 added new BoltCheckout implementation to custo…

Sireks:
…mer step

## What?
Added extra 'sign in', 'sign up' and 'continue as customer' handler to open BoltCheckout modal (with new sdk strategy) before passing customer step.

## Why?
Because of the task: https://jira.bigcommerce.com/browse/BOLT-19

## Sibling pull-request
checkout-sdk-js: https://github.com/bigcommerce/checkout-sdk-js/pull/1158

## Testing / Proof
Some gifs to figure out how it looks like:
continue_as_customer
sign_in
sign_up

@bigcommerce/checkout @BC-ACherednichenko

PR #619, opened 2021-06-24
feat(payment): BOLT-19 added BoltCheckout implementation for customer…

Sireks:
…s with bolt account

## What?
Added BoltCheckout implementation for customers with bolt account.

## Why?
We need to add custom flow for _continue as guest_, _sign in_ and  _sign up_ events, to show BoltCheckout modal if the customer has BoltAccount

Task: https://jira.bigcommerce.com/browse/BOLT-19

## Sibling pull-request
checkout-sdk-js: https://github.com/bigcommerce/checkout-sdk-js/pull/1163

## Testing / Proof
Some gifs to figure out how it looks like:
continue_as_customer
sign_in
sign_up

There how it will looks like if the user have default flow or customer hasn't bolt account:
continue_as_guest_without_bolt_account

### DIFF
--- a/src/app/customer/Customer.tsx
+++ b/src/app/customer/Customer.tsx
@@ -1,8 +1,9 @@
 import { CheckoutSelectors, CustomerAccountRequestBody, CustomerCredentials, CustomerInitializeOptions, CustomerRequestOptions, FormField, GuestCredentials, SignInEmail, StoreConfig } from '@bigcommerce/checkout-sdk';
 import { noop } from 'lodash';
-import React, { Component, Fragment, ReactNode } from 'react';
+import React, { Component, ReactNode } from 'react';
 
 import { withCheckout, CheckoutContextProps } from '../checkout';
+import { LoadingOverlay } from '../ui/loading';
 
 import { CreateAccountFormValues } from './getCreateCustomerValidationSchema';
 import mapCreateAccountFromFormValues from './mapCreateAccountFromFormValues';
@@ -29,6 +30,7 @@ export interface CustomerProps {
 
 export interface WithCheckoutCustomerProps {
     canSubscribe: boolean;
+    customContinueFlowProviderId?: string;
     customerAccountFields: FormField[];
     checkoutButtonIds: string[];
     defaultShouldSubscribe: boolean;
@@ -50,52 +52,83 @@ export interface WithCheckoutCustomerProps {
     createAccountError?: Error;
     signInError?: Error;
     clearError(error: Error): Promise<CheckoutSelectors>;
-    continueAsGuest(credentials: GuestCredentials): Promise<CheckoutSelectors>;
+    continueAsGuest(credentials: GuestCredentials, options: CustomerRequestOptions): Promise<CheckoutSelectors>;
     deinitializeCustomer(options: CustomerRequestOptions): Promise<CheckoutSelectors>;
     initializeCustomer(options: CustomerInitializeOptions): Promise<CheckoutSelectors>;
     sendLoginEmail(params: { email: string }): Promise<CheckoutSelectors>;
-    signIn(credentials: CustomerCredentials): Promise<CheckoutSelectors>;
-    createAccount(values: CustomerAccountRequestBody): Promise<CheckoutSelectors>;
+    signIn(credentials: CustomerCredentials, options: CustomerRequestOptions): Promise<CheckoutSelectors>;
+    createAccount(values: CustomerAccountRequestBody, options: CustomerRequestOptions): Promise<CheckoutSelectors>;
 }
 
 export interface CustomerState {
+    isReady: boolean;
     isEmailLoginFormOpen: boolean;
     hasRequestedLoginEmail: boolean;
 }
 
 class Customer extends Component<CustomerProps & WithCheckoutCustomerProps, CustomerState> {
     state: CustomerState = {
+        isReady: false,
         isEmailLoginFormOpen: false,
         hasRequestedLoginEmail: false,
     };
 
     private draftEmail?: string;
 
-    componentDidMount(): void {
+    async componentDidMount(): Promise<void> {
         const {
-            onReady = noop,
+            customContinueFlowProviderId,
             email,
+            initializeCustomer,
+            onReady = noop,
+            onUnhandledError = noop,
         } = this.props;
 
         this.draftEmail = email;
 
+        try {
+            if (customContinueFlowProviderId) {
+                await initializeCustomer({ methodId: customContinueFlowProviderId });
+            }
+        } catch (error) {
+            onUnhandledError(error);
+        }
+
+        this.setState({ isReady: true });
         onReady();
+    }
+
+    async componentWillUnmount(): Promise<void> {
+        const {
+            customContinueFlowProviderId,
+            deinitializeCustomer,
+            onUnhandledError = noop,
+        } = this.props;
+
+        try {
+            await deinitializeCustomer({methodId: customContinueFlowProviderId});
+        } catch (error) {
+            onUnhandledError(error);
+        }
     }
 
     render(): ReactNode {
         const { viewType } = this.props;
-        const { isEmailLoginFormOpen } = this.state;
+        const { isEmailLoginFormOpen, isReady } = this.state;
         const shouldRenderGuestForm = viewType === CustomerViewType.Guest;
         const shouldRenderCreateAccountForm = viewType === CustomerViewType.CreateAccount;
         const shouldRenderLoginForm = !shouldRenderGuestForm && !shouldRenderCreateAccountForm;
 
         return (
-            <Fragment>
+            <LoadingOverlay
+                isLoading={ !isReady }
+                unmountContentWhenLoading
+            >
                 { isEmailLoginFormOpen && this.renderEmailLoginLinkForm() }
                 { shouldRenderLoginForm && this.renderLoginForm() }
                 { shouldRenderGuestForm && this.renderGuestForm() }
                 { shouldRenderCreateAccountForm && this.renderCreateAccountForm() }
-            </Fragment>
+            </LoadingOverlay>
         );
     }
 
@@ -104,6 +137,7 @@ class Customer extends Component<CustomerProps & WithCheckoutCustomerProps, Cust
             canSubscribe,
             checkEmbeddedSupport,
             checkoutButtonIds,
+            customContinueFlowProviderId,
             defaultShouldSubscribe,
             deinitializeCustomer,
             email,
@@ -114,6 +148,8 @@ class Customer extends Component<CustomerProps & WithCheckoutCustomerProps, Cust
             requiresMarketingConsent,
             onUnhandledError = noop,
         } = this.props;
+
+        const continueAsGuestButtonLabelId = !!customContinueFlowProviderId ? 'customer.continue' : undefined;
 
         return (
             <GuestForm
@@ -128,6 +164,7 @@ class Customer extends Component<CustomerProps & WithCheckoutCustomerProps, Cust
                         onError={ onUnhandledError }
                     />
                 }
+                customContinueAsGuestButtonLabelId={ continueAsGuestButtonLabelId }
                 defaultShouldSubscribe={ defaultShouldSubscribe }
                 email={ this.draftEmail || email }
                 isLoading={ isContinuingAsGuest || isInitializing }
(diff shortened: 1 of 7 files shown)


src/app/locale/translations/en.json
```diff
@@ -100,7 +100,7 @@
         },
         "customer": {
             "checkout_as_guest_text": "Checking out as a <strong>Guest</strong>? You'll be able to save your details to create an account with us later.",
-            "continue_as_guest_action": "Continue as guest",
+            "continue_as_guest_action": "Continue",
```
icatalina, line 103: Why are we changing `Continue as guest` for everyone?

Sireks: Oh, I will fix it

Sireks: Fixed

icatalina (changes requested):
I don't think we should be repurposing _Continue as guest_ as the Bolt strategy. 

I would imagine we need to create a new strategy for these kind of login provider. Also, checkout-js should not be aware this is bolt, all the logic should ideally be abstracted in the SDK. 

Any thoughts @davidchin?

Sireks:
> I don't think we should be repurposing _Continue as guest_ as the Bolt strategy.
> 
> I would imagine we need to create a new strategy for these kind of login provider. Also, checkout-js should not be aware this is bolt, all the logic should ideally be abstracted in the SDK.
> 
> Any thoughts @davidchin?

@icatalina 
Are you talking about this kind of implementation?
https://github.com/bigcommerce/checkout-sdk-js/pull/1158
https://github.com/bigcommerce/checkout-js/pull/616

Sireks:
> I don't think we should be repurposing _Continue as guest_ as the Bolt strategy.
> 
> I would imagine we need to create a new strategy for these kind of login provider. Also, checkout-js should not be aware this is bolt, all the logic should ideally be abstracted in the SDK.
> 
> Any thoughts @davidchin?

Oh, I understood, but the main thing what we have here is that the Bolt wants to have 'one click' implementation to autocomplete all billing, shipping and payments tab using their own checkout. It will be working just for those merchants, who will enable specific Configuration mode in Bolt settings on CP.

It is an abstract functionality, that each provider will have an ability to use for providing their own Checkout, implementing some kind of analytics stuff or some other things.

Here is some tech design: https://github.com/bigcommerce/documentation/pull/394

BC-ACherednichenko:
@icatalina regarding 

> Also, checkout-js should not be aware this is bolt, all the logic should ideally be abstracted in the SDK.

I think this code has represented the misleading idea that provider specific logic will be in checkout-js. However, it is just a temporary approach for defining the whether we have to start strategy for a particular provider or not. 

```js
 const boltPaymentMethodEnabled = paymentMethods?.filter(paymentMethod => paymentMethod.id === 'bolt').length;

should be more something like that in the future, or via the separate repository

```js
const paymentProvidersWithCustomFlowExist = paymentMethods.filter(paymentMethod => paymentMethod?.initializationData?.withCustomContinueFlow).length;

After all a particular strategy will be executed only if we have an available provider that supports "continue with flow"

**Comment by icatalina** [comment]
Hey @Sireks, that's ok. All I'm referring to is we shouldn't be using the `Continue as Guest` for it and rather have a new component in checkout (eg. `CustomContinue`) to deal with the specifics of this type of login flow. 

@BC-ACherednichenko, the "future" approach is ok. Can we change the PR to use that instead?

**Comment by Sireks** [comment]
> Hey @Sireks, that's ok. All I'm referring to is we shouldn't be using the `Continue as Guest` for it and rather have a new component in checkout (eg. `CustomContinue`) to deal with the specifics of this type of login flow.
> 
> @BC-ACherednichenko, the "future" approach is ok. Can we change the PR to use that instead?

About `CustomContinue`. I think that it will be overhead, because we will need to duplicate `continue as guest` logic to make it works as Bolt expected.

**Comment by icatalina** [comment]
> About `CustomContinue`. I think that it will be overhead, because we will need to duplicate `continue as guest` logic to make it works as Bolt expected.

Will an order with Bolt end up as a guest order, or attached to a customer?

The duplication of the code on its own is not reason enough not to split the two. These two components might look the same at the moment, but they seem to represent completely different code paths and they are likely to evolve separately.

I've been pretty detached from checkout recently so I might be wrong. Let's wait for @davidchin to see what he thinks about this.

**Review by vneutrino — APPROVED** [review]
LGTM!

**Review by BC-ACherednichenko — APPROVED** [review]
LGTM

**Comment by Sireks** [comment]
> @icatalina regarding
> 
> > Also, checkout-js should not be aware this is bolt, all the logic should ideally be abstracted in the SDK.
> 
> I think this code has represented the misleading idea that provider specific logic will be in checkout-js. However, it is just a temporary approach for defining the whether we have to start strategy for a particular provider or not.
> 
> ```js
>  const boltPaymentMethodEnabled = paymentMethods?.filter(paymentMethod => paymentMethod.id === 'bolt').length;
> ```
> 
> should be more something like that in the future, or via the separate endpoint that returns such methods
> 
> ```js
> const paymentProvidersWithCustomFlowExist = paymentMethods.filter(paymentMethod => paymentMethod?.initializationData?.withCustomContinueFlow).length;
> ```
> 
> After all a particular strategy will be executed only if we have an available provider that supports "continue with flow"

Thanks! Already changed it from

```js
const boltPaymentMethodEnabled = paymentMethods?.filter(paymentMethod => paymentMethod.id === 'bolt').length;
```
to

```js
const paymentProvidersWithCustomFlowExist = paymentMethods.filter(paymentMethod => paymentMethod?.initializationData?.withCustomContinueFlow).length;
```

src/app/customer/Customer.tsx
```diff
@@ -49,52 +51,90 @@ export interface WithCheckoutCustomerProps {
...
+            loadPaymentMethods,
+            initializeCustomer,
+            onReady = noop,
+            onUnhandledError = noop,
         } = this.props;
 
         this.draftEmail = email;
 
+        try {
+            const { data } = await loadPaymentMethods();
```
davidchin, line 91: I'm not sure if we want to do this any differently than how we handle Masterpass and GooglePay. These remote checkout providers are listed under `config.checkoutSettings.remoteCheckoutProviders` so we don't need to make another call to load all the payment methods enabled for the store.

src/app/customer/Customer.tsx
```diff
@@ -49,52 +51,90 @@ export interface WithCheckoutCustomerProps {
...
+
+            this.providerIdWithCustomContinueFlow = this.getCustomCustomerFlowProviderId(data.getPaymentMethods());
+
+            if (this.providerIdWithCustomContinueFlow) {
```
davidchin, line 95: `initializeCustomer` should be no-op when called with a method that doesn't require initialisation. So you don't need to do this check.

src/app/customer/Customer.tsx
```diff
@@ -49,52 +51,90 @@ export interface WithCheckoutCustomerProps {
...
         onReady();
     }
 
+    async componentWillUnmount(): Promise<void> {
+        const {
+            deinitializeCustomer,
+            onUnhandledError = noop,
+        } = this.props;
+
+        if (this.providerIdWithCustomContinueFlow) {
```
davidchin, line 112: Same here.

src/app/customer/Customer.tsx
```diff
@@ -49,52 +51,90 @@ export interface WithCheckoutCustomerProps {
...
+            onReady = noop,
+            onUnhandledError = noop,
         } = this.props;
 
         this.draftEmail = email;
 
+        try {
+            const { data } = await loadPaymentMethods();
+
+            this.providerIdWithCustomContinueFlow = this.getCustomCustomerFlowProviderId(data.getPaymentMethods());
```
davidchin, line 93: You should avoid setting properties like this directly against the component, if they are referred in the `render` function. This is because setting them won't actually trigger a render. Right now you're relying on a change to another prop to trigger a re-render. Also, you do need it anyway because `paymentMethods` are passed through as props anyway.

src/app/customer/Customer.tsx
```diff
@@ -370,13 +426,19 @@ class Customer extends Component<CustomerProps & WithCheckoutCustomerProps, Cust
 
         onChangeViewType(CustomerViewType.Login);
     };
+
+    private getCustomCustomerFlowProviderId(paymentMethods: PaymentMethod[] = []) {
+        const paymentProvidersWithCustomFlow = paymentMethods.filter(paymentMethod => paymentMethod?.initializationData?.customContinueFlowEnabled);
+
+        return paymentProvidersWithCustomFlow[0]?.id;
```
davidchin, line 433: If there are more than one payment method with `customContinueFlowEnabled`, you just want the first one? I don't think you can make such an assumption.

src/app/customer/Customer.tsx
```diff
@@ -263,11 +307,16 @@ class Customer extends Component<CustomerProps & WithCheckoutCustomerProps, Cust
...
+            const credentials = {
                 email,
                 acceptsMarketingNewsletter: canSubscribe && formValues.shouldSubscribe ? true : undefined,
                 acceptsAbandonedCartEmails: formValues.shouldSubscribe ? true : undefined,
-            });
+            };
+
+            const { data } = await continueAsGuest(
+                credentials,
+                { methodId: this.providerIdWithCustomContinueFlow }
```
davidchin, line 318: Correct me if I'm wrong, if Bolt is enabled for the store, shoppers must use Bolt - there's no choice not to use it? Why do we want to remove that choice. And secondly why does it take priorities over other wallet payment methods that we offer, such as GooglePay, where we offer a choice for shoppers to choose?

cc @bc-charlesho

davidchin (changes requested):
Thank you @Sireks. I apologise for the delay. I think you've made some good progress. I have a question that requires some clarification from product. So I'm going to put a hold on it for now. 🙂

Sireks:
Ok

src/app/customer/Customer.tsx
Sireks: I didn't know that. I will check this out

src/app/customer/Customer.tsx
Sireks: It's a temporary decision. We think to do some functional to get paymentMethod with high priority, because with the implementation we can trigger just one customContinueFlow...

cc @BC-TKozub @BC-ACherednichenko

src/app/customer/Customer.tsx
Sireks: @davidchin shopper can close Bolt modal and continue with default flow.

I can't answer on the second question, @BC-TKozub can you help me with the second one?

fyi @BC-ACherednichenko

src/app/customer/Customer.tsx
Sireks: I will check this one, maybe I don't need the variable anymore

src/app/customer/Customer.tsx
BC-TKozub: @davidchin Only shoppers who have Bolt account will see a pop up on checkout.

src/app/customer/Customer.tsx
Sireks: Maybe we need to add new `config.checkoutSettings.customContinueFlowPovider`?

I need to check, if provider contains `customerContinueFlowEnabled` flag and that it is on.

Sireks: Or I can filter all `remoteCheckoutProviders` (if we will ad bolt there) on `customerContinueFlowEnabled` flag that will not be used in

BC-TKozub: @davidchin @Sireks here is PR ([41748](https://github.com/bigcommerce/bigcommerce/pull/41748)) for this changes.

src/app/customer/Customer.tsx
davidchin: I think this is an important question to answer. With this UX pattern, I think it only makes sense for one payment method that can take over the customer step. The selection of that one payment method should happen on the backend (based on whatever rule we decide). So I think switching over from a collection to a singular property (https://github.com/bigcommerce/bigcommerce/pull/41748) is a step in the right direction.

src/app/customer/Customer.tsx
davidchin: Ok thank you. That makes it _slightly_ better. But there's still a concern where Bolt has priority over other wallet methods as well as interfering with our own guest checkout flow where people are prompted to sign into their BC account if they have one. Anyway this is ultimately a Product question so I'll let Charles to get an alignment with his team.

Sireks:
@davidchin 
I fixed all those things that you tell me to change. Can you check the pull-request one more time, please?

PR #620, opened 2021-06-28
CHECKOUT-5892: Load default translations for various languages

davidchin:
## What?
Load the default translations for the shopper locale of a store. The app reads the global `language` variable, provided by Stencil, to determine which default translations to load and whether there are any theme-level overrides.

To minimise the file size, the translation file for non-English strings is loaded dynamically as a separate chunk in parallel with the rest of the application bundles during the bootstrap process.

## Why?
To improve localisation support in checkout by supplying a set of default translation strings.

### French
fr

### German
de

### Italian
it

### Dutch
nl

### Portuguese
pt

@bigcommerce/checkout @bigcommerce/localization


davidchin:
cc @bigcommerce/payments

src/app/loader.ts
```diff
@@ -56,12 +58,33 @@ export function loadFiles(options?: LoadFilesOptions): Promise<LoadFilesResult>
...
     return Promise.all([
+        getDefaultTranslations(languageConfig.locale),
         scripts,
         stylesheets,
     ])
-        .then(() => {
-            const { renderCheckout, renderOrderConfirmation } = (window as any)[LIBRARY_NAME];
+        .then(([defaultTranslations]) => {
+            // We have to use `any` here because we can't define an interface with a variable defined by Webpack
+            const appExport = (window as any)[LIBRARY_NAME];
```
icatalina, line 72:
```tsx
type KeyInWindow<T extends string> = Record<T, unknown> & Window; 

function isKeyInWindow<T extends string>(win: Window, key: T): win is KeyInWindow<T> {
    return key in win;
}
```

```tsx
if (!isKeyInWindow(window, LIBRARY_NAME)) {
  throw new Error(`App ${LIBRARY_NAME} not available in window`);
}

const appExport = window[LIBRARY_NAME];
```

Considering we, most of the time, just throw an error if the key is not available it might be better to use `asserts` 🤷

src/app/locale/translations/index.ts
```diff
@@ -1,3 +1,3 @@
-import DEFAULT_TRANSLATIONS from './en.json';
+import FALLBACK_TRANSLATIONS from './en.json';
 
-export { DEFAULT_TRANSLATIONS };
+export { FALLBACK_TRANSLATIONS };
```
icatalina, line 3:
```suggestion
export { default as FALLBACK_TRANSLATIONS } from './en.json';
```

src/app/locale/getLanguageService.ts
```diff
@@ -1,13 +1,30 @@
...
-        defaultTranslations: DEFAULT_TRANSLATIONS,
+let languageService: LanguageService | undefined;
+let fallbackLanguageService: LanguageService | undefined;
+
+export default function getLanguageService(): LanguageService {
+    if (languageService) {
+        return languageService;
+    }
+
+    fallbackLanguageService = fallbackLanguageService || createLanguageService({
```
icatalina, line 14: Not much of a difference, but I guess it is better to use `??` 🤷

src/app/locale/getLanguageService.ts
```diff
@@ -1,13 +1,30 @@
...
 
-import { DEFAULT_TRANSLATIONS } from './translations';
+import { FALLBACK_TRANSLATIONS } from './translations';
 
-function getLanguageService(): LanguageService {
-    return createLanguageService({
-        ...(window as any).language,
-        defaultTranslations: DEFAULT_TRANSLATIONS,
+let languageService: LanguageService | undefined;
+let fallbackLanguageService: LanguageService | undefined;
```
icatalina, line 7: Why use 2 copies? Can't we just reuse `languageService` for both?

```tsx
function getLanguageService(): {
  if (!languageService) {
    languageService = createLanguageService({
      defaultTranslations: FALLBACK_TRANSLATIONS,
    });
  }

  return languageService;
}
```

src/app/locale/getLanguageService.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+                'optimized_checkout.greeting': 'fr',
+            },
+            translations: {
+                'optimized_checkout.greeting': 'Bonjour',
+            },
+        });
+
+        expect(getLanguageService().translate('greeting'))
+            .toEqual('Bonjour');
+    });
```
icatalina, line 51: I see no test for when the translations are not initialized. I'm assuming that's what `fallbackLanguageService` is for... 😕

snaderiBC:
I've just noticed from screenshots the word `tax` doesn't seem to be translated 🤔

src/app/locale/getLanguageService.ts
davidchin: No reason. I think I was experimenting and forgot to clean up. 😂

src/app/locale/translations/index.ts
davidchin: I realised I couldn't do this because the JSON is exported as `export =` not as `default`. So I have to import and export in two separate steps.

davidchin:
@snaderiBC I had a quick check and found that 'Tax' is a configurable label that comes from the API (see `taxes` in https://developer.bigcommerce.com/api-reference/storefront/checkouts/checkout/checkoutsbycheckoutidget). Therefore the change introduced by this PR has no effect on it.

davidchin:
I just realised I couldn't do a simple merge with the default and fallback translations because it'd break localised formatting functionality. So I raised a related PR to address that issue: https://github.com/bigcommerce/checkout-sdk-js/pull/1166

icatalina (approved):

PR #624, opened 2021-07-01
CHECKOUT-5872: Handle locale with region identifier

davidchin:
## What?
Ignore the region identifier if only the language identifier matches with the value inherited from storefront.

## Why?
For example, if `fr-CA` is the preferred locale, display `fr` instead of `en`.

## Testing / Proof
Unit

@bigcommerce/checkout

### DIFF
--- a/src/app/locale/getDefaultTranslations.ts
+++ b/src/app/locale/getDefaultTranslations.ts
@@ -2,56 +2,42 @@ import { Translations } from '@bigcommerce/checkout-sdk';
 
 import { FALLBACK_TRANSLATIONS } from './translations';
 
-export default async function getDefaultTranslations(locale: string): Promise<Translations> {
-    switch (locale) {
-        case 'de':
-            return asTranslations((await import(
-                /* webpackChunkName: "translations-de" */
-                './translations/de.json'
-            )).default);
+const AVAILABLE_TRANSLATIONS: Record<string, () => Promise<{ default: unknown }>> = {
+    de: () => import(
+        /* webpackChunkName: "translations-de" */
+        './translations/de.json'
+    ),
+    fr: () => import(
+        /* webpackChunkName: "translations-fr" */
+        './translations/fr.json'
+    ),
+    it: () => import(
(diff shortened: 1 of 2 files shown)


src/app/locale/getDefaultTranslations.ts
```diff
@@ -3,55 +3,60 @@ import { Translations } from '@bigcommerce/checkout-sdk';
...
-            )).default);
-
-        case 'en':
-            return FALLBACK_TRANSLATIONS;
-
-        default:
-            return {};
+    if (matchLocale(locale, 'de')) {
+        return asTranslations((await import(
+            /* webpackChunkName: "translations-de" */
```
valentindellangela, line 8: 😮  Do you need all those comments?

src/app/locale/getDefaultTranslations.ts
```diff
@@ -3,55 +3,60 @@ import { Translations } from '@bigcommerce/checkout-sdk';
...
+            /* webpackChunkName: "translations-sv" */
+            './translations/sv.json'
+        )).default);
+    }
+
+    if (matchLocale(locale, 'en')) {
+        return FALLBACK_TRANSLATIONS;
+    }
+
+    return {};
```
valentindellangela, line 59: 🤔  No default translation?

src/app/locale/getDefaultTranslations.ts
davidchin: Yes, this is to tell Webpack how to name these chunks when they are compiled into JS, i.e.: https://webpack.js.org/api/module-methods/#magic-comments

src/app/locale/getDefaultTranslations.ts
davidchin: Yes, so I made a distinction between default translations and fallback translations. Fallback translations always exist and in English. Default translations may not exist if there are no default translations matching the specified locale.

valentindellangela (approved):
LGTM 👍

icatalina (approved):

icatalina (approved):

PR #628, opened 2021-07-07
fix(shipping): CHECKOUT-5895 Keep billingSameAsShipping state when navigating through steps

RobinOng:
## What?
When navigating to billing/payment step and back to shipping step, the "My billing address is the same as my shipping address" checkbox always default to `true` 

## Why?
The state of the checkbox is not saved

## Testing / Proof

https://user-images.githubusercontent.com/22089936/124700192-e17cd680-df1e-11eb-818a-e58fbe3d8233.mov

@bigcommerce/checkout

### DIFF
--- a/src/app/checkout/Checkout.tsx
+++ b/src/app/checkout/Checkout.tsx
@@ -66,6 +66,7 @@ export interface CheckoutProps {
 
 export interface CheckoutState {
     activeStepType?: CheckoutStepType;
+    isBillingSameAsShipping: boolean;
     customerViewType?: CustomerViewType;
     defaultStepType?: CheckoutStepType;
     error?: Error;
@@ -100,6 +101,7 @@ class Checkout extends Component<CheckoutProps & WithCheckoutProps & WithLanguag
     stepTracker: StepTracker | undefined;
 
     state: CheckoutState = {
+        isBillingSameAsShipping: true,
         isCartEmpty: false,
         isRedirecting: false,
         isMultiShippingMode: false,
@@ -319,7 +321,10 @@ class Checkout extends Component<CheckoutProps & WithCheckoutProps & WithLanguag
             consignments = [],
         } = this.props;
 
-        const { isMultiShippingMode } = this.state;
+        const {
+            isBillingSameAsShipping,
+            isMultiShippingMode,
+        } = this.state;
 
         if (!cart) {
             return;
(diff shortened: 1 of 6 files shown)


src/app/checkout/Checkout.tsx
```diff
@@ -100,6 +101,7 @@ class Checkout extends Component<CheckoutProps & WithCheckoutProps & WithLanguag
     stepTracker: StepTracker | undefined;
 
     state: CheckoutState = {
+        billingSameAsShipping: true,
```
animesh1987, line 104: Is that the default behaviour we want to go to?

src/app/checkout/Checkout.tsx
```diff
@@ -66,6 +66,7 @@ export interface CheckoutProps {
 
 export interface CheckoutState {
     activeStepType?: CheckoutStepType;
+    billingSameAsShipping: boolean;
```
animesh1987, line 69: Should this have the `is` prefix as per other booleans in this interface?

src/app/checkout/Checkout.tsx
```diff
@@ -343,6 +348,7 @@ class Checkout extends Component<CheckoutProps & WithCheckoutProps & WithLanguag
             >
                 <LazyContainer>
                     <Shipping
+                        billingSameAsShipping={ billingSameAsShipping }
```
animesh1987, line 351: Same as above

src/app/checkout/Checkout.tsx
RobinOng: Good idea. I've added `is` prefix to it

src/app/checkout/Checkout.tsx
RobinOng: Yeah, it was the behaviour before the change as well. The purpose of this is to reduce friction in checkout

animesh1987: Thanks for the explanation 👍

animesh1987 (approved):

animesh1987:
The code looks good, there seem to be a few lint errors that circle is failing on.

src/app/shipping/BillingSameAsShippingField.tsx
```diff
@@ -17,7 +17,7 @@ const BillingSameAsShippingField: FunctionComponent<BillingSameAsShippingFieldPr
     return <CheckboxFormField
         id="sameAsBilling"
         labelContent={ labelContent }
-        name="billingSameAsShipping"
+        name="isBillingSameAsShipping"
```
davidchin, line 20: Is this change intentional? I believe it's going break any customisation that targets `[name=billingSameAsShipping]` so it's better not to change it. You can keep the name of the variable as `isBillingSameAsShipping` in other places, but not this one as it can affect the HTML output.

RobinOng: Good point. I've changed this back

davidchin (approved):
LGTM!

PR #631, opened 2021-07-08
feat(payment): PAYMENTS-6812 add PPSDK payment methods to the handleBeforeUnload exemption list

leeBigCommerce:
## What?

- Add PPSDK methods to the before unload warning exemption list

## Why?

- So that redirect flows are not interrupted by the "Are you sure?" alert

## Testing / Proof

**Before**

image

**After**

https://user-images.githubusercontent.com/56807262/124855426-43efd880-dfec-11eb-8dbd-61e9998becaa.mov

@bigcommerce/checkout

### DIFF
--- a/src/app/payment/Payment.tsx
+++ b/src/app/payment/Payment.tsx
@@ -270,6 +270,7 @@ class Payment extends Component<PaymentProps & WithCheckoutPaymentProps & WithLa
         });
     };
 
+    // tslint:disable:cyclomatic-complexity
     private handleBeforeUnload: (event: BeforeUnloadEvent) => string | undefined = event => {
         const { defaultMethod, isSubmittingOrder, language } = this.props;
         const { selectedMethod = defaultMethod } = this.state;
@@ -281,6 +282,7 @@ class Payment extends Component<PaymentProps & WithCheckoutPaymentProps & WithLa
         if (!isSubmittingOrder ||
             !selectedMethod ||
             selectedMethod.type === PaymentMethodProviderType.Hosted ||
+            selectedMethod.type === PaymentMethodProviderType.PPSDK ||
             selectedMethod.id === PaymentMethodId.Amazon ||
             selectedMethod.id === PaymentMethodId.AmazonPay ||
             selectedMethod.id === PaymentMethodId.Checkoutcom ||


leeBigCommerce:
Not proud of the use of `tslint:disable:cyclomatic-complexity`, my addition to the list seems to be the last straw.

I had attempted something a little more clean (https://github.com/bigcommerce/checkout-js/pull/621/files), but this too seems like treating the symptom rather than the cause.

We seem to have both `checkout-sdk` and `checkout-js` wanting to handle navigation - the former redirecting, the latter guarding  page unloads. We also have many payment methods, not of type `PaymentMethodProviderType.Hosted`, using redirects.

I'd recommend either we centralise 'navigation' concerns solely in either `checkout-sdk` or `checkout-js`. Or, if that isn't viable for some reason, consider extending our Payment Method type to have a kind of `navigatesOnly` attribute to indicate more explicitly what `PaymentMethodProviderType.Hosted` once inferred. Or something better 😄 .

Either way, such a solution would have wider repercussions than felt right to draw into this current ticket, but is something we should look into sooner than later.

src/app/payment/Payment.tsx
```diff
@@ -281,6 +282,7 @@ class Payment extends Component<PaymentProps & WithCheckoutPaymentProps & WithLa
         if (!isSubmittingOrder ||
             !selectedMethod ||
             selectedMethod.type === PaymentMethodProviderType.Hosted ||
+            selectedMethod.type === PaymentMethodProviderType.PPSDK ||
```
davidchin, line 285: So all of these payment methods are hosted payment methods. The only reason why they are listed is because they are incorrectly returned as "credit card" methods instead, due to the way they are implemented by the integration team. As far as I can tell, there's no real reason why they can't return the correct type, except that it involves a bit of work.

Anyway, what you have is fine I suppose... but I believe not all PPSDK methods require a redirect. So having this line means the alert won't show for credit card PPSDK methods. But you can't really tell if a PPSDK method requires a redirect because that information comes from the "steps". For example, regular CC should prompt an alert but if it triggers a 3DS challenge it shouldn't.

Anyway, I think for now this okay, at least you can test the flow. But as you said, we should put some thoughts into this - perhaps moving the `beforeunload` handler into the SDK, which can make more granular decisions.

davidchin (approved):

src/app/payment/Payment.tsx
leeBigCommerce: Yep, I know what you mean about the non-redirecting PPSDK methods - I figured we'd live without the potentially useful unload warning until a proper solution surfaces.

We could turn them all into `PaymentMethodProviderType.Hosted` (and find a different attribute to use to pivot between PPSDK/non-PPSDK), but I think centralising 'navigation' concerns might be better.

That if the SDK does the `beforeunload` _and_ the redirecting, it can more elegantly skip the `beforeunload` as needed.

PR #636, opened 2021-07-12
fix(checkout): INT-4504 Fix 404 on mandate link for SEPA

FranciscoOnt:
## What?
Render Mandate as plaint text for SEPA on Checkout.com

## Why?
currently SEPA on Checkout.com returns a Mandate Reference only which is treated as a link which causes a 404 error when clicking on it.

## Testing / Proof
...

### DIFF
--- a/src/app/order/OrderStatus.tsx
+++ b/src/app/order/OrderStatus.tsx
@@ -11,6 +11,8 @@ export interface OrderStatusProps {
     order: Order;
 }
 
+const checkoutcomSEPAMethod = 'SEPA Direct Debit (via Checkout.com)';
+
 const OrderStatus: FunctionComponent<OrderStatusProps> = ({
     order,
     supportEmail,
@@ -26,11 +28,22 @@ const OrderStatus: FunctionComponent<OrderStatusProps> = ({
             { method: 'Stripe (SEPA)', value: 'sepa_link_text' },
             { method: 'OXXO (via Checkout.com)', value: 'oxxo_link_text' },
             { method: 'Boleto Bancário (via Checkout.com)', value: 'boleto_link_text' },
+            { method: checkoutcomSEPAMethod, value: 'mandate_text_only' },
         ];
 
         const mandateText = Mandates.find(pair => pair.method === order?.payments?.[0].description);
 
         return mandateText ? mandateText.value : 'mandate_link_text';
+    }, [order]);
+
+    const showMandateAsTextOnly = useCallback(() => {
+        const methods = [
+            { method: checkoutcomSEPAMethod },
+        ];
+
+        const mandateText = methods.find(pair => pair.method === order?.payments?.[0].description);
+
+        return !!mandateText;
     }, [order]);
 
     return <OrderConfirmationSection>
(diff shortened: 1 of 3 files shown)


roberbnd (approved):
LGTM

src/app/order/OrderStatus.tsx
```diff
@@ -26,13 +28,26 @@ const OrderStatus: FunctionComponent<OrderStatusProps> = ({
...
             { method: 'Boleto Bancário (via Checkout.com)', value: 'boleto_link_text' },
+            { method: checkoutcomSEPAMethod, value: 'mandate_text_only' },
         ];
 
         const mandateText = Mandates.find(pair => pair.method === order?.payments?.[0].description);
 
         return mandateText ? mandateText.value : 'mandate_link_text';
     }, [order]);
 
+    const showMandateAsTextOnly = useCallback(() => {
```
leeBigCommerce, line 39: What's the benefit of deferring the execution of this function?

We only seem to use it once and within the same scope as its creation.

`textOnlyMethods` seems to be constant, so we can move it out of this FC as you've done with `checkoutcomSEPAMethod`.

`order?.payments?.[0].description` seems to be used various other places within this component, so we could build a shared reference to it:

```
const mandateProvider = `${order?.payments?.[0].description}`
```

which all should let us simplify this function to just:

```
const showMandateAsTextOnly = TEXT_ONLY_METHODS.includes(mandateProvider);
```

src/app/order/OrderStatus.tsx
```diff
@@ -51,12 +66,20 @@ const OrderStatus: FunctionComponent<OrderStatusProps> = ({
             />
         </p>
 
-        { order.mandateUrl && <a data-test="order-confirmation-mandate-link-text" href={ order.mandateUrl } rel="noopener noreferrer" target="_blank">
+        { showMandateAsTextOnly() ?
```
leeBigCommerce, line 69: Inline turnaries like this can be pretty hard to read, especially as it contains two conditionals.

It's better to split out the sections of JSX into their own components, e.g.

```
interface MandateTextProps { provider: string; mandate?: string; id: string; }

const MandateText = ({ provider, mandate = '', id }: MandateTextProps) => (
    <div data-test="order-confirmation-mandate-text-only">
        <br />
        <TranslatedString
            data={ { provider, mandate } }
            id={ 'order_confirmation.' + id }
        />
    </div>
);
```

Then this turnary becomes something like:

```
        { showMandateAsTextOnly || !order.mandateUrl ?
            <MandateText id={ mandateTextId } mandate={ order.mandateUrl } provider={ mandateProvider } /> :
            <MandateLink id={ mandateTextId } mandate={ order.mandateUrl } provider={ mandateProvider } /> }
```

By splitting out the defining of the JSX, we can just concentrate on what will be executed given the different scenarios.

src/app/order/OrderStatus.tsx
```diff
@@ -111,4 +113,29 @@ const OrderStatusMessage: FunctionComponent<OrderStatusMessageProps> = ({
...
+    <div data-test="order-confirmation-mandate-text-only">
+        <br />
+        <TranslatedString
+            data={ { provider, mandate } }
+            id={ 'order_confirmation.' + id }
+        />
+    </div>
+);
+
+const MandateLink = ({ provider, id, mandate = '' }: MandateTextProps) => (
```
leeBigCommerce, line 132: Do we ever want `mandate` to be an empty string? This could lead to a link to `''`

src/app/order/OrderStatus.tsx
```diff
@@ -11,27 +11,32 @@ export interface OrderStatusProps {
     order: Order;
 }
 
+const checkoutcomSEPAMethod = 'SEPA Direct Debit (via Checkout.com)';
```
leeBigCommerce, line 14: As this is also a constant config value, we can use the same all capitals, snake case as `TEXT_ONLY_METHODS` to better convey its purpose.

leeBigCommerce:
Sorry, just realised that we're treating the value `order.mandateUrl` sometimes as a url, sometimes as a string.

Shouldn't we fix it so that the BE only returns urls in the `mandateUrl` attribute? 

If we want to display something along the lines of a mandate ID or reference - would it make sense to create a sibling attribute instead?

cmedinilla:
hi @leeBigCommerce we need to release SEPA on checkout.com this month, can we use this for now and create a ticket to refactor this after launch?

leeBigCommerce:
@cmedinilla Ideally we wouldn't release `checkout-js` with this kind of bending of APIs. Once it is released and forked, consumers may come to rely on this odd, dual behaviour of the `mandateUrl` field.

@davidchin @animesh1987 ?

animesh1987:
Hi @cmedinilla, agreed with @leeBigCommerce above. We should try and expose a new property `mandateText` and use that.

PR #638, opened 2021-07-13
feat(payment): INT-3946 added locale from browser on masterpass SRC

vmparra:
## What? [INT-3946](https://jira.bigcommerce.com/browse/INT-3946)
Added locale from browser on masterpass SRC.

## Why?
In order to be able to see masterpass or SRC button based on shopper locale.

## Testing / Proof
Screen Shot 2021-06-30 at 6 53 52 PM

@bigcommerce/checkout

### DIFF
--- a/src/app/payment/paymentMethod/MasterpassPaymentMethod.tsx
+++ b/src/app/payment/paymentMethod/MasterpassPaymentMethod.tsx
@@ -1,4 +1,5 @@
 import { PaymentInitializeOptions } from '@bigcommerce/checkout-sdk';
+import { getUserLocale } from 'get-user-locale';
 import React, { useCallback, useMemo, FunctionComponent } from 'react';
 import { Omit } from 'utility-types';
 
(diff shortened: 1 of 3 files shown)


src/app/payment/paymentMethod/MasterpassPaymentMethod.tsx
```diff
@@ -20,8 +20,17 @@ const MasterpassPaymentMethod: FunctionComponent<MasterpassPaymentMethodProps &
         },
     }), [initializePayment]);
 
+    const formatLocale = useCallback((localeLanguage: string) => {
```
animesh1987, line 23: This looks more like a util method, can it be moved outside the component?

TomA-R:
AFAIK we've recently changed our approach of determining a user's expected locale. This is now based on a store setting, so I imagine we should use the same mechanism here. Otherwise shoppers might get an experience where their checkout is presented in Dutch (because of the store setting), but the master button is in French, because the browser's Accept-Language is set to that.

I'll have to defer to the checkout team on _how_ to actually access this from within the checkout, cc @davidchin

src/app/payment/paymentMethod/MasterpassPaymentMethod.tsx
vmparra: I switch approach to consume a function that returns user's locale as an IETF language tag, based on all available sources.

src/app/payment/paymentMethod/MasterpassPaymentMethod.tsx
```diff
@@ -20,17 +21,19 @@ const MasterpassPaymentMethod: FunctionComponent<MasterpassPaymentMethodProps &
         },
     }), [initializePayment]);
 
+    const userLocale = getUserLocale();
```
davidchin, line 24: Yes as, @TomA-R said, can you follow the advice I gave to @BC-AxelMalagon in https://github.com/bigcommerce/checkout-sdk-js/pull/1168#discussion_r661892982. For Checkout JS, what you need to do here is to use to simply call `language.getLocale()`. `language` instance is already in the scope of the function so you can just use it. I believe you still need to transform the value somehow in the format Masterpass expects (i.e.: `_` instead of `-`)

davidchin (changes requested):

PR #647, opened 2021-07-22
fix(checkout): CUST-1572 accept marketing emails should be true when requires marketing consent is false

mjhmatt:
## What?
If the shopper doesn't require marketing consent, when a customer is created on checkout we should send the value to receive marketing emails

## Why?
Currently, that value is always false so even if the setting is off, customers won't receive marketing emails

## Testing / Proof
Update test 

@bigcommerce/checkout
### Require consent is false
Screen Shot 2021-07-30 at 1 59 47 pm
Screen Shot 2021-07-30 at 1 59 05 pm
### Require consent is true
image
Screen Shot 2021-07-30 at 2 01 37 pm

### DIFF
--- a/src/app/customer/CreateAccountForm.tsx
+++ b/src/app/customer/CreateAccountForm.tsx
@@ -18,6 +18,7 @@ export interface CreateAccountFormProps {
     formFields: FormField[];
     createAccountError?: Error;
     isCreatingAccount?: boolean;
+    requiresMarketingConsent: boolean;
     onCancel?(): void;
     onSubmit?(values: CreateAccountFormValues): void;
 }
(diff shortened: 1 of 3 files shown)


RobinOng:
Overall LGTM. I think the test to check accept marketing emails is true when requires marketing consent is false is missing and also a gif/screenshot that it works?

src/app/customer/getCreateCustomerValidationSchema.ts
```diff
@@ -62,6 +63,7 @@ export default memoize(function getCreateCustomerValidationSchema({
                 .matches(alpha, description || language.translate('customer.password_letter_required_error'))
                 .min(minLength, description || language.translate('customer.password_under_minimum_length_error'))
                 .max(100, language.translate('customer.password_over_maximum_length_error')),
+            requiresMarketingConsent: boolean().required(),
```
davidchin, line 66: Not sure what you're trying to do here but `requiresMarketingConsent` is not the input that needs validating.

src/app/customer/CreateAccountForm.tsx
```diff
@@ -103,13 +104,14 @@ export default withLanguage(withFormik<CreateAccountFormProps & WithLanguageProp
...
         onSubmit(values);
     },
-    mapPropsToValues: () => ({
+    mapPropsToValues: ({requiresMarketingConsent}) => ({
         firstName: '',
         lastName: '',
         email: '',
         password: '',
         customFields: {},
         acceptsMarketingEmails: [],
```
davidchin, line 113: I think what you're looking for is to provide a different default value for `acceptsMarketingEmails` if `requiresMarketingConsent` is `false`.

animesh1987 (approved):

davidchin (approved):

PR #651, opened 2021-07-26
fix(payment): INT-3946 using supported locales on masterpass SRC

BC-PabloRivera:
## What? [INT-3946](https://jira.bigcommerce.com/browse/INT-3946)
Added locale from store config and validate that the locale is supported
(Using INT-3408.display_SRC_masterpass_button experiment)

## Why?
In order to be able to see masterpass or SRC button based on merchant settings.

## Testing / Proof
Before, with unsupported locale
image

Now
image

## Depends on
https://github.com/bigcommerce/checkout-sdk-js/pull/1197

@bigcommerce/checkout


vmparra (approved):
Hi @BC-PabloRivera, looking good. Perhaps a few test cases?

vmparra (changes requested):
Just left a comment.

src/app/payment/paymentMethod/MasterpassPaymentMethod.tsx
```diff
@@ -20,8 +20,43 @@ const MasterpassPaymentMethod: FunctionComponent<MasterpassPaymentMethodProps &
         },
     }), [initializePayment]);
 
+    const formatLocale = useCallback((localeLanguage: string) => {
+        const supportedLocales: {[language: string]: string[]} = {es: ['es_es', 'es_mx', 'es_pe', 'es_co', 'es_ar', 'es_cl'],
+            en: ['en_us', 'en_gb', 'en_ca', 'en_es', 'en_fr', 'en_ie', 'en_sg', 'en_au', 'en_nz', 'en_my', 'en_hk', 'en_th', 'en_ae', 'en_sa', 'en_qa', 'en_kw', 'en_za'],
```
JoseAlvarezHdz, line 25: I dont get it, why are you using the same function in SDK and in JS? i think if you add it to the SDK then you could import it here and re use it

JoseAlvarezHdz (changes requested):

BC-PabloRivera:
@vmparra @JoseAlvarezHdz Can you check the changes? ♻️

src/app/locale/masterpassFormatLocale.ts
```diff
@@ -0,0 +1,32 @@
...
+            it: ['it_it'],
+            de: ['de_de'],
+            fr: ['fr_fr', 'fr_ca']};
+    let formatedLocale = localeLanguage.replace('-', '_').toLowerCase();
+    const regexLocale = formatedLocale.match(/^([a-z]{2})((?:\_)([a-z]{2}))?$/);
+    if (regexLocale && regexLocale[3]) {
+        if (regexLocale[1] in supportedLocales) {
+            formatedLocale = supportedLocales[regexLocale[1]].includes(regexLocale[0]) ? formatedLocale : supportedLocales[regexLocale[1]][0];
+        } else {
+            formatedLocale = 'en_us';
```
JoseAlvarezHdz, line 23: i'll suggest only the same change you did in SDK

JoseAlvarezHdz (approved):
LGTM 👍🏾

vmparra (approved):
LGTM 👍🏻

src/app/locale/masterpassFormatLocale.ts
BC-PabloRivera: Updated

JelopezBC:
💚 Tested!

PR has been validated, masterpass is working as expected:

Tested scenarios:

- Orders with Masterpass
- Orders with Auth Only and "Enable receipt email" toggle button is enable
- Orders with Auth & Capture and "Enable receipt email" toggle button is enable
- Different credit cards (VISA, MC, Discover, etc).
- 3DS doesn't applies with wallets
- Etc.

Please review attached email with details.
[INT-3946 Stripe V3 Masterpass.docx](https://github.com/bigcommerce/checkout-js/files/6925304/INT-3946.Stripe.V3.Masterpass.docx)

src/app/locale/masterpass-supported-locales.ts
```diff
@@ -0,0 +1,15 @@
+export const supportedLocales: {[language: string]: string[]} = {es: ['es_es', 'es_mx', 'es_pe', 'es_co', 'es_ar', 'es_cl'],
```
leeBigCommerce, line 1: Why are we validating the locales against this list rather than just formatting it and letting the Masterpass domain decide what to do?

The domain seems to have logic to handle non supported locales:

https://sandbox.src.mastercard.com/assets/img/btn/src_chk_btn_126x030px.svg?paymentmethod=master,visa,amex,discover&locale=xx_xx

Whereas we seem to be trying to force a defaulting to 'en_us'

src/app/payment/paymentMethod/MasterpassPaymentMethod.tsx
```diff
@@ -20,8 +20,14 @@ const MasterpassPaymentMethod: FunctionComponent<MasterpassPaymentMethodProps &
         },
     }), [initializePayment]);
 
+    const formatLocale = useCallback((localeLanguage: string) => {
```
leeBigCommerce, line 23: `useCallback` isn't required for functions imported from modules.

Typically we only need it when functions are going in or out of props, and only then to preserve reference equality: https://reactjs.org/docs/hooks-reference.html#usecallback

src/app/locale/masterpass-supported-locales.ts
BC-PabloRivera: We validate the locales based on the masterpass supported locale list so it always shows the new button.
We use getLanguageService().getLocale() to fetch the locale, but this gives back a language only locale like 'en', so we need to validate it against a list to get a language_country locale.
I can change this to get the locale from the store profile, which gives back a language_country locale

BC-PabloRivera:
@leeBigCommerce I did some changes, can you check them out? ♻️

src/app/locale/masterpass-supported-locales.ts
leeBigCommerce: I wasn't suggest that we change the source of the locale, more questioning the need to filter it.

Do we want the button's display based on the store's locale or the customer's locale?

If we _do_ want the customers I think we'll need to read it from `window.navigator.language`, though I'm not sure about browser support or formatting gotchas, so maybe look into that if it is needed.

BC-PabloRivera: We want to show it based on the store locale

leeBigCommerce (approved):

leeBigCommerce:
@bigcommerce/checkout could a code owner please take a look for us 🙇

animesh1987 (approved):

PR #663, opened 2021-08-06
fix(payment): INT-4709 Display "place order button" when payment is not required

AlanMunozN:
## What? [INT-4709](https://jira.bigcommerce.com/browse/INT-4709)
Fix logic to render on submit button when a discount is applied and the total is 0.

## Why?
we detected this scenario in our QE phase and was necessary to fix it.

## Testing / Proof
**BEFORE:**
Screen Shot 2021-08-06 at 10 26 27 AM
**AFTER:**
Screen Shot 2021-08-06 at 10 29 40 AM


clopezh (approved):
LGTM 👍🏻

davidchin (approved):

TomA-R (approved):
Nice job!

PR #674, opened 2021-08-20
feat(order): INT-4776 Create a new field for the mandate reference ID

FranciscoOnt:
## What?
Display mandate reference ID  as text only when mandate url is empty.

Also remove the part of the code that searched for the mandate text and used it to build the mandate text, instead use the new method Id to use it as a key and map it to the correct mandate text.

## Why?
So the mandate link does not redirect to a 404 page.

Also the translation of the mandate text is done properly and not built into the code.

## Testing / Proof
[Demo Videos](https://drive.google.com/drive/folders/1VcyX_5nMpL3Ri0nMPOcgDqWWBq0Kmzvx?usp=sharing)
Screen Shot 2021-08-23 at 3 41 02 PM

## Sibling PRs
https://github.com/bigcommerce/bigcommerce/pull/42364
https://github.com/bigcommerce/bigpay/pull/4234

https://github.com/bigcommerce/checkout-sdk-js/pull/1220

## Merge Order
1) https://github.com/bigcommerce/checkout-sdk-js/pull/1220

2) https://github.com/bigcommerce/bigcommerce/pull/42364
https://github.com/bigcommerce/bigpay/pull/4234
https://github.com/bigcommerce/checkout-js/pull/674

### DIFF
--- a/src/app/order/OrderStatus.tsx
+++ b/src/app/order/OrderStatus.tsx
@@ -1,5 +1,5 @@
-import { Order } from '@bigcommerce/checkout-sdk';
-import React, { memo, useCallback, FunctionComponent } from 'react';
+import { GatewayOrderPayment, GiftCertificateOrderPayment, Order } from '@bigcommerce/checkout-sdk';
+import React, { memo, FunctionComponent } from 'react';
 
 import { TranslatedHtml, TranslatedString } from '../locale';
 
@@ -11,29 +11,18 @@ export interface OrderStatusProps {
     order: Order;
 }
 
+type PaymentWithMandate = GatewayOrderPayment & Required<Pick<GatewayOrderPayment, 'mandate' | 'methodId'>>;
+
+const isPaymentWithMandate = (payment: GatewayOrderPayment | GiftCertificateOrderPayment): payment is PaymentWithMandate =>
+    !!payment.methodId && ('mandate' in payment && !!payment.mandate);
+
 const OrderStatus: FunctionComponent<OrderStatusProps> = ({
     order,
     supportEmail,
     supportPhoneNumber,
 }) => {
 
-    const getMandateProvider = useCallback(() => {
-        return order?.payments?.[0].description;
-    }, [order]);
-
-    const getMandateTextId = useCallback(() => {
-        const Mandates = [
-            { method: 'Stripe (SEPA)', value: 'sepa_link_text' },
-            { method: 'OXXO (via Checkout.com)', value: 'oxxo_link_text' },
-            { method: 'OXXO (via Stripe)', value: 'oxxo_link_text'},
-            { method: 'Boleto Bancário (via Checkout.com)', value: 'boleto_link_text' },
-            { method: 'Boleto (via Stripe)', value: 'boleto_link_text'},
-        ];
-
-        const mandateText = Mandates.find(pair => pair.method === order?.payments?.[0].description);
-
-        return mandateText ? mandateText.value : 'mandate_link_text';
-    }, [order]);
+    const paymentsWithMandates = order.payments?.filter(isPaymentWithMandate) || [];
 
     return <OrderConfirmationSection>
         { order.orderId &&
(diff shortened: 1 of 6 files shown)


JoseAlvarezHdz (approved):
LGTM 👍🏾

vmparra:
@funivan Please make sure you validate this PR with AdyenV2 also.

src/app/order/OrderStatus.tsx
```diff
@@ -17,21 +17,20 @@ const OrderStatus: FunctionComponent<OrderStatusProps> = ({
     supportPhoneNumber,
 }) => {
 
-    const getMandateProvider = useCallback(() => {
-        return order?.payments?.[0].description;
-    }, [order]);
+    const mandateProvider = `${order?.payments?.[0].description}`;
```
leeBigCommerce, line 20: Why are we picking the first rather than looping?

src/app/order/OrderStatus.tsx
```diff
@@ -17,21 +17,17 @@ const OrderStatus: FunctionComponent<OrderStatusProps> = ({
...
+    let mandateMethod = '';
+    let orderMandate;
 
-    const getMandateTextId = useCallback(() => {
-        const Mandates = [
-            { method: 'Stripe (SEPA)', value: 'sepa_link_text' },
-            { method: 'OXXO (via Checkout.com)', value: 'oxxo_link_text' },
-            { method: 'Boleto Bancário (via Checkout.com)', value: 'boleto_link_text' },
-        ];
+    const mandatePayment = order?.payments?.find((payment: GatewayOrderPayment | GiftCertificateOrderPayment)  => 'mandate' in  payment) as GatewayOrderPayment;
```
leeBigCommerce, line 24: Using `as` is an anti-pattern that we should try and avoid. Casting as `GatewayOrderPayment` is also not quite specific enough for our use cases, which is why we still have so many existence checks for certain attributes.

We can ensure our required entities more cleanly with a Type Guard, something like:

```tsx
type PaymentWithMandate = GatewayOrderPayment & Required<Pick<GatewayOrderPayment, 'mandate' | 'methodId'>>;

const isPaymentWithMandate = (payment: GatewayOrderPayment | GiftCertificateOrderPayment): payment is PaymentWithMandate =>
    !!payment.methodId && !!payment.mandate;
```

Where `PaymentWithMandate` is a sub-type of `Payment` which conforms to shape of data that our component requires. 

With that we can remove the `let`s (a bit of a code smell in functional component) and instead filter the payments upfront:

```tsx
const paymentsWithMandates = order?.payments?.filter(isPaymentWithMandate) || [];
```

Meaningly we can safely cycle over them in the JSX (with all values ensured, except `url`):

```tsx
{ paymentsWithMandates.map(payment => {
    if (payment.mandate.url) {
            // render it with a link... 
        );
    }

    // else render it as just text...
}) }
```

It might also be easier to read if the two different mandate JSX blocks are turned into their own Functional Components 🍹 

(Ideally we'd move all this data shaping out of this component to the source, but that'd be too much refactoring for this ticket)

leeBigCommerce:
Looking a lot better 👍 

Think we just need https://github.com/bigcommerce/checkout-sdk-js/pull/1220 finalised to get this build passing

src/app/order/OrderStatus.tsx
```diff
@@ -11,27 +11,18 @@ export interface OrderStatusProps {
     order: Order;
 }
 
+type PaymentWithMandate = GatewayOrderPayment & Required<Pick<GatewayOrderPayment, 'mandate' | 'methodId'>>;
+
+const isPaymentWithMandate = (payment: GatewayOrderPayment | GiftCertificateOrderPayment): payment is PaymentWithMandate =>
+    !!payment.methodId && ('mandate' in payment && !!payment.mandate);
```
leeBigCommerce, line 17: once you have the updated type from the sdk, this can just be:

`!!payment.methodId && !!payment.mandate`

FranciscoOnt: @leeBigCommerce I tired to make a build with the updated sdk and it's throwing the error
```
TS2339: Property ‘mandate’ does not exist on type ‘GiftCertificateOrderPayment | GatewayOrderPayment’.
      Property ‘mandate’ does not exist on type ‘GiftCertificateOrderPayment’. 
```

I will revert to the  `'mandate' in payment`.

JuanTorres077:
💚 Tested
QA Details: 
Working as intended, We raised a defect -5429, and there is the conclusion for the return button.

From checkout.com:
Please see this explanation in our docs. As long as the payment has not expired (unrelated to clicking return), the customer will be returned to your order confirmation (success) page and the payment will remain pending until the customer either pays at an ATM or the ticket expires.
In Sandbox, the three simulator buttons will do the following:
PAY - Captures the payment as if the customer has paid at an ATM (payment_captured)
EXPIRE - Expires the ticket, effectively canceling the payment (payment_expired)
RETURN - Returns to your site (payment_pending)
You can also test further scenarios via our API by including “simulate” in the source. description. See more info here: https://www.checkout.com/docs/payments/payment-methods/cash-and-atm-payment/boleto-bancario#Testing (edited)

Please take a look at the  below link:
[evidence| https://drive.google.com/drive/folders/1I4RfErHNQIGZ8_C41eCXYANPrh7wQUZH]

FranciscoOnt:
@leeBigCommerce  could you please re-review? ♻️

src/app/order/OrderStatus.tsx
```diff
@@ -11,29 +11,18 @@ export interface OrderStatusProps {
...
-            { method: 'OXXO (via Stripe)', value: 'oxxo_link_text'},
-            { method: 'Boleto Bancário (via Checkout.com)', value: 'boleto_link_text' },
-            { method: 'Boleto (via Stripe)', value: 'boleto_link_text'},
-        ];
-
-        const mandateText = Mandates.find(pair => pair.method === order?.payments?.[0].description);
-
-        return mandateText ? mandateText.value : 'mandate_link_text';
-    }, [order]);
+    const paymentsWithMandates = order?.payments?.filter(isPaymentWithMandate) || [];
```
leeBigCommerce, line 25: `order` seems to be mandatory, doesn't need the optional check

src/app/order/OrderStatus.tsx
leeBigCommerce: Makes sense, the current type definitions don't help us out in sorting `GiftCertificateOrderPayment`s from `GatewayOrderPayment`s

leeBigCommerce (approved):

src/app/order/OrderStatus.tsx
FranciscoOnt: done :+1:

FranciscoOnt:
Couldn't bump to latest checkout-sdk-js because it requires https://github.com/bigcommerce/checkout-js/pull/718 to be merged.

animesh1987 (approved):

PR #679, opened 2021-08-25
fix(checkout): CHECKOUT-5961 Screen reader does not announce error messages and order confirmation messages during checkout

bc-NikitaPuzanenko:
## What?
Adding an `aria-labelledby` attribute to error modal. Adding `id="errorMessage"` to error message <p> element so we can point to the correct element in `aria-labelledby`.
## Why?
Currently screen reader software does not read anything when an error message modal is rendered during checkout. This change will have screen readers read the existing error message that we already display in the error modal.
## Testing / Proof
Before:
Screen Shot 2021-08-25 at 3 56 19 PM

After:
Screen Shot 2021-08-25 at 3 54 54 PM

I am not sure that `errorMessage` the best id to use for the <p> element but, happy to change if anyone has a better name.
@bigcommerce/checkout


src/app/common/error/ErrorModal.tsx
```diff
@@ -63,7 +68,7 @@ export default class ErrorModal extends PureComponent<ErrorModalProps> {
 
         return (
             <Fragment>
-                { message && <p>{ message }</p> }
+                { message && <p id="errorMessage">{ message }</p> }
```
davidchin, line 71: I'd give it a more specific ID, e.g.: `errorModalMessage`. The reason is because ID has to be unique for the whole document. So `errorMessage` could potentially clash with other error messages in the document, e.g.: alert messages, error messages for a form etc...

davidchin (approved):
Just one more small suggestion, otherwise, LGTM, nice work! 🙂

src/app/common/error/ErrorModal.tsx
bc-NikitaPuzanenko: updated, thanks!

bc-peng (approved):
Thanks for making `checkout-js` more accessible.

RobinOng (approved):

briansilvey:
@bigcommerce/checkout  Hey all, we don't have the ability to merge this on the SSE side. Would someone from the team be able to merge this for us?

PR #682, opened 2021-08-30
fix(checkout): CHECKOUT-5980 invalid aria-labelledby values

bc-peng:
## What? [CHECKOUT-5980](https://jira.bigcommerce.com/browse/CHECKOUT-5980)
- Use `getLabelProps()` from [Downshift](https://www.downshift-js.com) to create a dynamic `id` attribute for an autocomplete input field's label, which aligns with the sample usage of [\<Downshift\> component](https://www.downshift-js.com/downshift).
- Update two snapshots to cover the code change.
## Why?
_1__-_Hi__We_re_ready_right_now_to_provide_you_with_a_free_LIVE_face_to_face_personalized_skin_consultation_-2

As stated in [CHECKOUT-5980](https://jira.bigcommerce.com/browse/CHECKOUT-5980), the screen reader fails now because of  a label's missing dynamic `id` attribute(e.g., `id="downshift-0-label"`). Sometimes, the `id` can be `downshift-2-label`. Therefore,  this fix will create label `id` dynamically with Downshift.

## Testing / Proof
Screen Shot 2021-08-30 at 10 48 23 am

Demo after code change.

https://user-images.githubusercontent.com/88361607/131297982-62e5efe3-3ca1-460c-a51d-99eeb893a2f2.mov

HTML of the address input field. Before:

```HTML
<div class="dynamic-form-field dynamic-form-field--addressLineAutocomplete">
    <div class="form-field">
        <label for="shippingAddress.address1" class="form-label optimizedCheckout-form-label">Address</label>
        <div role="combobox" aria-expanded="false" aria-haspopup="listbox" aria-labelledby="downshift-0-label">
            <input aria-autocomplete="list" aria-labelledby="downshift-0-label" autocomplete="off"
                   id="addressLine1Input" class="form-input optimizedCheckout-form-input" value="">
        </div>
    </div>
</div>
```

After:
```HTML
<div class="dynamic-form-field dynamic-form-field--addressLineAutocomplete">
    <div class="form-field">
        <div role="combobox" aria-expanded="false" aria-haspopup="listbox" aria-labelledby="downshift-0-label">
            <label
                    class="form-label optimizedCheckout-form-label" for="downshift-0-input"
                    id="downshift-0-label">Address</label>
            <input class="form-input optimizedCheckout-form-input"
                   id="downshift-0-input" autocomplete="off"
                   aria-autocomplete="list"
                   aria-labelledby="downshift-0-label" value="">
        </div>
    </div>
</div>
```

@bigcommerce/checkout

### DIFF
--- a/src/app/address/googleAutocomplete/GoogleAutocompleteFormField.tsx
+++ b/src/app/address/googleAutocomplete/GoogleAutocompleteFormField.tsx
@@ -56,6 +56,7 @@ const GoogleAutocompleteFormField: FunctionComponent<GoogleAutocompleteFormField
             isAutocompleteEnabled={ countryCode ?
                 supportedCountries.indexOf(countryCode) > -1 :
                 false }
+            labelContent={ labelContent }
             nextElement={ nextElement }
             onChange={ onChange }
             onSelect={ onSelect }
@@ -65,6 +66,7 @@ const GoogleAutocompleteFormField: FunctionComponent<GoogleAutocompleteFormField
         apiKey,
         countryCode,
         inputProps,
+        labelContent,
         nextElement,
         onChange,
         onSelect,
@@ -76,7 +78,7 @@ const GoogleAutocompleteFormField: FunctionComponent<GoogleAutocompleteFormField
         <div className={ `dynamic-form-field dynamic-form-field--addressLineAutocomplete` }>
             <FormField
                 input={ renderInput }
-                labelContent={ labelContent }
+                labelContent={ null }
                 name={ fieldName }
             />
         </div>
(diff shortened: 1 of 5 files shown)


src/app/address/googleAutocomplete/GoogleAutocompleteFormField.tsx
```diff
@@ -76,7 +78,7 @@ const GoogleAutocompleteFormField: FunctionComponent<GoogleAutocompleteFormField
         <div className={ `dynamic-form-field dynamic-form-field--addressLineAutocomplete` }>
             <FormField
                 input={ renderInput }
-                labelContent={ labelContent }
+                labelContent={ null }
```
animesh1987, line 81: I see you are lifting the label to the component where autocomplete is rendered? But why don't we pass the label inside here so `<FormField` is updated accordingly?

animesh1987: Think in that case you can get rid of the above line completely, it's an optional prop

bc-peng: I have removed this line.

animesh1987 (approved):

PR #696, opened 2021-09-10
fix(checkout): CHECKOUT-5980 invalid aria-labelledby values

bc-peng:
## What? 
- Calculate an `id` value for a `<label>`, and pass it to `<Downshift>` as the value of `aria-labelledby`, so that a screen reader can pick up an autocomplete input field's name.
-  No modification on tests.
## Why?
- As stated in [CHECKOUT-5980](https://jira.bigcommerce.com/browse/CHECKOUT-5980), a screen reader software fails now because the autocomplete address input does not have a matching `id` attribute.

## Testing / Proof
HTML code before:
```
<div class="dynamic-form-field dynamic-form-field--addressLineAutocomplete">
  <div class="form-field">
    <label
      for="shippingAddress.address1"
      class="form-label optimizedCheckout-form-label"
      >Address</label
    >
    <div
      role="combobox"
      aria-expanded="false"
      aria-haspopup="listbox"
      aria-labelledby="downshift-0-label"
    >
      <input
        aria-autocomplete="list"
        aria-labelledby="downshift-0-label"
        autocomplete="off"
        id="addressLine1Input"
        class="form-input optimizedCheckout-form-input"
        value=""
      />
    </div>
  </div>
</div>
```

After:
```
<div class="dynamic-form-field dynamic-form-field--addressLineAutocomplete">
  <div class="form-field">
    <label
      for="addressLine1Input"
      id="addressLine1Label"
      class="form-label optimizedCheckout-form-label"
      >Address</label
    >
    <div
      role="combobox"
      aria-expanded="false"
      aria-haspopup="listbox"
      aria-labelledby="addressLine1Label"
    >
      <input
        aria-autocomplete="list"
        aria-labelledby="addressLine1Label"
        autocomplete="off"
        id="addressLine1Input"
        class="form-input optimizedCheckout-form-input"
        value=""
      />
    </div>
  </div>
</div>

```

- An alpha release of `checkout-js` has been tested on Launchbay at:
https://app.circleci.com/pipelines/github/bigcommerce/bigcommerce?branch=pull%2F42735
All Google address autocomplete tests passed.

### DIFF
--- a/src/app/address/googleAutocomplete/GoogleAutocompleteFormField.tsx
+++ b/src/app/address/googleAutocomplete/GoogleAutocompleteFormField.tsx
@@ -5,7 +5,7 @@ import React, { memo, useCallback, useMemo, FunctionComponent } from 'react';
 import { TranslatedString } from '../../locale';
 import { AutocompleteItem } from '../../ui/autocomplete';
 import { FormField } from '../../ui/form';
-import { getAddressFormFieldInputId } from '../getAddressFormFieldInputId';
+import { getAddressFormFieldInputId, getAddressFormFieldLabelId } from '../getAddressFormFieldInputId';
 
 import GoogleAutocomplete from './GoogleAutocomplete';
 
@@ -40,10 +40,13 @@ const GoogleAutocompleteFormField: FunctionComponent<GoogleAutocompleteFormField
         <TranslatedString id="address.address_line_1_label" />
     ), []);
 
+    const labelID = getAddressFormFieldLabelId(name);
+
     const inputProps = useMemo(() => ({
         className: 'form-input optimizedCheckout-form-input',
         id: getAddressFormFieldInputId(name),
-    }), [name]);
+        'aria-labelledby': labelID,
+    }), [name, labelID]);
 
     const renderInput = useCallback(({ field }: FieldProps) => (
         <GoogleAutocomplete
@@ -76,7 +79,7 @@ const GoogleAutocompleteFormField: FunctionComponent<GoogleAutocompleteFormField
         <div className={ `dynamic-form-field dynamic-form-field--addressLineAutocomplete` }>
             <FormField
                 input={ renderInput }
-                labelContent={ labelContent }
+                label={ <label className="form-label optimizedCheckout-form-label" htmlFor={ inputProps.id } id={ labelID }>{ labelContent }</label> }
                 name={ fieldName }
             />
         </div>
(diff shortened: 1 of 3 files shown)


src/app/ui/autocomplete/Autocomplete.tsx
```diff
@@ -46,8 +48,12 @@ class Autocomplete extends PureComponent<AutocompleteProps> {
                     getMenuProps,
                     getItemProps,
                     highlightedIndex,
-                }) => (
+                }) => {
+                    const downshiftInputProps = getInputProps();
+
+                    return (
                     <div>
+                        <span id={ downshiftInputProps['aria-labelledby'] } style={ {width: 100, height: 100, position: 'absolute', top: ' -2000px ', left: 0} }>{ labelContent }</span>
```
davidchin, line 56: I think a better approach might be to pass `labelId` to `Downshift` component so that it doesn't have auto-generate a label ID that points to a non-existent element (https://github.com/downshift-js/downshift#labelid). And then you can point to the actual label that's visible to the users instead of having to create an invisible one?

bc-peng: Thanks for the feedback. I  have changed the code and tested it on Launchbay.  Also changed the `for` value of the `<label>` to the correct `id` of the `<input>`.

src/app/address/googleAutocomplete/GoogleAutocompleteFormField.tsx
```diff
@@ -76,7 +79,7 @@ const GoogleAutocompleteFormField: FunctionComponent<GoogleAutocompleteFormField
         <div className={ `dynamic-form-field dynamic-form-field--addressLineAutocomplete` }>
             <FormField
                 input={ renderInput }
-                labelContent={ labelContent }
+                label={ <label className="form-label optimizedCheckout-form-label" htmlFor={ inputProps.id } id={ labelID }>{ labelContent }</label> }
```
davidchin, line 82: Can we use the `Label` component instead?

bc-peng: Yes, I should use `Label` as much as possible. Will update you once I finish the test.

src/app/address/googleAutocomplete/GoogleAutocompleteFormField.tsx
```diff
@@ -40,10 +40,13 @@ const GoogleAutocompleteFormField: FunctionComponent<GoogleAutocompleteFormField
         <TranslatedString id="address.address_line_1_label" />
     ), []);
 
+    const labelID = getAddressFormFieldLabelId(name);
```
davidchin, line 43:
```suggestion
    const labelId = getAddressFormFieldLabelId(name);
```

davidchin (approved):
Nice one, LGTM 👍

src/app/address/googleAutocomplete/GoogleAutocompleteFormField.tsx
bc-peng: Thanks for the suggestion. I have replaced four occurrences of `labelID` with `labelId` in a new commit.

PR #700, opened 2021-09-20
fix(checkout): CHECKOUT-5928 Shopper cannot switch back to use single shipping address in checkout

bc-peng:
## What?
- When multiple consignments are detected in single shipping form, make an API call to reduce consignments number to one with the first shipping address.
- This check only happens at loading a single shipping form, so it won't introduce redundant API calls.

## Why? [Checkout-5928](https://jira.bigcommerce.com/browse/Checkout-5928)
When a shopper clicks `Ship to a single address` on the shipping form now, there is no API call to reduce consignments number to one. Therefore,  the checkout flow will continue with multiple shipping addresses.

## Testing / Proof
Link: [Functional tests on Launchbay](https://app.circleci.com/pipelines/github/bigcommerce/bigcommerce?branch=pull%2F42841).

Video demo:

https://user-images.githubusercontent.com/88361607/133955316-b76310fd-7fa8-4711-9f36-20b7cacbd5a2.mov

### DIFF
--- a/src/app/shipping/Shipping.tsx
+++ b/src/app/shipping/Shipping.tsx
@@ -120,7 +120,7 @@ class Shipping extends Component<ShippingProps & WithCheckoutShippingProps, Ship
                 <ShippingHeader
                     isGuest={ isGuest }
                     isMultiShippingMode={ isMultiShippingMode }
-                    onMultiShippingChange={ onToggleMultiShipping }
+                    onMultiShippingChange={ this.handleMultiShippingModeSwitch }
                     shouldShowMultiShipping={ shouldShowMultiShipping }
                 />
 
@@ -147,13 +147,36 @@ class Shipping extends Component<ShippingProps & WithCheckoutShippingProps, Ship
         );
     }
 
+    private handleMultiShippingModeSwitch: (isFromMultiToSingle: boolean) => void = async isFromMultiToSingle => {
+        const {
+            consignments,
+            onToggleMultiShipping = noop,
+            onUnhandledError = noop,
+            updateShippingAddress,
+        } = this.props;
+
+        if (isFromMultiToSingle && consignments.length > 1) {
+            this.setState({ isInitializing: true });
+
+            try {
+                // Collapse all consignments into one
+                await updateShippingAddress(consignments[0].shippingAddress);
+            } catch (error) {
+                onUnhandledError(error);
+            } finally {
+                this.setState({ isInitializing: false });
+            }
+        }
+
+        onToggleMultiShipping();
+    };
+
     private handleSingleShippingSubmit: (values: SingleShippingFormValues) => void = async ({
         billingSameAsShipping,
         shippingAddress: addressValues,
         orderComment,
     }) => {
         const {
-            consignments,
             customerMessage,
             updateCheckout,
             updateShippingAddress,
@@ -169,7 +192,7 @@ class Shipping extends Component<ShippingProps & WithCheckoutShippingProps, Ship
         const promises: Array<Promise<CheckoutSelectors>> = [];
         const hasRemoteBilling = this.hasRemoteBilling(methodId);
 
-        if (!isEqualAddress(updatedShippingAddress, shippingAddress) || consignments.length > 1 ) {
+        if (!isEqualAddress(updatedShippingAddress, shippingAddress)) {
             promises.push(updateShippingAddress(updatedShippingAddress || {}));
         }
 
--- a/src/app/shipping/Shipping.tsx
+++ b/src/app/shipping/Shipping.tsx
@@ -153,6 +153,7 @@ class Shipping extends Component<ShippingProps & WithCheckoutShippingProps, Ship
         orderComment,
     }) => {
         const {
+            consignments,
             customerMessage,
             updateCheckout,
             updateShippingAddress,
@@ -168,7 +169,7 @@ class Shipping extends Component<ShippingProps & WithCheckoutShippingProps, Ship
         const promises: Array<Promise<CheckoutSelectors>> = [];
         const hasRemoteBilling = this.hasRemoteBilling(methodId);
 
-        if (!isEqualAddress(updatedShippingAddress, shippingAddress)) {
+        if (!isEqualAddress(updatedShippingAddress, shippingAddress) || consignments.length > 1 ) {
             promises.push(updateShippingAddress(updatedShippingAddress || {}));
         }
 
--- a/src/app/shipping/Shipping.tsx
+++ b/src/app/shipping/Shipping.tsx
@@ -147,15 +147,16 @@ class Shipping extends Component<ShippingProps & WithCheckoutShippingProps, Ship
         );
     }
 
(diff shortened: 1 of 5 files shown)


src/app/shipping/SingleShippingForm.tsx
```diff
@@ -119,6 +119,10 @@ class SingleShippingForm extends PureComponent<SingleShippingFormProps & WithLan
         const PAYMENT_METHOD_VALID = ['amazon', 'amazonpay'];
         const shouldShowBillingSameAsShipping = !PAYMENT_METHOD_VALID.some(method => method === methodId);
 
+        if (consignments.length > 1) {
+            this.handleAddressSelect(consignments[0].shippingAddress);
```
davidchin, line 123: I don't think it's a good idea to perform any updates inside a `render` function. `render` should just focus on returning the React elements that need rendering whenever there's a change in state. It shouldn't make any API calls or call `setState`, which is what `handleAddressSelect` is doing.

A separate question is, how come updating the address of the first consignment can cause all the other consignments collapse into one 🤔

bc-peng: Thank you for the feedback. I have moved it to `componentDidMount()`.
All tests are passed now without modifications.

animesh1987: What about the second question DC mentioned above?

bc-peng: @animesh1987 The video is a recording of the current situation. The bug only appears when a user switches back to the single shipping-form and click `continue` without any interactions with address dropdown menu.

The bugfix acts the same as choosing the first option in the single shipping-form's dropdown menu. It does not change the current consignment update logic.

https://user-images.githubusercontent.com/88361607/133966139-06b5945e-3494-450f-9514-02dd75bb6508.mov

src/app/shipping/SingleShippingForm.tsx
```diff
@@ -89,6 +89,14 @@ class SingleShippingForm extends PureComponent<SingleShippingFormProps & WithLan
         }, SHIPPING_AUTOSAVE_DELAY);
     }
 
+    componentDidMount(): void {
+        const { consignments } = this.props;
+
+        if ( consignments.length > 1) {
+            this.handleAddressSelect(consignments[0].shippingAddress);
```
davidchin, line 96: Ok I did some digging and it seems SDK does some magic where calling `CheckoutService#updateAddress` would [re-assign](https://github.com/bigcommerce/checkout-sdk-js/blob/master/src/shipping/consignment-action-creator.ts#L156) all line items to the consignment of the address.

Could we add some unit tests for this? It is a fix so it means you're fixing a scenario not covered by our existing tests.

bc-peng: Sure, I'll add unit test(s) and run functional tests on Launchbay.

davidchin: Sorry @bc-peng, I had a look at this again and realised that perhaps a better solution is to make a call to combine the consignments when a user clicks on the "Ship to single address" button. The reason is because we're making a destructive call when this form component is mounted. I'm concerned that it's likely to cause problems in the future because engineers don't usually expect things to be deleted/updated just for showing something. But if the delete/update is explicitly tied to a user action, it's going to be easier to reason about. I know functionally speaking,  the two options have the same outcome because the form component is only mounted when a shopper clicks on the "Ship to single address" button. But latter option is more explicit. What do you think about this?

davidchin (commented):
Nice one, LGTM

src/app/shipping/SingleShippingForm.tsx
bc-peng: Hi @davidchin , the first thing I did in approaching this ticket was to locate [the function](https://github.com/bigcommerce/checkout-js/blob/84d758cbbd71be98a3c307bd0f34fc941248779a/src/app/checkout/Checkout.tsx#L465) triggered by the `Ship to single address` button. It was in the very high level `Checkout.tsx`.  [The API call](https://github.com/bigcommerce/checkout-js/blob/84d758cbbd71be98a3c307bd0f34fc941248779a/src/app/shipping/SingleShippingForm.tsx#L224) was in `singleShippingForm.tsx`, far away from `Checkout.tsx`.

I didn't figure out a straightforward way to trigger an API request from `Checkout.tsx`. Hence I had chosen to put everything inside `singleShippingForm.tsx`.

bc-peng: Update: It would make sense to add consignment number check to the existing single shipping handle function🤦🏻‍♂️,, `handleSingleShippingSubmit`, [Shipping.tsx#L171](https://github.com/bigcommerce/checkout-js/blob/84d758cbbd71be98a3c307bd0f34fc941248779a/src/app/shipping/Shipping.tsx#L171).

I have submitted the new code based on this idea.

src/app/shipping/ShippingHeader.tsx
```diff
@@ -32,7 +32,7 @@ const ShippingHeader: FunctionComponent<ShippingHeaderProps> = ({
             <a
                 data-test="shipping-mode-toggle"
                 href="#"
-                onClick={ preventDefault(onMultiShippingChange) }
+                onClick={ preventDefault(() => onMultiShippingChange(isMultiShippingMode)) }
```
davidchin, line 35: Do you need this change? AFAIK `isMultiShippingMode` is passed down from `Shipping`, so there's no need to pass it back up.

src/app/shipping/Shipping.tsx
```diff
@@ -147,6 +147,31 @@ class Shipping extends Component<ShippingProps & WithCheckoutShippingProps, Ship
...
+        const {
+            consignments,
+            isMultiShippingMode,
+            onToggleMultiShipping = noop,
+            onUnhandledError = noop,
+            updateShippingAddress,
+        } = this.props;
+
+        if (isMultiShippingMode && consignments.length > 1) {
+            this.setState({ isInitializing: true });
```
animesh1987, line 160: Do we need to set `isInitializing` here ? Since it's passed down here https://github.com/bigcommerce/checkout-js/pull/700/files#diff-c4c5cfa55abd9305c6b9130693f13101c7eeacfea5e8cf9f4901533dea05aa8eR370 which is a derived prop.

src/app/shipping/Shipping.tsx
```diff
@@ -147,6 +147,31 @@ class Shipping extends Component<ShippingProps & WithCheckoutShippingProps, Ship
         );
     }
 
+    private handleMultiShippingModeSwitch: () => void = async () => {
+        const {
+            consignments,
+            isMultiShippingMode,
+            onToggleMultiShipping = noop,
```
animesh1987, line 154: Do we need this `noop` props in this method?

src/app/shipping/Shipping.tsx
davidchin: We do. This is to activate the loading spinner - https://github.com/bigcommerce/checkout-js/blob/48fb255dfd398722fbc8aca0e100a6f5f9e31f56/src/app/shipping/Shipping.tsx#L128

davidchin (approved):
Nice work @bc-peng! Sorry for the back-and-forth 🙂

PR #724, opened 2021-10-12
fix(checkout): CHECKOUT-6030 remove child item from checkout address entry

bc-peng:
## What?
Screen Shot 2021-10-08 at 9 39 30 am
Remove child items from shipping address entry and shipping item statistics.

The difference between a parent item and a child item is `parentId` value. A parent item has a null or undefined  `parentId`. After limiting the line items to parent items only, the checkout interface will present them as desired.

Also, two unit tests are added to cover the parent and child item scenario.

## Why? 
As described in [Checkout-6030](https://jira.bigcommerce.com/browse/Checkout-6030), child items won't affect consignments. So removing them can avoid confusion for a shopper.

## Testing / Proof
The code has been tested with the `integration_full_UI_test` and `checkout` test group on Launchbay([Link](https://app.circleci.com/pipelines/github/bigcommerce/bigcommerce?branch=pull%2F43100)).

Demo:

https://user-images.githubusercontent.com/88361607/136871133-71a70ecb-0918-4b5b-b943-5dc64b80f714.mov

@bigcommerce/checkout


bc-peng:
> Nice work @bc-peng, just want to check this seems like a pretty fundamental change. Does it make sense to wrap this in an experiment?
> 
> Thoughts @davidchin ?

Hi @animesh1987 , I have checked the usages of `findLineItems.ts` and `getShippableLineItems.ts` are only limited to the shipping form(green area in What section). All UI tests are passed without modifications.

PR #745, opened 2021-11-05
feat(payment): BOLT-92 shows 'display name' value from method config

serhii-tkachenko:
## What?
When merchant changes display name in Control Panel, the display name doesn't show

## Why?
Because we need to add an ability to chose what payment title will be shown on Checkout page

## Testing / Proof
Manual tests
Unit tests

Tests:
Screenshot 2021-11-05 at 10 33 11

Screenshots of the changes:
Screenshot 2021-11-05 at 10 30 45

Screenshot 2021-11-05 at 10 03 26

Screenshot 2021-11-05 at 10 30 55

Screenshot 2021-11-05 at 10 07 33

Screenshot 2021-11-05 at 10 31 06

Screenshot 2021-11-05 at 10 04 08

@bigcommerce/checkout

### DIFF
--- a/src/app/payment/paymentMethod/PaymentMethodTitle.tsx
+++ b/src/app/payment/paymentMethod/PaymentMethodTitle.tsx
@@ -9,6 +9,7 @@ import { withLanguage, WithLanguageProps } from '../../locale';
 import { mapFromPaymentMethodCardType, CreditCardIconList } from '../creditCard';
 import { PaymentFormValues } from '../PaymentForm';
 
+import getPaymentMethodDisplayName from './getPaymentMethodDisplayName';
 import getPaymentMethodName from './getPaymentMethodName';
 import PaymentMethodId from './PaymentMethodId';
 import PaymentMethodType from './PaymentMethodType';
@@ -30,6 +31,7 @@ function getPaymentMethodTitle(
 
     return method => {
         const methodName = getPaymentMethodName(language)(method);
+        const methodDisplayName = getPaymentMethodDisplayName(language)(method);
         // TODO: API could provide the data below so UI can read simply read it.
         // However, I'm not sure how we deal with translation yet. TBC.
         const customTitles: { [key: string]: { logoUrl: string; titleText: string } } = {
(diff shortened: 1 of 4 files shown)


src/app/payment/paymentMethod/PaymentMethodTitle.tsx
```diff
@@ -79,7 +79,7 @@ function getPaymentMethodTitle(
             },
             [PaymentMethodId.Bolt]: {
                 logoUrl: '',
-                titleText: language.translate('payment.credit_card_text'),
+                titleText: (method.config.displayName?.toLowerCase() === 'credit card' ? language.translate('payment.credit_card_text') : method.config.displayName) || '',
```
animesh1987, line 82: For here why don't we write a method, something like `getBoltDisplayName(config) || ''`?

serhii-tkachenko: I thought to write a common method for this case, because we have almost the same for Klarna and some other payment providers.

serhii-tkachenko: I will create the method and let you know. Thanks! 🙏

src/app/payment/paymentMethod/PaymentMethodTitle.tsx
```diff
@@ -51,7 +53,7 @@ function getPaymentMethodTitle(
             },
             [PaymentMethodId.PaypalCommerceAlternativeMethod]: {
                 logoUrl: method.logoUrl || '',
-                titleText: (method.logoUrl ? '' : method.config.displayName) || '',
+                titleText: method.logoUrl ? '' : methodDisplayName,
```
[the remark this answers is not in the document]
serhii-tkachenko: We can take the logo url from method object and for this particular case we check if we have the logo we shows logo, otherwise we will show display name here. But it is a part of the old logic and I just changed method.config.displayName to methodDisplayName  what made the same under the hood

animesh1987 (approved):
Nice work 👍

PR #748, opened 2021-11-08
feat(payment): INT-5000 DigitalRiver: populate onUnhandledError correctly

clopezh:
## What?
Populate onUnhandledError correctly to show up the error message we sent from the strategy when the payment method isn't available.

## Why?
Digital River requested it because the shopper doesn't know what happened if the drop in element isn't rendered
image

## Testing / Proof
image
image

## Depends on
https://github.com/bigcommerce/checkout-sdk-js/pull/1286

### DIFF
--- a/src/app/locale/translations/en.json
+++ b/src/app/locale/translations/en.json
@@ -229,6 +229,7 @@
             "credit_card_number_required_error": "Credit Card Number is required",
             "credit_card_number_mismatch_error": "The card number entered does not match the card stored in your account",
             "digitalriver_dropin_error": "There was an error while processing your payment. Please try again or contact us.",
+            "digitalriver_checkout_error": "There was a problem with your checkout, please check your details and try again or contact customer service",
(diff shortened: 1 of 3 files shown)


AlanMunozN (approved):
lgtm

BC-ISilva:
💚 Tested!

QE details:
PR has been validated, issue is no longer available.

Following scenario has been validates:

- Use a Canadian shipping address, but introduce a US post code such as: 55343 **(positive and negative scenarios)**
- Use a US shipping address, but introduce a Canada post code such as: V6C 1B4 **(positive and negative scenarios)**

Please review attached document with validation details.
[INT-5000 - Post code validation does not occur causing checkouts call to error.docx](https://github.com/bigcommerce/checkout-js/files/7521712/INT-5000.-.Post.code.validation.does.not.occur.causing.checkouts.call.to.error.docx)

bc-peng (approved):
LGTM.👍🏻

PR #761, opened 2021-12-02
feat(checkout): CHECKOUT-6248 Display Error When Tax Service Unavailable

bc-peng:
## What? [CHECKOUT-6248](https://jira.bigcommerce.com/browse/CHECKOUT-6248)
* Parent ticket: [CHECKOUT-6174](https://jira.bigcommerce.com/browse/CHECKOUT-6174).
* When a tax service is not available, an error modal will appear as soon as a shopper clicks to pay.
* The error message is `Sales tax could not be calculated. Please try again. If unsuccessful, please reach out to our support team to complete checkout`.
* When a customer hits `Ok`, the page will reload.

Pending:
* Message translations.

## Depends on
https://github.com/bigcommerce/checkout-sdk-js/pull/1306

https://github.com/bigcommerce/bigcommerce/pull/43996
The back-end response is expected to be:
```
{
   "status":500,
   "code":10000,
   "title":"The tax provider is unavailable.",
   "type":"tax_provider_unavailable"
}
```

## Why?
When a merchant wants to be able to block transactions if the automated tax provider cannot be accessed, we want Checkout to stop customers from placing orders.

## Testing / Proof

https://user-images.githubusercontent.com/88361607/146317182-a09fd3f1-c92e-46ee-9abf-863efc528587.mov


src/app/checkout/Checkout.tsx
```diff
@@ -179,7 +180,16 @@ class Checkout extends Component<CheckoutProps & WithCheckoutProps & WithLanguag
                 this.handleReady();
             }
         } catch (error) {
-            this.handleUnhandledError(error);
+            if ( error.status === 500 && error.message === 'Tax Service Unavailable' ) {
```
bc-peng, line 183: Assuming we can identify `Tax Service Unavailable` with these.

theromulans: Is `502` a better fit here? Are there any other similar circumstances that we handle in the checkout context? If so, what HTTP error codes are expected there? e.g. payment provider is down.

theromulans:
Exciting to see the message in the video - thank you @bc-peng!

src/app/checkout/Checkout.tsx
bc-peng: We will work on the error code and message when the backend work begins. 👨🏻‍💻

src/app/checkout/Checkout.tsx
```diff
@@ -142,11 +143,11 @@ class Checkout extends Component<CheckoutProps & WithCheckoutProps & WithLanguag
             const errorFlashMessages = data.getFlashMessages('error') || [];
 
             if (errorFlashMessages.length) {
-                const { language } = this.props;
+                const {language: {translate}} = this.props;
```
animesh1987, line 146:
```suggestion
                const { language: { translate } } = this.props;
```

src/app/checkout/Checkout.tsx
```diff
@@ -179,7 +180,16 @@ class Checkout extends Component<CheckoutProps & WithCheckoutProps & WithLanguag
                 this.handleReady();
             }
         } catch (error) {
-            this.handleUnhandledError(error);
+            if ( error.status === 500 && error.message === 'Tax Service Unavailable' ) {
+                this.handleUnhandledError(new CustomError({
```
davidchin, line 184: So that tax error can pop up at any point, not just the initial load, e.g.: submit order, setting shipping address etc.... What I'd do is modify `handleUnhandledError` to do the error handling instead, because that method is triggered whenever there's an exception that's not caught at the layers below.

Also, I'd have the SDK throw a typed error, e.g.: `TaxServiceUnavailableError` and you can catch it directly rather than having to check the exact error message, which is comparatively more brittle.

animesh1987:
Also can you please rebase the branch, might be easier to review. Thanks

animesh1987 (approved):
Looks good to me. But will let @davidchin to give the final ✅

davidchin (approved):
Nice work @bc-peng!

PR #773, opened 2021-12-15
feat(payment): INT-4893 Add translations for errors from SDK - Humm

josesonora:
## What?  [INT-4893](https://jira.bigcommerce.com/browse/INT-4893)
Add support to translate, if necessary, error messages from the execute step in the payment strategy

## Why?
We need to translate the messages that are shown to customer to achieve a successful internationalization

## Testing / Proof
image

@bigcommerce/apex-integrations

### DIFF
--- a/src/app/payment/mapSubmitOrderErrorMessage.ts
+++ b/src/app/payment/mapSubmitOrderErrorMessage.ts
@@ -10,6 +10,9 @@ export default function mapSubmitOrderErrorMessage(
         case 'not_initialized':
             return translate('payment.payment_error');
 
+        case 'custom_provider_execute_error':
+            return translate(error.subtype);
+
         case 'payment_cancelled':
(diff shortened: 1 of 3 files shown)


src/app/payment/mapSubmitOrderErrorMessage.ts
```diff
@@ -10,6 +10,9 @@ export default function mapSubmitOrderErrorMessage(
         case 'not_initialized':
             return translate('payment.payment_error');
 
+        case 'payment_execute_error':
```
clopezh, line 13:
```suggestion
        case 'custom_provider_execute_error':
``` 
🍹

clopezh (approved):
I just recommend to use another word to specify that the error will be an specific per provider, otherwise 
lgtm

animesh1987 (approved):

PR #778, opened 2021-12-23
feat(payment): CHECKOUT-6070 Add applepay button component

animesh1987:
## What?
Add apple pay wallet method component

## Why?
In order to complete payment via apple pay as a wallet button, we need to create a component to render apple pay with it's extra config.

## Testing / Proof
- Screencast
- Test

https://user-images.githubusercontent.com/7134802/146874158-8d39a584-d572-4f9f-b4e5-77117ee67673.mov

@bigcommerce/checkout

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1196,9 +1196,9 @@
           }
         },
         "@babel/parser": {
-          "version": "7.16.10",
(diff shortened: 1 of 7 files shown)


davidchin (approved):
LGTM 👍

PR #802, opened 2022-01-31
feat(checkout): add zh and zh-TW locales

imgarylai:
## What?
Add `zh` and `zh-TW` locale support 

## Why?
Make big commerce support zh and zh-TW locale at checkout

## Testing / Proof
I had it tested on my store. (Let me know if you need store id) 

@bigcommerce/checkout


imgarylai:
This PR can't pass the commit validation because I don't have a ticket number.

snaderiBC:
Thanks @imgarylai for submitting this translation PR.
Unfortunately we cannot merge this PR because we only accept translation that is coming from our internal translation agency.
For that reason we've decided to close down this PR.

PR #808, opened 2022-02-10
feat(payment): INT-5384 Openpay: Add support to hosting their widget

mauricio-sg:
## What? [INT-5384](https://jira.bigcommerce.com/browse/INT-5384)
Created a new `OpyPaymentMethod` that wraps a `HostedWidgetPaymentMethod` to allow [the learn more button widget](https://widgets.openpay.com.au/widgets/learn-more-button) to be injected in the initialization step.

## Why?
Openpay wants to show this widget when the shopper selects Openpay as a payment option.

## Testing / Proof

**_Demo:_**

https://user-images.githubusercontent.com/4843328/153463964-97fb66af-4d49-4c93-b674-f01adae34025.mov

---

`OpyPaymentMethod`
OpyPaymentMethod
`HostedWidgetPaymentMethod`
HostedWidgetPaymentMethod

---

**_OnClick:_**

modal

---

**_AUSTRALIA_**

Australia

## Depends on:
👉 https://github.com/bigcommerce/checkout-sdk-js/pull/1344 to recognize `opy` as a known property of `PaymentInitializeOptions` and mount Opy's widget.


src/app/payment/paymentMethod/OpyPaymentMethod.tsx
```diff
@@ -0,0 +1,49 @@
...
+    const initializeOpyPayment = useCallback((options: PaymentInitializeOptions) => initializePayment({
+        ...options,
+        opy: {
+            containerId: 'learnMoreButton',
+        },
+    }), [initializePayment]);
+
+    return (
+      <LoadingOverlay hideContentWhenLoading isLoading={ isInitializing }>
+        <strong>Buy now. Pay smarter.</strong>
```
JoseAlvarezHdz, line 29: i think this should be added to the translated file

mauricio-sg: Thanks, I think it's some kind of slogan that I need to ask the Openpay team if it should be translated.

animesh1987: Agreed with @JoseAlvarezHdz, this should be in the translation file

src/app/payment/paymentMethod/OpyPaymentMethod.tsx
```diff
@@ -0,0 +1,49 @@
...
+    initializePayment,
+    method,
+    language,
+    isInitializing = false,
+    ...rest
+}) => {
+    const initializeOpyPayment = useCallback((options: PaymentInitializeOptions) => initializePayment({
+        ...options,
+        opy: {
+            containerId: 'learnMoreButton',
```
animesh1987, line 23: Can we move the name of this containerId, to a variable. Since it can break if we change it here but not on line 33

mauricio-sg: good idea!

src/app/payment/paymentMethod/OpyPaymentMethod.tsx
mauricio-sg: At the beginning of the project they told us that English is the only supported language, anyway, I asked them but they have not yet confirmed if their slogan needs to be translated.

ggarcia-bc: @animesh1987 there is some evidence they told us there is no need to translate, 
image
you can find it here:
https://intranet.bigcommerce.com/pages/viewpage.action?pageId=145859374

I wonder if there would be any internal constraint for us to add translation anyways

animesh1987: We have a centralised way of putting strings in the codebase, so you can just add it in en.json file if translation is not a concern. But we tend to avoid the above.

bc-peng: Agree with Animesh. I also did this before. We could put text into `en.json` even we don't need translations. 
That's the way we manage text content.

mauricio-sg: Ok, I didn't know it worked that way. Thanks, guys.

bc-peng (changes requested):
Just an FYI, CircleCI fails to build due to the recent code change.

mauricio-sg:
> Just an FYI, CircleCI fails to build due to the recent code change.

@bc-peng That's because of the dependency with https://github.com/bigcommerce/checkout-sdk-js/pull/1344

bc-peng (commented):
LGTM. Please run another CircleCI build once you have bumped `sdk`🙇🏻‍♂️

mauricio-sg:
> LGTM. Please run another CircleCI build once you have bumped `sdk`🙇🏻‍♂️

@bc-peng done ♻️

bc-peng (approved):

animesh1987 (approved):

PR #820, opened 2022-02-28
fix(checkout): CHECKOUT-6207 entered information lost after window resizing

bc-peng:
## What?
Always return `<CSStransition>` to avoid rerendering the payment step when the window is resizing.

## Why?
Previously, when the window width is smaller than `968px`, `checkout-js` renders `<div>`; when the window gets bigger, `checkout-js` renders  `<CSStransition>`. 

This results in the loss of information entered for the payment step. This bugfix will always return `<CSStransition>` while keeping the same logic that we only present CSS effects on larger screens.

Related doc: `enter` and `exit` control whether the `<CSStransition>` displays animations ([link](http://reactcommunity.org/react-transition-group/transition#Transition-prop-enter)).

## Testing / Proof

https://user-images.githubusercontent.com/88361607/155924316-40d4a19b-66bb-4c8d-8f16-ab9620937061.mov

### DIFF
--- a/src/app/checkout/CheckoutStep.tsx
+++ b/src/app/checkout/CheckoutStep.tsx
@@ -112,16 +112,12 @@ export default class CheckoutStep extends Component<CheckoutStepProps, CheckoutS
 
         return <>
             <MobileView>
-                { matched => {
-                    if (matched) {
-                        return !isActive ? null : <div className="checkout-view-content">
-                            { children }
-                        </div>;
-                    }
-
-                    return <CSSTransition
+                { matched =>
+                    <CSSTransition
                         addEndListener={ this.handleTransitionEnd }
                         classNames="checkout-view-content"
+                        enter={ !matched }
+                        exit={ !matched }
                         in={ isActive }
                         mountOnEnter
                         timeout={ {} }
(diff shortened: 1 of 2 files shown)


src/app/checkout/CheckoutStep.tsx
```diff
@@ -112,16 +112,12 @@ export default class CheckoutStep extends Component<CheckoutStepProps, CheckoutS
...
-                            { children }
-                        </div>;
-                    }
-
-                    return <CSSTransition
+                { matched =>
+                    <CSSTransition
                         addEndListener={ this.handleTransitionEnd }
                         classNames="checkout-view-content"
+                        enter = { !matched }
```
davidchin, line 119:
```suggestion
                        enter={ !matched }
```

src/app/checkout/CheckoutStep.tsx
```diff
@@ -112,16 +112,12 @@ export default class CheckoutStep extends Component<CheckoutStepProps, CheckoutS
...
+                        exit = { !matched }
```
davidchin, line 120:
```suggestion
                        exit={ !matched }
```

davidchin (approved):
Nice detective work @bc-peng

animesh1987 (approved):

PR #837, opened 2022-03-30
fix(payment): CHECKOUT-6549 Add check if window has apple pay supported

animesh1987:
## What?
- Add check before rendering component if apple pay is supported in the browser or not
- Bump checkout-sdk to get apple pay shipping fix

## Why?
If a browser doesn't support apple pay, at the moment there is an empty container rendered, therefore return nothing from the component if apple pay is not supported 

Also bump SDK https://github.com/bigcommerce/checkout-sdk-js/commit/e79f27b7727222b155677b84541926aff4859529

## Testing / Proof
- Tests

@bigcommerce/checkout

### DIFF
--- a/src/app/customer/CheckoutButtonList.tsx
+++ b/src/app/customer/CheckoutButtonList.tsx
@@ -1,16 +1,19 @@
 import { CustomerInitializeOptions, CustomerRequestOptions } from '@bigcommerce/checkout-sdk';
 import React, { memo, Fragment, FunctionComponent } from 'react';
 
+import { isApplePayWindow } from '../common/utility';
 import { TranslatedString } from '../locale';
 
 import { ApplePayButton } from './customWalletButton';
 import CheckoutButton from './CheckoutButton';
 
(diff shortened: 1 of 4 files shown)


bc-peng (approved):
`'ApplePaySession' in window` is good. I have tested it in Chrome and Safari.

src/app/customer/CheckoutButtonList.tsx
```diff
@@ -1,16 +1,19 @@
 import { CustomerInitializeOptions, CustomerRequestOptions } from '@bigcommerce/checkout-sdk';
 import React, { memo, Fragment, FunctionComponent } from 'react';
 
+import { isApplePayWindow } from '../common/utility';
 import { TranslatedString } from '../locale';
 
 import { ApplePayButton } from './customWalletButton';
 import CheckoutButton from './CheckoutButton';
 
+const APPLE_PAY = 'applepay';
```
bc-peng, line 10: Decent solution. 👏
You might also want to update the old code on line 79.

PR #844, opened 2022-04-06
PAYPAL-1329

bc-nick:
## What?
Change "Continue with PayPal" to "Place Order"

## Why?
According to https://jira.bigcommerce.com/browse/PAYPAL-1329

## Testing / Proof
Screenshot 2022-04-06 at 16 55 37

@bigcommerce/checkout


TomA-R (changes requested):
I don't think this is according to the task laid out in the ticket. It seems that this would replace the button text with "Place Order" in all cases, provided there's a single payment method available. I believe the ticket has a very particular scenario in mind, where if you return from PayPal express to the checkout, it becomes confusing to present "Continue with PayPal", in which case "Place Order" is more intuitive

BC-ACherednichenko (changes requested):
@TomA-R yeah, makes sense

BC-ACherednichenko (commented):
Per our discussion, 

PayPalCommerce - for the use case when a buyer goes directly to checkout uses an iframe button which is exposed via this interface 
https://github.com/bigcommerce/checkout-sdk-js/blob/master/src/payment/strategies/paypal-commerce/paypal-commerce-payment-initialize-options.ts#L70

consequently, a BC checkout button is not shown at all.

However, for the case when a buyer goes from the cart SPB a regular button is shown hence I believe we can update the label permanently without any condition

PR #862, opened 2022-05-04
feat(payment): INT-5904 Remove formatting for payment method when Mollie

clopezh:
## What?
Remove startCase method of Lodash to avoid the formatting of payment method names.

## Why?
According with the last call with Mollie's team, we shouldn't modify the payment method name in any way, we need to trust the mollie api source.

## Testing / Proof
image

### DIFF
--- a/src/app/payment/paymentMethod/PaymentMethodTitle.tsx
+++ b/src/app/payment/paymentMethod/PaymentMethodTitle.tsx
@@ -1,6 +1,6 @@
 import { LanguageService, PaymentMethod } from '@bigcommerce/checkout-sdk';
 import { number } from 'card-validator';
-import { compact, startCase } from 'lodash';
+import { compact } from 'lodash';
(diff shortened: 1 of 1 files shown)


vmparra (approved):
LGTM 👍🏻

BC-ISilva:
💚 Tested!!

**QE Details:**
Issue is not present anymore, fix is working as expected and now name for payment method is displaying correct whenever merchant has location in en_GB

- The display name for credit cards is "Credit / Debit Card"

Please review attached word file with validation details.
[  INT-5904 - [Mollie] Slash Is Removed From Credit Card Display Name When Locale Is en_GB  .docx](https://github.com/bigcommerce/checkout-js/files/8635391/INT-5904.-.Mollie.Slash.Is.Removed.From.Credit.Card.Display.Name.When.Locale.Is.en_GB.docx)

bc-peng (approved):

PR #864, opened 2022-05-09
CHECKOUT-6265: Add build script to generate export file that can automatically surface new payment integrations

davidchin:
## What?
* Added a build task that automatically generates an index file, re-exporting members matching a particular pattern from index files in another location.
* Added a function that can resolve to a React component by `id`, `gateway` and/or `type` property of a payment method object. Components that are resolvable must be exported with an additional property named `resolveIds`, indicating what properties to match against.
* Implemented an interface where dependencies can be injected to payment method components from their parent when they are loaded and rendered.

## Why?
* This will allow checkout team to auto-load new payment method components written by payment integration teams. This means in the future once the repo is restructured as a monorepo, there's no need for the payment integration teams to make a change to the `core` repo in order to integrate a new payment method into checkout, as long as it conforms to the predefined interface.

## Testing / Proof
Unit

@bigcommerce/checkout @bigcommerce/apex-integrations @bigcommerce/kyiv-payments-team

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ build
 coverage
 tests/_report/
 tests/_screenshots/
+src/app/generated/
+__temp__
--- a/auto-export.config.json
+++ b/auto-export.config.json
@@ -1,0 +1,9 @@
+{
+    "entries": [
+        {
+            "inputPath": "src/app/paymentIntegration/**/index.ts",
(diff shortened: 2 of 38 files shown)


animesh1987 (approved):
Really nice work @davidchin 👏

PR #871, opened 2022-05-16
feat(checkout): CHECKOUT-6643 add checkoutBillingSameAsShipping check for checkout

DeveloperAlan:
## What?
Allows store admin to enable/disable "Billing Address same as Shipping Address" check when a shopper is using checkout process. This can be checked within the admin panel of the store under the Checkout Settings.

## Why?
Store admin suggestion to BC as the customer would sometimes skip or ignore the check where it could be incorrect

## Testing / Proof

Screen Shot 2022-05-13 at 9 54 48 am
8-7316e97b9f82.png)

ping @bigcommerce/checkout

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,10 +22,9 @@
       }
     },
     "@babel/compat-data": {
-      "version": "7.17.7",
-      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.17.7.tgz",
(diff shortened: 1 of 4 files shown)


animesh1987 (approved):

bc-peng (approved):
Nice job! Welcome to the world of `checkout-js`😃.

rajguru827:
On page refresh, it's not preserving the checkbox state of Billing Address

animesh1987:
Hey @rajguru827 that is intended, we have moving to make that particular value merchant's config driven. So the API will populate that particular value going forward.

PR #879, opened 2022-05-25
fix(shipping): CHECKOUT-6643 Remove or check for billingSameAsShipping Flag

animesh1987:
## What?
Remove `||` check for billing same as shipping flag

## Why?
The `||` check disrespects the value of the store setting and always defaults to true.

## Testing / Proof
- Tests

@bigcommerce/checkout


TomA-R (approved):
Your build is failing though!

DeveloperAlan (approved):

bc-peng (approved):

PR #881, opened 2022-05-26
feat(checkout): CHECKOUT-6653 make cart drawer interactable via keyboard

huntario:
## What?
...
When screen width is narrowed, it moves to the cart summary to the bottom of the page. For accessibility purposes, this should be accessible with the 'tab' key, and the modal should display when clicking the "enter" key.  

## Why?
...
For better accessibility for screen reader users
https://bigcommercecloud.atlassian.net/browse/CHECKOUT-6653

## Testing / Proof

Before:
(cannot tab to modal trigger)

https://user-images.githubusercontent.com/5630999/171290143-94c10a7e-e581-4543-8c73-fb906fc89442.mp4

After: 
(tab to modal trigger, press enter to open modal)

https://user-images.githubusercontent.com/5630999/171289101-00569788-6324-41e0-8da4-cd44ad311d6e.mp4

...
POC for now (5/25) , will need to update this to only use the enter key to bring up the modal. With these immediate changes it becomes accessible via the tab key, but it should only react to the 'enter' key and right now it opens with any key. 

(5/27) Updated so that only the Enter key will open the modal when tabbed to. The focus is trapped once the users tabs into the modal, but it still need to update this so that the modal traps the focus without requiring hitting tab the first time.

(5/31) It is not clear if autofocus on the modal is desirable from an accessibility perspective. Upon hitting tab once the modal is opened, the user will be able to tab around the modal. Trapping the tab into the modal without the possibility of shift-tab should probably be handled in a separate PR.  For now this PR will make the modal accessible via tab and the enter button on the keyboard. 

@bigcommerce/checkout

### DIFF
--- a/src/app/order/OrderSummaryDrawer.spec.tsx
+++ b/src/app/order/OrderSummaryDrawer.spec.tsx
@@ -100,4 +100,28 @@ describe('OrderSummaryDrawer', () => {
                 });
         });
     });
+
+    describe('when active and enter is pressed', () => {
+        beforeEach(() => {
+            orderSummary.find('.cartDrawer').simulate('keypress', {key: 'Enter'});
+        });
+
+        it('renders order summary modal with the right props', () => {
+            expect(orderSummary.find(OrderSummaryModal).length)
+                .toEqual(1);
+
+            expect(orderSummary.find(OrderSummaryModal).find('#cart-print-link').length)
+                .toEqual(1);
+
+            expect(orderSummary.find(OrderSummaryModal).props())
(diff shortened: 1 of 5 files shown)


animesh1987 (commented):
Thanks @huntario looks good, can we add some tests to cover for the keyboard behaviour.

huntario:
Thanks for the feedback @animesh1987 . Please, let me know if there are any remaining concerns.

animesh1987 (approved):

PR #898, opened 2022-06-11
fix(payment): INT-6061 StripeUPE Vaulting selecting new card is failing

BC-PabloRivera:
## What? [INT-6061](https://bigcommercecloud.atlassian.net/browse/INT-6061)
Renders elements needed to get optimizedCheckout styles

## Why?
It's failing because renderContainer() function sets the display to none when certain conditions are met, and this elements need to always be rendered

## Testing / Proof
image

@bigcommerce/checkout @bigcommerce/payments @bigcommerce/kyiv-payments-team

### DIFF
--- a/src/app/payment/paymentMethod/StripeUPEPaymentMethod.tsx
+++ b/src/app/payment/paymentMethod/StripeUPEPaymentMethod.tsx
@@ -57,7 +57,7 @@ const StripeUPEPaymentMethod: FunctionComponent<StripePaymentMethodProps & WithI
         return getAppliedStyles(parentContainer, properties);
     };
 
-    const renderCustomPaymentForm = () => {
+    const renderCheckoutThemeStylesForStripeUPE = () => {
         return (
             <div
                 className={ 'optimizedCheckout-form-input' }
@@ -87,9 +87,8 @@ const StripeUPEPaymentMethod: FunctionComponent<StripePaymentMethodProps & WithI
             hideContentWhenSignedOut
             initializePayment={ initializeStripePayment }
             method={ method }
-            renderCustomPaymentForm={ renderCustomPaymentForm }
-            shouldRenderCustomInstrument={ true }
         />
+        { renderCheckoutThemeStylesForStripeUPE() }
     </>;
 };
 


vmparra (approved):
LGTM 👍🏻

FranciscoOnt (approved):
Adding some unit tests would be nice 🍹

src/app/payment/paymentMethod/StripeUPEPaymentMethod.tsx
```diff
@@ -87,9 +87,8 @@ const StripeUPEPaymentMethod: FunctionComponent<StripePaymentMethodProps & WithI
             hideContentWhenSignedOut
             initializePayment={ initializeStripePayment }
             method={ method }
-            renderCustomPaymentForm={ renderCustomPaymentForm }
-            shouldRenderCustomInstrument={ true }
         />
+        { renderOptimizedCheckoutStyles() }
```
bc-peng, line 91: Moving `renderOptimizedCheckoutStyles` out creates a new use case for `HostedWidgetPaymentMethod`. Can we avoid this by improving `HostedWidgetPaymentMethod` component?

BC-PabloRivera:
♻️ @bc-peng I did some changes, is this what you had in mind?

src/app/payment/paymentMethod/HostedWidgetPaymentMethod.tsx
```diff
@@ -226,6 +228,8 @@ class HostedWidgetPaymentMethod extends Component<
                         method={ method }
                         onSignOut={ this.handleSignOut }
                     /> }
+
+                    { renderAdditionalElements && renderAdditionalElements() }
```
bc-peng, line 232: This is still outside the designated container. Almost identical to the last solution. If the problem is that `renderContainer()` fails to render the element you want, we should aim to improve the way of invoking `renderContainer()`.

BC-PabloRivera: The div in renderContainer sets the display style to none when certain conditions are met, which deletes the custom payment form; this behavior is expected in payment methods that use custom payment forms.
Here I am trying to render some elements to load a stylesheet and get the styles from it, and I want it to always be rendered, so I shouldn't use the same renderCustomPaymentForm prop that is used for different purposes

BC-PabloRivera:
♻️ @bc-peng I left a comment on your review, can you check it?

BC-PabloRivera:
♻️ @bc-peng now rendering outside hostedWidget

src/app/payment/paymentMethod/StripeUPEPaymentMethod.tsx
```diff
@@ -57,7 +57,7 @@ const StripeUPEPaymentMethod: FunctionComponent<StripePaymentMethodProps & WithI
         return getAppliedStyles(parentContainer, properties);
     };
 
-    const renderCustomPaymentForm = () => {
+    const renderCheckoutThemeStyles = () => {
```
bc-peng, line 60:
```suggestion
    const renderCheckoutThemeStylesForStripeUPE = () => {
```

bc-peng (approved):

JelopezBC:
💚 Tested!

PR has been validated and is working as expected.

Please review the evidence bellow.

https://user-images.githubusercontent.com/84359490/173958373-1879c18a-70ab-411e-b1c6-cb932c5bc323.mov

PR #899, opened 2022-06-14
feat(checkout): CHECKOUT-6595 remove the edit cart button for buy now carts

DeveloperAlan:
## What?
Remove the edit cart button on the checkout page for buy now carts

## Why?
Shoppers shouldn't need to edit the cart if they use the buy now button for items

## Testing / Proof
Screen Shot 2022-06-14 at 3 55 08 pm

@bigcommerce/checkout

### DIFF
--- a/src/app/cart/CartSummary.tsx
+++ b/src/app/cart/CartSummary.tsx
@@ -20,14 +20,18 @@ export type WithCheckoutCartSummaryProps = {
 const CartSummary: FunctionComponent<WithCheckoutCartSummaryProps> = ({
     cartUrl,
     ...props
-}) => (
-    withRedeemable(OrderSummary)({
-        ...props,
-        cartUrl,
-        headerLink: (
-            <EditLink url={ cartUrl } />
-        ),
-    })
-);
+}) => {
+    const isBuyNowCart = window.location.pathname.replace('/checkout', '').replace('/embedded-checkout', '').length > 0;
+
+    const headerLink = isBuyNowCart ? null : <EditLink url={ cartUrl } />;
+
+    return (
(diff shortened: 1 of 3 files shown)


src/app/cart/CartSummary.tsx
```diff
@@ -20,14 +20,19 @@ export type WithCheckoutCartSummaryProps = {
...
-        ...props,
-        cartUrl,
-        headerLink: (
-            <EditLink url={ cartUrl } />
-        ),
-    })
-);
+}) => {
+    // if there's a sub path name in checkout, assume it's a buy now cart
+    const isBuyNowCart = window.location.pathname.replace('/checkout', '').length > 0;
```
bc-peng, line 25: In addition to standard checkout, we have [embedded checkout](https://www.mody.com.au/embedded-checkout) for Wordpress intergration. This line treats embedded checkout as if it were a buy now cart checkout. Think we also need to rule out the `/embedded-checkout` case.

src/app/cart/CartSummary.tsx
```diff
@@ -20,14 +20,19 @@ export type WithCheckoutCartSummaryProps = {
...
-    withRedeemable(OrderSummary)({
-        ...props,
-        cartUrl,
-        headerLink: (
-            <EditLink url={ cartUrl } />
-        ),
-    })
-);
+}) => {
+    // if there's a sub path name in checkout, assume it's a buy now cart
```
bc-peng, line 24: If the code isn't self-explanatory, we usually add a comment. Your solution is perfectly comprehensible. This line appears to be a little redundant.

bc-peng (approved):

PR #902, opened 2022-06-17
feat(checkout): CHECKOUT-6711 Update buy now cart order-confirmation URL

bc-peng:
## What?
When a checkout URL is a buy now cart URL,  `/checkout/:checkout-id`, its order confirmation page will put `orderId` in the URL, `/checkout/order-confirmation/:order-id`.

## Why?
We can load the right information for a buy now cart checkout.

## Testing / Proof
https://user-images.githubusercontent.com/88361607/174198576-cbca992b-c480-4c67-9c78-f0744e0fd622.mov

@bigcommerce/checkout


src/app/checkout/navigateToOrderConfirmation.tsx
```diff
@@ -1,7 +1,14 @@
 import { noop } from 'lodash';
 
-export default function navigateToOrderConfirmation(location = window.location): Promise<never> {
-    const url = `${location.pathname.replace(/\/$/, '')}/order-confirmation`;
+export default function navigateToOrderConfirmation(location = window.location, orderId?: number): Promise<never> {
```
bc-peng, line 3: By the time, `checkout-js` excutes ` navigateToOrderConfirmation()`, the order is finalised and the cart is gone. As a result, the `orderId` must be passed here.

bc-peng, line 3: By the time, `checkout-js` executes `navigateToOrderConfirmation()`, the order is finalised and the cart is gone. As a result, the `orderId` must be passed here.

src/app/checkout/navigateToOrderConfirmation.tsx
```diff
@@ -1,7 +1,14 @@
...
+    let url: string;
+    const checkoutId = window.location.pathname.replace('/checkout/', '').replace('embedded-checkout', '');
```
animesh1987, line 5: You can do this `window.location.pathname.split('/').pop()` and that way we don't need to rely on strings in the url
?

src/app/checkout/navigateToOrderConfirmation.tsx
```diff
@@ -1,7 +1,14 @@
...
+
+    if (orderId && checkoutId.length > 0) {
+        url = `/checkout/order-confirmation/${orderId.toString()}`;
```
animesh1987, line 8: You can use the same `${location.pathname.replace(/\/$/, '')}/order-confirmation/${orderId.toString()}` here as well?

bc-peng: I was told we don't support embedded checkout to use Buy Now Cart, hence I made the url in this way.

src/app/checkout/navigateToOrderConfirmation.tsx
bc-peng: Thanks for the tip. Thought it would be benificial to have a single method to tell whether it is a Buy Now Cart. So I also modified the code that displays the edit cart button.

src/app/common/utility/isBuyNowCart.ts
```diff
@@ -0,0 +1,3 @@
+export default function isBuyNowCart(): boolean {
+    return window.location.pathname.split('/').pop()?.length === 36;
```
animesh1987, line 2: Do we need to check for length 36? Can we check the inverse if the value is not `checkout` or `embedded-checkout` in that case

animesh1987 (approved):
Nice work @bc-peng LGTM

PR #904, opened 2022-06-20
fix(common): CHECKOUT-6674 Fix issue with alpha prerelease

animesh1987:
## What?
Fix issue with creating alpha prerelease

## Why?
As part of webpack upgrade, alpha prerelease is broken.  Which is an important part of dev cycle. Hence fix it so we can test changes in functional tests before releasing code.

https://webpack.js.org/guides/environment-variables/

## Testing / Proof
- Screenshot
Screen Shot 2022-06-20 at 11 05 53 pm

@bigcommerce/checkout

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "dev": "webpack --mode development --watch",
     "dev:server": "http-server build",
     "release": "echo 'Please do not release locally, use CircleCi'",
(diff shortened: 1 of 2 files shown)


bc-peng (approved):
Thanks.

PR #906, opened 2022-06-23
feat(checkout): CHECKOUT-6711 Add buy now cart experiment

bc-peng:
## What?
Add buy now cart experiment to wrap up the related changes.

## Why?
To control the risk of rolling out buy now cart.

## Testing / Proof
- Circle
- When `CHECKOUT-3190.enable_buy_now_cart` is on and a shopper is signed-in, the order confirmation page will be `/checkout/order-confirmation/:id/`.

https://user-images.githubusercontent.com/88361607/175231095-613295b7-8690-4fdd-b034-fd6400fad4df.mov

@bigcommerce/checkout

### DIFF
--- a/src/app/checkout/Checkout.tsx
+++ b/src/app/checkout/Checkout.tsx
@@ -75,6 +75,7 @@ export interface CheckoutState {
     isCartEmpty: boolean;
     isRedirecting: boolean;
     hasSelectedShippingOptions: boolean;
+    isBuyNowCartEnabled: boolean;
 }
 
 export interface WithCheckoutProps {
@@ -106,6 +107,7 @@ class Checkout extends Component<CheckoutProps & WithCheckoutProps & WithLanguag
         isRedirecting: false,
         isMultiShippingMode: false,
         hasSelectedShippingOptions: false,
+        isBuyNowCartEnabled: false,
     };
 
     private embeddedMessenger?: EmbeddedCheckoutMessenger;
@@ -170,12 +172,13 @@ class Checkout extends Component<CheckoutProps & WithCheckoutProps & WithLanguag
 
             const hasMultiShippingEnabled = data.getConfig()?.checkoutSettings?.hasMultiShippingEnabled;
             const checkoutBillingSameAsShippingEnabled = data.getConfig()?.checkoutSettings?.checkoutBillingSameAsShippingEnabled ?? true;
+            const buyNowCartFlag = data.getConfig()?.checkoutSettings?.features['CHECKOUT-3190.enable_buy_now_cart'] ?? false;
             const isMultiShippingMode = !!cart &&
                 !!consignments &&
                 hasMultiShippingEnabled &&
                 isUsingMultiShipping(consignments, cart.lineItems);
 
-            this.setState({ isBillingSameAsShipping: checkoutBillingSameAsShippingEnabled });
+            this.setState({ isBillingSameAsShipping: checkoutBillingSameAsShippingEnabled, isBuyNowCartEnabled: buyNowCartFlag });
 
             if (isMultiShippingMode) {
                 this.setState({ isMultiShippingMode }, this.handleReady);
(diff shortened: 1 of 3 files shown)


src/app/checkout/Checkout.tsx
```diff
@@ -502,7 +506,7 @@ class Checkout extends Component<CheckoutProps & WithCheckoutProps & WithLanguag
         }
 
         this.setState({ isRedirecting: true }, () => {
-            navigateToOrderConfirmation(window.location, orderId);
+            navigateToOrderConfirmation(window.location, orderId , isBuyNowCartEnabled);
```
animesh1987, line 509: I think we should reorder the params since `isBuyNowCartEnabled` will always have a value. Where as orderId is optional.

src/app/checkout/navigateToOrderConfirmation.tsx
```diff
@@ -2,10 +2,10 @@ import { noop } from 'lodash';
 
 import { isBuyNowCart } from '../common/utility';
 
-export default function navigateToOrderConfirmation(location = window.location, orderId?: number): Promise<never> {
+export default function navigateToOrderConfirmation(location = window.location, orderId?: number, isBuyNowCartEnabled?: boolean): Promise<never> {
```
animesh1987, line 5: Here we can reorder params and not have `isBuyNowCartEnabled` as undefined.

src/app/checkout/navigateToOrderConfirmation.tsx
```diff
@@ -2,10 +2,10 @@ import { noop } from 'lodash';
...
     let url: string;
 
-    if (orderId && isBuyNowCart()) {
+    if (orderId && isBuyNowCartEnabled && isBuyNowCart()) {
```
animesh1987, line 8: I am thinking does it make sense to split this method into two, one for if the buyNowcart experiment is enabled and another for the present workflow?

src/app/checkout/Checkout.tsx
bc-peng: I have changed the parameter as an object, think this is the most future proof solution. Then it turns out Apple Pay requires all `navigateToOrderConfirmation` params to be optional ([see here](https://github.com/bigcommerce/checkout-js/blob/a7fa190d45b55500eeacb7d7c837f5b771578d85/src/app/customer/customWalletButton/ApplePayButton.tsx#L21)).

src/app/checkout/navigateToOrderConfirmation.tsx
bc-peng: `location = window.location` seems for testing purpose only. We always use `window.location`. So I have removed `location` and kept the rest.

src/app/checkout/navigateToOrderConfirmation.tsx
bc-peng: I have wrapped up `isBuyNowCartEnabled` for simple removal as we'll remove it eventually.

src/app/checkout/navigateToOrderConfirmation.tsx
```diff
@@ -2,16 +2,22 @@ import { noop } from 'lodash';
 
 import { isBuyNowCart } from '../common/utility';
 
-export default function navigateToOrderConfirmation(location = window.location, orderId?: number): Promise<never> {
+export default function navigateToOrderConfirmation(isBuyNowCartEnabled?: boolean, orderId?: number): Promise<never> {
```
animesh1987, line 5: We removing the `location` param? 
Also `isBuyNowCartEnabled: boolean`

bc-peng: I see `location` is only used in testing, so leaving it there looks kind of redundant.
If `isBuyNowCartEnabled` is not optional, [this line](https://github.com/bigcommerce/checkout-js/blob/a7fa190d45b55500eeacb7d7c837f5b771578d85/src/app/customer/customWalletButton/ApplePayButton.tsx#L21) will complain.

animesh1987: In that case maybe need to open a PR for SDK to accept that param,https://github.com/bigcommerce/checkout-sdk-js/blob/master/packages/core/src/customer/strategies/apple-pay/apple-pay-customer-initialize-options.ts#L27

`    onPaymentAuthorize(isBuyNowCartEnabled: boolean): void;
`

animesh1987 (approved):

PR #910, opened 2022-06-29
fix(checkout): CHECKOUT-6756 fix up alignment for paymentProviders

DeveloperAlan:
## What?
Fix up alignment for payment logos

## Why?
The credit card icons in the payment step are misaligned when there are many icons. This is creating a poor shopper experience at the checkout. Merchants have reported that this UX is underwhelming.

## Testing / Proof

Screen Shot 2022-07-07 at 4 44 09 pm
Screen Shot 2022-07-07 at 12 12 15 pm

@bigcommerce/checkout

### DIFF
--- a/packages/core/src/app/payment/paymentMethod/PaymentMethodTitle.tsx
+++ b/packages/core/src/app/payment/paymentMethod/PaymentMethodTitle.tsx
@@ -1,7 +1,7 @@
 import { LanguageService, PaymentMethod } from '@bigcommerce/checkout-sdk';
 import { number } from 'card-validator';
 import { compact } from 'lodash';
-import React, { memo, Fragment, FunctionComponent } from 'react';
+import React, { memo, FunctionComponent } from 'react';
 
 import { withCheckout, CheckoutContextProps } from '../../checkout';
 import { connectFormik, ConnectFormikProps } from '../../common/form';
@@ -229,7 +229,7 @@ const PaymentMethodTitle: FunctionComponent<PaymentMethodTitleProps & WithLangua
     };
 
     return (
-        <Fragment>
+        <div className="paymentProviderHeader-container">
             { logoUrl && <img
                 alt={ methodName }
                 className="paymentProviderHeader-img"
@@ -237,12 +237,12 @@ const PaymentMethodTitle: FunctionComponent<PaymentMethodTitleProps & WithLangua
                 src={ logoUrl }
             /> }
 
-            { titleText && <span
+            { titleText && <div
                 className="paymentProviderHeader-name"
                 data-test="payment-method-name"
             >
                 { titleText }
-            </span> }
+            </div> }
 
             <div className="paymentProviderHeader-cc">
                 <CreditCardIconList
@@ -250,7 +250,7 @@ const PaymentMethodTitle: FunctionComponent<PaymentMethodTitleProps & WithLangua
                     selectedCardType={ getSelectedCardType() }
                 />
             </div>
-        </Fragment>
+        </div>
     );
 };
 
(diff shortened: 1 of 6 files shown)


bc-peng:
Hi there, it would be nice to have a before/after comparsion in the description. Also, this work seems to be a bugfix, hence the tag should be `fix(checkout)` rather than `chore(checkout)`.

bc-charlesho:
@DeveloperAlan the changes look good to me. 

@valterfatia can you review from a design perspective?

valterfatia:
Overall it looks good, but a couple of questions/feedback
- Why is the alignment different on the Google Pay role? Does one of the icons have a smaller size?
- What would happen if there are less than 5 marks on the provider? The original alignment should be right aligned and the way the last example shows leads me to think that will be different… aligning the marks to the right should resolve this and also look fine.

bc-peng:
Hi, Alan! On second thought, could you please test the changes on different screen sizes?

DeveloperAlan:
Here how it looks with less than 5 cards per row
Screen Shot 2022-07-07 at 12 10 23 pm
:

DeveloperAlan:
The Google Pay icon is a bit bigger by 39 x 25px and the credit cards are around 36 x 24px

DeveloperAlan:
Different screen sizes:
Screen Shot 2022-07-07 at 12 12 15 pm

valterfatia:
Overall it looks good to me, but can keep the radio/title aligned to the top like before?

DeveloperAlan:
Updated

Screen Shot 2022-07-07 at 4 44 09 pm

valterfatia:
Nice! Thanks @DeveloperAlan

valentindellangela:
Can you please update the image on the ticket proof of testing with your latest screenshot?

packages/core/src/scss/components/checkout/checklist/_checklist.scss
```diff
@@ -66,12 +66,11 @@
 
         background-color: $input-bg-color;
         border-radius: 100%;
-        bottom: 0;
         content: "";
         left: $checklist-pip-spacing--small;
         margin: auto;
         position: absolute;
-        top: 0;
+        top: 19px;
```
bc-peng, line 73: Could we use an existing function/variable to replace the hardcoded`19px`?

packages/core/src/scss/components/bigcommerce/credit-card-types/_credit-card-types.scss
```diff
@@ -13,9 +13,20 @@
...
+        flex-wrap: wrap;
+        justify-content: end;
+        margin-left: spacing("single");
+        max-width: 17rem;
+    }
+
     .creditCardTypes-list-item {
         opacity: 1;
         transition: all 0.6s ease-out;
+        width: 3rem;
```
bc-peng, line 29: Same here.

packages/core/src/app/payment/paymentMethod/PaymentMethodTitle.tsx
```diff
@@ -230,25 +230,27 @@ const PaymentMethodTitle: FunctionComponent<PaymentMethodTitleProps & WithLangua
 
     return (
         <Fragment>
```
davidchin, line 232: Do we still need a `Fragment` as we now have a container div?

packages/core/src/scss/settings/checkout/checklist/_settings.scss
```diff
@@ -15,3 +15,4 @@ $checklist-pip-backgroundImage:         url("data:image/svg+xml;utf8, %3Csvg%20w
 $checklist-pip-spacing:                 spacing("base") + spacing("third");
 $checklist-pip-spacing--small:          $checklist-pip-spacing / 1.5;
 $checklist-pip-spacingVertical:         remCalc(20px);
+$checklist-pip-spacing-top:             19px;
```
animesh1987, line 18:
```suggestion
$checklist-pip-spacing-top:             remCalc(19px);
```

animesh1987 (approved):

bc-peng (approved):

PR #912, opened 2022-07-01
feat(checkout): CHECKOUT-6269 Restructure checkout-js into monorepo

bc-peng:
## What?
Restructure `checkout-js` repository into a monorepo. 

Since the PR had more than 1000 files, we decided to only apply `nx` to `checkout-js` this week in order to make the PR smaller.

Three commits have been made:
1. moving files from `src` to `packages/core/src`;
2. applying `nx`;
3. fixing broken unit tests & upgrading `jest`.

## Next Steps
🔲 Configure linter in `nx`.
🔲 Create payment integration package.

## Why?
Better enforce code boundary and give greater autonomy to the payment integration team.

## Testing / Proof
CI.

@bigcommerce/checkout


animesh1987 (approved):
Nice work @bc-peng 👏

PR #914, opened 2022-07-06
fix(shipping): CHECKOUT-6773 Fix bug with address auto complete

animesh1987:
## What?
Fix issue with google address auto complete

## Why?
As of now if an address with a subpremise is returned we don't map it to the shipping address form. Hence we need to add subpremise to address mapper that can place the form value correctly.

## Testing / Proof
- Circle
- Screencast

https://user-images.githubusercontent.com/7134802/177555440-d3a97b46-7cf9-46c7-a5f5-98ebfdd9e8f7.mov

@bigcommerce/checkout

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "lint": "eslint 'packages/**/*.{ts,tsx}'",
     "pretest": "bin/auto-export.js",
     "test": "npx nx run-many --target=test --all",
-    "test:watch": "jest --watch"
+    "test:watch": "npx nx run-many --target=test --all --watch"
   },
   "repository": {
     "type": "git",
--- a/packages/core/src/app/address/googleAutocomplete/AddressSelector.ts
+++ b/packages/core/src/app/address/googleAutocomplete/AddressSelector.ts
@@ -22,7 +22,7 @@ export default class AddressSelector {
     }
 
     getStreet2(): string {
-        return '';
+        return this._get('subpremise', 'short_name');
     }
 
     getCity(): string {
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -14,6 +14,7 @@
     "build": "nx run core:build",
     "build:server": "http-server dist",
     "dev": "nx run core:dev",
+    "vm-dev": "WEBPACK_DONE='rsync -rvc --progress --delete --compress-level=9 build store.bcdev:/opt/bigcommerce_app/vagrant_code/vendor/bower_components/checkout/' npm run dev",
```
bc-peng, line 17:
```suggestion
    "dev:vm": "WEBPACK_DONE='rsync -rvc --progress --delete --compress-level=9 build store.bcdev:/opt/bigcommerce_app/vagrant_code/vendor/bower_components/checkout/' npm run dev",
```
Think `dev:vm` is better in line with other commands.

packages/core/src/app/address/googleAutocomplete/AddressSelector.ts
```diff
@@ -22,7 +22,7 @@ export default class AddressSelector {
     }
 
     getStreet2(): string {
-        return '';
+        return this._get('subpremise', 'short_name');
```
bc-peng, line 25: It turns out that Google Map API 's defintion of `subpremise` is not always constant. Sometimes, unit number is not a `subpremise` type. But this should be the best we could do.

bc-peng (approved):

packages/core/src/app/address/googleAutocomplete/AddressSelector.ts
animesh1987: yes we just have to address what's returned for the value.

PR #916, opened 2022-07-08
feat(checkout): CHECKOUT-6277 Apply eslint

bc-peng:
## What?
- Apply `@bigcommerce/eslint-config`.
- Remove `tslint`.

## Why?
Configure a comprehensive set of linter rules for `checkout-js` in order to programatically enforce a consistent coding style.

## Testing / Proof
CI.


packages/core/src/app/address/googleAutocomplete/AddressSelector.spec.ts
```diff
@@ -57,7 +57,7 @@ describe('AddressSelector', () => {
         });
 
         it('returns the locality as city if no postal town', () => {
-            // tslint:disable-next-line:no-non-null-assertion
+            // eslint-disable-next-line
```
davidchin, line 60:
```suggestion
            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```

davidchin (approved):
It'd be good to disable specific rule rather than disabling all rules like before. Otherwise, LGTM!

package.json
```diff
@@ -83,12 +83,12 @@
   "devDependencies": {
     "@babel/core": "^7.5.5",
     "@babel/preset-env": "^7.5.5",
-    "@bigcommerce/eslint-config": "^1.0.0",
-    "@bigcommerce/tslint-config": "^4.0.1",
+    "@bigcommerce/eslint-config": "^1.0.1",
     "@faker-js/faker": "^6.3.1",
     "@nrwl/cli": "^13.10.3",
-    "@nrwl/eslint-plugin-nx": "^14.3.6",
+    "@nrwl/eslint-plugin-nx": "^13.10.3",
```
DeveloperAlan, line 89: Why are we downgrading the plugin version?

package.json
```diff
@@ -83,12 +83,12 @@
   "devDependencies": {
     "@babel/core": "^7.5.5",
     "@babel/preset-env": "^7.5.5",
-    "@bigcommerce/eslint-config": "^1.0.0",
-    "@bigcommerce/tslint-config": "^4.0.1",
+    "@bigcommerce/eslint-config": "^1.0.1",
```
DeveloperAlan, line 86: could we upgrade this to v2.6.1?

package.json
bc-peng: Because `@nrwl/****` should match the version with `nx`. We are on `nx` 13.x, not the lastest 14. I learned this when I came across a strange error message.

package.json
davidchin: Nice catch @DeveloperAlan. @bc-peng I'm happy for you to upgrade to the latest version in a follow-up PR if the diff becomes too big.

bc-peng: Hi @DeveloperAlan , @DeveloperAlan . I think `v1.0.1` is the last `@bigcommerce/eslint-config` that can work with `typescript 3.9`. The `^2.0` requires `typescript @^4.0` as stated in its [package.json](https://github.com/bigcommerce/eslint-config/blob/a919ee7c6d4b04e4ad12604852423121dae68921/packages/eslint-config/package.json#L46). Also, below is a warning I'm receiving after running `npm install`.
> npm WARN @bigcommerce/eslint-plugin@1.1.1 requires a peer of typescript@^4.0.0 but none is installed. You must install peer dependencies yourself.

davidchin: FYI, we'll need to eventually upgrade the version of TS to at least 4.1 so it's compatible with the version of NX we've chosen. See https://github.com/bigcommerce/checkout-sdk-js/pull/1492

animesh1987: Yes but I think its good to proceed with this PR, and create a ticket and upgrade TS + Lint in a follow up/another PR.

animesh1987 (approved):

DeveloperAlan (approved):

packages/core/src/app/address/googleAutocomplete/AddressSelector.spec.ts
bc-peng: Every inline eslint-disable comment has named a specific rule now. Thanks.

PR #923, opened 2022-07-14
fix(order): INT-4776 Fix wrong mandate translation keys

FranciscoOnt:
## What?
change 'stripe_upe' key to 'stripeupe' and add prevent mandate to be rendered if mandateId is empty.

## Why?
Stripe UPE was showing the key instead of the correct string due to wrong key , and other providers that support mandate would render the mandate Id line even when method was not supported due to not being validated if a mandate ID was present.

## Testing / Proof
Pending


vmparra (approved):
LGTM 👍🏻

bc-peng (approved):

DeveloperAlan (approved):

PR #928, opened 2022-07-19
fix(checkout): CHECKOUT-6760 [Accessibility] - simplymedical.com - Ensure color is not the sole means of indicating error messages

bc-NikitaPuzanenko:
## What?
Reversing order for description and custom errors for password related error messages in the create customer form validation.

## Why?
Currently custom password related errors are not rendered on the create customer form during checkout. This is because the default description is evaluated and used first. This change will first check for a custom error message instead.

## Testing / Proof
Before:
Screen Shot 2022-07-19 at 3 24 08 PM
(note that the password error message is the default one)
After:
Screen Shot 2022-07-19 at 3 10 15 PM
(password error message is now read from the locale language file and updates based on what condition is failing)

@bigcommerce/checkout

### DIFF
--- a/packages/core/src/app/customer/CreateAccountForm.spec.tsx
+++ b/packages/core/src/app/customer/CreateAccountForm.spec.tsx
@@ -71,6 +71,33 @@ describe('CreateAccountForm Component', () => {
         });
     });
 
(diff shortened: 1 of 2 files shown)


DeveloperAlan (commented):
Does this PR need a unit test for the reverse order?

bc-NikitaPuzanenko:
@DeveloperAlan I added a unit test for each password condition. I am not sure if this is the correct way to test multiple conditions please let me know if there is a better way.

animesh1987 (approved):

PR #945, opened 2022-07-22
STRIPE-52 Auto-Select Address and mount shipping component for Stripe Link

AlanMunozN:
## What? [STRIPE-52](https://bigcommercecloud.atlassian.net/browse/STRIPE-52)
Adapt logic to support stripe link flow including:
- Stripe customer component,
- Stripe shipping component,
- Hide APMS when Stripe link is authenticated,
- Auto step for authenticated stripe customers
- Logic to handle shipping methods in different cases (least expensive, more expensive, no method selected, excluding pickup)
## Why?
As a Guest Shopper, I want to use the Stripe Link experience.

## Testing / Proof
**HOW TO SETUP**

## Dependency of
https://github.com/bigcommerce/checkout-sdk-js/pull/1518


packages/core/src/app/shipping/Shipping.tsx
```diff
@@ -345,6 +424,10 @@ export function mapToShippingProps({
     );
     const countriesWithAutocomplete = ['US', 'CA', 'AU', 'NZ'];
 
+    const card = getCart() || undefined;
```
FranciscoOnt, line 427: there is already a Cart in this scope, check line 393

packages/core/src/app/shipping/Shipping.tsx
```diff
@@ -345,6 +424,10 @@ export function mapToShippingProps({
...
+    const stripeUpe = getPaymentMethod('card', PaymentMethodId.StripeUPE);
+    const stripeUpeLinkEnable = stripeUpe && stripeUpe.initializationData.enableLink && card && card.currency.code === 'USD';
```
FranciscoOnt, line 429: you can break all those conditions in separate lines, if that works you may remove the complexity hack 
ie:
``` ts
    const linkEnabled = stripeUpe?.initializationData.enableLink || false;
    const stripeUpeSupportedCurrency = cart?.currency.code === 'USD' || false;
    const stripeUpeLinkEnable = linkEnabled && stripeUpeSupportedCurrency;
```

FranciscoOnt (changes requested):

AlanMunozN:
Thanks for the feedback @FranciscoOnt, I made changes, and It's ready for another review ♻️

FranciscoOnt (approved):
LGTM

BCHMercado:
Testing Done 💚 

packages/core/src/app/customer/Customer.tsx
```diff
@@ -148,10 +157,43 @@ class Customer extends Component<CustomerProps & WithCheckoutCustomerProps, Cust
...
+            updateStripeLinkAuthenticated,
+
         } = this.props;
 
         return (
+            isStripeLinkEnabled ?
+                <StripeGuestForm
+                    canSubscribe={ canSubscribe }
+                    checkoutButtons={
+                        <CheckoutButtonList
```
bc-peng, line 172: Can you explain why do we put `<CheckoutButtonList />` inside `<StripeGuestForm/>`?
Also, can we make stripe distinction in the `<GuestForm/>`?

packages/core/src/app/checkout/Checkout.tsx
```diff
@@ -76,6 +76,7 @@ export interface CheckoutState {
     isRedirecting: boolean;
     hasSelectedShippingOptions: boolean;
     isBuyNowCartEnabled: boolean;
+    isStripeLinkAuthenticated: boolean;
```
bc-peng, line 79: This looks like an ever growing list of states? Can we find another way to add StripeLink state?

BCHMercado:
💚 testing done 

packages/core/src/app/customer/Customer.tsx
AlanMunozN: Sure @bc-peng, I'm using CheckoutButtonList because `StripeLink` should have the same behavior that BC users are used to, which includes showing buttons for `Apple pay`, `Google pay`, `Paypal`, etc... and the difference between `guest form` and `stripe guest form` is:

Instead of loading a component for the email, we create a `div` in which the `stripe component for the customer` will be inserted.

We use extra information from the `stripe iframe`, it provides us with additional information when it's a stripe user or not, it makes a challenge for login and has its own logout, the logic of the button also has been changed for example:

If you are a detected `Stripe user`, you will automatically go to the next step without clicking (could be shipping)

the button is shown if it is a `BC guest`, if it is a stripe customer it is hidden, and if you are a `stripe customer` and decided to return to the `customer step` the button will be displayed with a specific text in case you do not want to make any change in the mail.

I hope the videos above can better explain the behavior, if not, please let me know and I will record a video specifically, explaining the customer's step

As a resume: We are using `CheckoutButtonList` inside `StripeGuestForm` because we want to show `wallets` such as `Apple pay`, `Google pay`, `Amazon pay`, etc...
And we made a copy of `GuestForm` because we have a very specific logic only for `StripeLink`, and we don't want to affect the `guest customer` just in case we have to make any other change

packages/core/src/app/checkout/Checkout.tsx
AlanMunozN: Hey @bc-peng, I was thinking to create an useContext or create a new interface with that field and extend by the class or inserting a hidden element with that info into the dom (but I don’t like it), please let me know if any of those options would be great or If you have any other idea would be great too :D

packages/core/src/app/customer/Customer.tsx
bc-peng: Hi Alan, in this case, it is logical to make`<GuestForm />` flexible enough to accept this kind of customization like `StripeLink`. Can you consider creating a `prop` or using `children`?

packages/core/src/app/checkout/Checkout.tsx
bc-peng: In `checkout-js`, we use `checkout-sdk-js` as the global state store. We can tell if checkout is in a loading state by using a sdk function. Is there a similar solution on `Stripe` end?

animesh1987:
Sorry for being late on this, know that we have already tested this and have done some review + feedback assessment.

Does it make sense to break this PR into 2:
- PR 1 dealing with customer step
- PR 2 dealing with shipping step

I ask so since this will reduce the number of lines and files that we need to look into. It eases the cognitive load on the reviewer and we can together work on a solution that scales, where as it's pretty easy to miss things in large PRs. 

Just a thought since it may speed up the whole process in this case.

AlanMunozN:
Hey @animesh1987, totally agree with you I'm working on the break into 2 PRs

AlanMunozN:
Hey @animesh1987  and @bc-peng, sorry for the late answer, I already created 2 PRs, please let me know any concerns https://github.com/bigcommerce/checkout-js/pull/1017 (customer and payment code), https://github.com/bigcommerce/checkout-js/pull/1018 (shipping code)

bc-peng:
@AlanMunozN Is it possible to close this PR as we have already got two other PRs?

PR #947, opened 2022-07-26
feat(checkout): STRF-9829 Redirect to cart on sign-out if prices are hidden from guests

jordanarldt:
### Depends on:
- https://github.com/bigcommerce/checkout-sdk-js/pull/1521

## What?
When a user clicks the "Sign Out" button during checkout while the "HidePriceFromGuests" configuration is turned on, they will be redirected to the cart page.

## Why?
**STRF-9829** will move the "Hide pricing for guests" functionality into store/channel configuration. While prices are restricted to login, we should redirect a user to the cart page if the log out during the checkout. This ensures stability and will prevent prices from being exposed if a customer signs out during the checkout.

Typically, a merchant would use the checkout setting to require an account. However, if prices should be hidden for guests, we should redirect back to the cart page on sign-out to ensure prices are not visible to the guest.

## Testing / Proof
- Added Unit Tests
- Video proof below

**BigCommerce Storefront:**

https://user-images.githubusercontent.com/19815878/181115867-441cfd19-2984-448d-bd74-ce77036896c8.mov

**Headless Embedded Checkout with HidePriceFromGuests enabled on the channel (redirects to channel cart route):**

https://user-images.githubusercontent.com/19815878/181116393-b7199a4f-cea4-4f65-998a-e11ad59ba6fa.mov

**Headless Redirected Checkout with HidePriceFromGuests enabled on the channel:**

https://user-images.githubusercontent.com/19815878/181116703-74c1df59-046e-4a6b-9445-39329922f024.mov

@bigcommerce/checkout

### DIFF
--- a/packages/core/src/app/checkout/mapToCheckoutProps.ts
+++ b/packages/core/src/app/checkout/mapToCheckoutProps.ts
@@ -21,7 +21,11 @@ export default function mapToCheckoutProps(
         links: {
             loginLink: loginUrl = '',
             createAccountLink: createAccountUrl = '',
+            cartLink: cartUrl = '',
         } = {},
+        displaySettings: {
+            hidePriceFromGuests: isPriceHiddenFromGuests = false,
+        } = {}
     } = data.getConfig() || {};
 
     const subscribeToConsignmentsSelector = createSelector(
@@ -40,8 +44,10 @@ export default function mapToCheckoutProps(
         isGuestEnabled,
         isLoadingCheckout: statuses.isLoadingCheckout(),
         isPending: statuses.isPending(),
+        isPriceHiddenFromGuests,
         loadCheckout: checkoutService.loadCheckout,
         loginUrl,
+        cartUrl,
         createAccountUrl,
         canCreateAccountInCheckout: features['CHECKOUT-4941.account_creation_in_checkout'],
         promotions,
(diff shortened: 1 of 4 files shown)


packages/core/src/app/checkout/mapToCheckoutProps.ts
```diff
@@ -40,8 +44,10 @@ export default function mapToCheckoutProps(
         isGuestEnabled,
         isLoadingCheckout: statuses.isLoadingCheckout(),
         isPending: statuses.isPending(),
+        isPriceHiddenFromGuests,
```
bc-peng, line 47: Can I know how did you set this prop? Think it should be something like `isPriceHiddenFromGuests: features['XXXX']`?

jordanarldt: Hey @bc-peng on line 26-28, I destructure the "hidePriceFromGuests" value returned from the `checkout-settings` endpoint and rename it to `isPriceHiddenFromGuests`. That prop is set directly from the checkout-settings API response.

I expose this store config in BCApp to the checkout-settings endpoint here: https://github.com/bigcommerce/bigcommerce/blob/0097e6a8ffd964c973224b6ec66810ecdcee7ec4/lib/Services/Checkout/Helpers/CheckoutDataHelper.php#L808

bc-peng: Thanks for the detailed explantation. All clear!

bc-peng (approved):
LGTM. Thanks.

PR #949, opened 2022-08-01
feat(payment): CHECKOUT-6460 Add Apple pay payment and customer integration test

animesh1987:
## What?
As above

## Why?
As part of TFA we want to protect our changes by adding integration tests for providers.

## Testing / Proof
- Circle

@bigcommerce/checkout


tests/apple-pay/ApplePaySessionMockObject.js
```diff
@@ -0,0 +1,64 @@
...
+
+        static supportsVersion(versionNumber) {
+            return true;
+        }
+
+        static canMakePayments() {
+            return true;
+        }
+
+       static canMakePaymentsWithActiveCard() {
```
davidchin, line 19: Tab alignment? 💅

tests/apple-pay/ApplePaySessionMockObject.js
```diff
@@ -0,0 +1,64 @@
+const addApplePaySessionToChrome = () => {
+    class ApplePaySession {
```
davidchin, line 2: Could we move the class definition out of the function?

tests/apple-pay/ApplePayTestMockResponse.ts
```diff
@@ -0,0 +1,342 @@
+export const payments = `[{
+    "id": "applepay",
+    "gateway": null,
+    "logoUrl": "https:\/\/shariceapplauded-cloud-dev-vm.store.bcdev\/rHEAD\/modules\/checkout\/applepay\/images\/applepay-header@2x.png",
```
davidchin, line 4: Could we substitute data that's specific to your dev env with dummy data (e.g.: logoUrl, merchantId, sentry etc...)?

tests/apple-pay/applepay.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+    test('ApplePay in payment step', async ({ assertions, checkout, page }) => {
+        // Testing environment setup
+        await page.addInitScript({ path: './tests/apple-pay/ApplePaySessionMockObject.js' });
+
+        const storeUrl = 'https://my-dev-store-117450812.store.bcdev/';
+        const responseProps = { status: 200, contentType: 'application/json' };
+        await checkout.use(new PaymentStepAsGuestPreset(storeUrl));
+        await checkout.create('ApplePay in Payment Step', storeUrl);
+        await checkout.route(/order-confirmation.*/, './tests/support/orderConfirmation.ejs', { orderId: '124' });
+        await page.route(/.*\/api\/storefront\/payments\?cartId=.*/, route => route.fulfill({ ...responseProps, body: payments }));
```
davidchin, line 14: Actually, I think all these endpoints should be automatically mocked via PollyJS. Is there a reason why it needs to mocked manually?

animesh1987: @bc-peng

bc-peng: `playwright` is unable to use `ApplePay` to place a legitimate order because `ApplePay` runs outside of the area where `playwright` can click. 

In order to fill in the responses that should be produced by interactions with `ApplePay`, for example, `/api/storefront/orders/124`, we mocked some API responses.

`/api/storefront/payments\payments` can be removed,  I mocked it to only return `ApplePay` in order to speed up the payment section loading.

davidchin: I think we can unless I'm mistaken. Although in our test we're using the fake `ApplePaySession` instead of the real `ApplePaySession`, we are still making HTTP calls to our APIs therefore they can be recorded by PollyJS and replayed in subsequent runs.

bc-peng: @davidchin I understand your concern that if the mocked `ApplePaySession` can actually make a valid API call, we shouldn't mock the call.

In this case, I think that 3 out of 7 mockups can be omitted:

- Loading payment methods.
- Loading `ApplePay`.
- Validating a merchant with `ApplePay` server.

Safari's bulit-in `ApplePay` JS generates a token during authorizing payment. We probably won't be able to reproduce it on Chrome. 

As a result, we have to mock everything from authorizing payment to the end.

https://developer.apple.com/documentation/passkit/apple_pay/payment_token_format_reference
Screen Shot 2022-08-05 at 3 52 22 pm

DeveloperAlan: I've removed the mock routes that Peng has suggested.

tests/apple-pay/applepay.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+        await page.addInitScript({ path: './tests/apple-pay/ApplePaySessionMockObject.js' });
+
+        const storeUrl = 'https://my-dev-store-759854325.store.bcdev';
+        const responseProps = { status: 200, contentType: 'application/json' };
+        await checkout.use(new PaymentStepAsGuestPreset(storeUrl));
+        await checkout.create('ApplePay in Payment Step', storeUrl);
+        await checkout.route(/order-confirmation.*/, './tests/support/orderConfirmation.ejs', { orderId: '124' });
+        await page.route('**/api/public/v1/payments/applepay/validate_merchant', route => route.fulfill({ ...responseProps, body: validateMerchantResponse }));
+        await page.route(/.*\/api\/storefront\/orders\/124.*/, route => route.fulfill({...responseProps, body: order }));
+        await page.route('**/api/public/v1/orders/payments', route => route.fulfill({ ...responseProps, body: orderPayment }));
```
animesh1987, line 16: Do we still need to mock these calls?

DeveloperAlan: My thought was we would still need the cart to be added to the checkout page?

playwright.config.ts
```diff
@@ -39,9 +39,9 @@ const config: PlaywrightTestConfig = {
   },
   projects: [
     {
-      name: 'chromium',
+      name: 'webkit',
       use: {
-        ...devices['Desktop Chrome'],
+        ...devices['Desktop Safari'],
```
bc-peng, line 44: Think we don't want to switch to `webkit`.
Also, this version of `Safari` does not have `ApplePay` inside.

DeveloperAlan: Thanks for the find, will update

bc-peng:
As we are going to convert an actual store URL to a generic store URL, an update to the HAR files seems inevitable.
Please wait for the test framework change.🙏🏻

tests/apple-pay/applepay.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+        await page.route('**/internalapi/v1/checkout/order', route => route.fulfill({
```
davidchin, line 17: I believe this endpoint `/internalapi/v1/checkout/order` can be mocked by PollyJS.

tests/fixture/pagePreset/ApiRequestsSender.ts
```diff
@@ -62,6 +62,7 @@ export class ApiRequestsSender {
 
         // Set shipping address
         const stateCode = faker.address.stateAbbr();
+        console.log('stateCode', stateCode)
```
davidchin, line 65: Can remove the `console.log`

tests/apple-pay/applepay.spec.ts
```diff
@@ -0,0 +1,58 @@
+import { test, CustomerStepPreset, PaymentStepAsGuestPreset } from '../';
+import { internalOrder, order, orderPayment, validateMerchantResponse, consignments } from './ApplePayTestMockResponse';
+
+test.describe('ApplePay', () => {
+    test.skip('ApplePay in payment step', async ({ assertions, checkout, page }) => {
```
davidchin, line 5: It looks like we're skipping this test?

davidchin (commented):
Are we actually running this test in CircleCI?

tests/apple-pay/applepay.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+        // Playwright actions
+        await checkout.goto();
+        await page.locator('label div').first().click();
+        await checkout.placeOrder();
+
+        // Assertions
+        await assertions.shouldSeeOrderConfirmation();        
+    })
+
+    test('ApplePay in customer step', async ({ assertions, checkout, page }) => {
```
davidchin, line 32: Forgot to mention, could we have a bit more descriptive test titles?

bc-peng:
> Are we actually running this test in CircleCI?

Not yet. Setting up CircleCI to run e2e tests is part of my current work.

tests/apple-pay/applepay.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+        const storeUrl = 'https://my-dev-store-759854325.store.bcdev';
+        const responseProps = { status: 200, contentType: 'application/json' };
+        await checkout.use(new CustomerStepPreset(storeUrl));
+        await checkout.create('ApplePay in Customer Step', storeUrl);        
+
+        await checkout.route(/order-confirmation.*/, './tests/support/orderConfirmation.ejs', { orderId: '124' });
+        await page.route('**/api/public/v1/payments/applepay/validate_merchant', route => route.fulfill({ ...responseProps, body: validateMerchantResponse }));        
+        await page.route(/.*\/api\/storefront\/orders\/124.*/, route => route.fulfill({...responseProps, body: order }));
+        await page.route('**/api/public/v1/orders/payments', route => route.fulfill({ ...responseProps, body: orderPayment }));
+        await page.route(/.*\/api\/storefront\/checkouts\/124.*\/consignments/, route => route.fulfill({...responseProps, body: consignments }));
```
bc-peng, line 45: It appears this test does not make calls to update shipping and contact. I see there is only a validating merchant event on line 82 of `ApplePaySessionMockObject.js`. Perhaps we can make another `ApplePaySessionMockObject` for this test.

tests/apple-pay/applepay.spec.ts
DeveloperAlan: I've updated the test to have more a descriptive title

tests/apple-pay/applepay.spec.ts
DeveloperAlan: Had that as a temporary measure, removed

tests/fixture/pagePreset/ApiRequestsSender.ts
DeveloperAlan: Removed

tests/apple-pay/applepay.spec.ts
DeveloperAlan: I've added the consignment and the billing address mock routes to match the shipping and the contact function calls. Is this adequate?

bc-peng: > 1. `this.onvalidatemerchant(event);` on [ApplePaySessionMockObject.js](https://github.com/bigcommerce/checkout-js/blob/e0f10d9a34ff3add5562be228f169424713f1c96/tests/apple-pay/ApplePaySessionMockObject.js#L63) will run the event defined in `checkout-sdk`.
> 2. API calls will be made by the actual code we want to test.
> 3. `playwright` will intercept the call and return mocked data.

It looks like you have finished step 3. Step 1 and 2 are still missing. Hence the test does not cover the code we want to test.

bc-peng:
There are many updates to [package-lock.json](https://github.com/bigcommerce/checkout-js/pull/949/files#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519), but not `package.json`.
Just want to confirm is this intentional?

.env-example
```diff
@@ -1,2 +1,3 @@
 PORT=8080
 MODE=replay
+STOREURL=https://my-dev-store-759854325.store.bcdev
```
animesh1987, line 3: end of line?

animesh1987:
Good catch @bc-peng, @DeveloperAlan can you please revert the package-lock.json file in the pr.

.env-example
davidchin, line 3: It should be a placeholder, not your own VM store URL.

DeveloperAlan: updated

tests/apple-pay/ApplePaySessionPaymentStepMockObject.js
```diff
@@ -0,0 +1,66 @@
+class ApplePaySession {
```
davidchin, line 1: I just noticed this, why is the mock file in JS not TS? 🤔

bc-peng: We learned (copied) this mock from [PineapplePay](https://chrome.google.com/webstore/detail/pineapple-pay/pkohlkkjckhbnapmcmodmnimndfcgpll?hl=en) Chrome extension at beginning. 🫣

davidchin: Ok let's convert this into TS 🙂

DeveloperAlan: @davidchin Playwright doesn't seem to allow TS files to be added as part of the `initScript` call. doc: https://playwright.dev/docs/api/class-browsercontext#browser-context-add-init-script-option-script

DeveloperAlan: The Playwright execution context does allow TS environment but doesn't allow base classes to extend from, nor the ability to get types from a node_module

packages/apple-pay-integration/e2e/ApplePaySessionPaymentStepMockObject.ts
```diff
@@ -0,0 +1,88 @@
...
+                this.onpaymentauthorized(event, this);
+            }, 1000)
+        }
+    }
+
+    // @ts-ignore
+    window.ApplePaySession = ApplePaySessionPaymentStep;
+}
+
+export default addApplePaySessionToChrome;
```
animesh1987, line 88: EOL?

packages/apple-pay-integration/e2e/ApplePaySessionCustomerStepMockObject.ts
```diff
@@ -0,0 +1,77 @@
+/* eslint-disable @typescript-eslint/ban-ts-comment */
+const addApplePaySessionToChrome = () => {
+    class ApplePaySessionCustomerStep {
+        [x: string]: any;
+        begin() {
+            setTimeout(() => {
+                console.log('testing')
```
animesh1987, line 7: console.log ?

.gitignore
```diff
@@ -7,4 +7,4 @@ coverage
 /dist-tools
 /packages/**/generated
 /packages/payment-integration-test-framework/report
-/packages/payment-integration-test-framework/screenshots
+/packages/payment-integration-test-framework/screenshots
```
animesh1987, line 10: EOL?

packages/apple-pay-integration/e2e/ApplePaySessionCustomerStepMockObject.ts
DeveloperAlan: fixed up

packages/apple-pay-integration/e2e/ApplePaySessionPaymentStepMockObject.ts
DeveloperAlan: fixed up

.gitignore
DeveloperAlan: fixed up

.env-example
```diff
@@ -1,2 +1,3 @@
 PORT=8080
 MODE=replay
+STOREURL=http://localhost:8081
```
davidchin, line 3:
```suggestion
STOREURL=
```

packages/apple-pay-integration/e2e/ApplePaySessionCustomerStepMockObject.ts
```diff
@@ -0,0 +1,111 @@
+/* eslint-disable @typescript-eslint/ban-ts-comment */
+/* eslint-disable @typescript-eslint/no-empty-function */
+const addApplePaySessionToChrome = () => {
+    class ApplePaySessionCustomerStep {
+        [x: string]: any;
```
davidchin, line 5: Is this necessary? Also, we should avoid using `any` if possible.

packages/apple-pay-integration/e2e/ApplePaySessionCustomerStepMockObject.ts
```diff
@@ -0,0 +1,111 @@
+/* eslint-disable @typescript-eslint/ban-ts-comment */
+/* eslint-disable @typescript-eslint/no-empty-function */
```
davidchin, line 2: Prefer `eslint-disable-next-line`, as we want to disable the linter on a case-by-case basis, not the entire file.

packages/apple-pay-integration/e2e/ApplePaySessionCustomerStepMockObject.ts
```diff
@@ -0,0 +1,111 @@
...
+                        payment: {
+                            shippingContact: {
+                                emailAddress: "mock@mock.com",
+                                familyName: "mock",
+                                givenName: "mock",
+                                phoneNumber: "00000000",
+                            }
+                        },
+                    }
+                } as unknown as ApplePayJS.ApplePayPaymentAuthorizedEvent;
```
davidchin, line 100: Is there a reason why you have to use `as unknown as`? Can you just provide dummy data that fully conforms to the `ApplePayPaymentAuthorizedEvent` interface?

packages/apple-pay-integration/e2e/applepay.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+        await page.route('**/api/public/v1/orders/payments', route => route.fulfill({ ...responseProps, body: orderPayment }));
+        await page.route('**/internalapi/v1/checkout/order', route => route.fulfill({
+            ...responseProps,
+            headers: { Token: 'White shirt now red, my bloody nose' },
+            body: internalOrder,
+        }));
+
+        // Playwright actions
+        await checkout.goto();
+        await page.locator('label div').first().click();
```
davidchin, line 27: Prefer to have a non-ambiguous selector, like `id` or `data-test`. `label div` is too ambiguous because it doesn't refer to a specific element and any changes to the HTML hierarchy could break the test.

packages/apple-pay-integration/e2e/ApplePaySessionCustomerStepMockObject.ts
DeveloperAlan: I've added this item to remove typescript errors on the Apple Pay methods that aren't declared on the file. eg.                 `this.onvalidatemerchant(validationEvent);`. I can just do a ts-ignore instead.

DeveloperAlan: I'll remove it as it's not part of the `ApplePaySession` definition

packages/apple-pay-integration/e2e/ApplePaySessionCustomerStepMockObject.ts
DeveloperAlan: We would need to add the Event typescript object that isn't related to the ApplePaySession itself, rather the Event focuses more on the browser event handling. Part of it also means we would need to mock EventTarget.

packages/apple-pay-integration/e2e/ApplePaySessionCustomerStepMockObject.ts
DeveloperAlan: Removed the next line and added it instead as next line

packages/apple-pay-integration/e2e/ApplePaySessionCustomerStepMockObject.ts
DeveloperAlan: Removed the key value type

packages/apple-pay-integration/e2e/ApplePaySessionCustomerStepMockObject.ts
DeveloperAlan: Updated the mock object to have the Event definition

packages/apple-pay-integration/e2e/applepay.spec.ts
DeveloperAlan: Currently I have used alt text to select Apple Pay, but after discussion https://bigcommerce.slack.com/archives/C02PJTFL5P0/p1661385791604789, we would also be adding `data-testid` instead to have an opinionated way of doing payment step tests

packages/core/src/app/ui/accordion/AccordionItem.tsx
```diff
@@ -46,6 +46,7 @@ const AccordionItem: FunctionComponent<AccordionItemProps> = ({
             className,
             { [classNameSelected]: isSelected }
         ) }
+        data-test={ `${itemId}_accordion-item` }
```
DeveloperAlan, line 49: Added a `data-test` in the accordion item for every accordion item however, what are people's thoughts

packages/apple-pay-integration/e2e/ApplePaySessionCustomerStepMockObject.ts
```diff
@@ -0,0 +1,180 @@
...
+                    KEYUP: 0,
+                    MOUSEDOWN: 0,
+                    MOUSEDRAG: 0,
+                    MOUSEMOVE: 0,
+                    MOUSEOUT: 0,
+                    MOUSEOVER: 0,
+                    MOUSEUP: 0,
+                    NONE: 0,
+                    SELECT: 0,
+                } as ApplePayJS.ApplePayPaymentAuthorizedEvent;
```
davidchin, line 166: Sorry I initially thought you can do something like:
```ts
class ApplePayPaymentAuthorizedEventMock extends Event implements ApplePayJS.ApplePayPaymentAuthorizedEvent {
    payment: {
        // ...
    }
}
```
But it turns out `ApplePayJS.ApplePayPaymentAuthorizedEvent` isn't typed properly - it has its own definition of `Event` rather than the one from `lib.dom.d.ts`. So in this case, I think it's justified to force cast. Could you please revert back to what you had?

packages/core/src/app/ui/accordion/AccordionItem.tsx
davidchin: Thanks for doing this. Could you flip it instead, e.g.: `accordion-item_${itemId}`

packages/apple-pay-integration/e2e/ApplePaySessionCustomerStepMockObject.ts
```diff
@@ -0,0 +1,180 @@
...
+                    shippingMethod: {
+                        label: 'xx',
+                        detail: 'xx',
+                        amount: "xx",
+                        identifier: 'xx'
+                    }
+                } as ApplePayJS.ApplePayShippingMethodSelectedEvent
+
+                // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+                // @ts-ignore
```
davidchin, line 67: What are you ignoring? Sorry, just to provide you with some context, we're against using `ts-ignore` unless there's a valid reason, for example, to work around an incorrect type definition provided by a third party library etc...

packages/apple-pay-integration/e2e/ApplePaySessionCustomerStepMockObject.ts
DeveloperAlan: reverted back to the old event object

packages/apple-pay-integration/e2e/ApplePaySessionCustomerStepMockObject.ts
DeveloperAlan:
```
Property 'onpaymentauthorized' does not exist on type 'ApplePaySessionPaymentStep'.
```

We don't want to mock the native Apple Pay methods so I kept these empty

packages/apple-pay-integration/e2e/ApplePaySessionCustomerStepMockObject.ts
DeveloperAlan: Updated

packages/apple-pay-integration/e2e/ApplePaySessionCustomerStepMockObject.ts
davidchin: Ok, the compiler is actually correct as the method is not defined. So what you need to do is to provide a no-op callback method. That'd be overridden by the implementation we're testing. e.g.:
```ts
onpaymentauthorized: (event: ApplePayJS.ApplePayPaymentAuthorizedEvent) => void = () => {
    console.log('onpaymentauthorized');
};
```

packages/apple-pay-integration/e2e/ApplePaySessionCustomerStepMockObject.ts
```diff
@@ -0,0 +1,130 @@
+const addApplePaySessionToChrome = () => {
+    class ApplePaySessionCustomerStep {
```
davidchin, line 2:
```suggestion
    class ApplePaySessionCustomerStep implements ApplePaySession {
```

We should probably do this to ensure the mock implementation conforms to the interface we are mocking.

packages/core/src/app/ui/accordion/AccordionItem.tsx
DeveloperAlan: updated

packages/apple-pay-integration/e2e/ApplePaySessionCustomerStepMockObject.ts
DeveloperAlan: Implemented the ApplePaySession

packages/apple-pay-integration/e2e/ApplePaySessionCustomerStepMockObject.ts
DeveloperAlan: Cool, I've implemented the no-op methods and override the implementations with the methods from the `ApplePaySession`

packages/apple-pay-integration/project.json
```diff
@@ -20,5 +20,6 @@
       }
     }
   },
-  "tags": ["scope:integration"]
+  "tags": ["scope:integration"],
+  "types": "index.d.ts"
```
bc-peng, line 24: I cannot find the doc about `types` here. Could you please explain what this is for?

packages/apple-pay-integration/e2e/applepay.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+        await page.route('**/api/public/v1/orders/payments', route => route.fulfill({ ...responseProps, body: orderPayment }));
+        await page.route('**/internalapi/v1/checkout/order', route => route.fulfill({
+            ...responseProps,
+            headers: { Token: 'White shirt now red, my bloody nose' },
+            body: internalOrder,
+        }));
+
+        // Playwright actions
+        await checkout.goto();
+        await page.locator('[data-test=accordion-item_applepay]').click();
```
bc-peng, line 27: Thanks for adding `data-test`. Will wrap it as a helper function once this PR is merged.

packages/apple-pay-integration/e2e/ApplePayTestMockResponse.ts
```diff
@@ -0,0 +1,573 @@
...
+            "merchantId":"00000000-0000-0000-0000-000000000000",
+            "paymentsUrl":"https://bigpay.service.bcdev",
+            "sentry":"https://example@sentry.io/example",
+            "confirmationLink":"/checkout/order-confirmation"
+        },
+        "clientToken":null,
+        "returnUrl":null
+    }`
+
+export const payments = `[{
```
bc-peng, line 49: We may remove this unused mockup.

packages/apple-pay-integration/e2e/ApplePaySessionPaymentStepMockObject.ts
```diff
@@ -0,0 +1,137 @@
...
+            return true;
+        }
+
+        static canMakePayments() {
+            return true;
+        }
+
+        static canMakePaymentsWithActiveCard() {
+            return Promise.resolve(this.canMakePayments());
+        }
```
bc-peng, line 53: Think this section can be moved to the top for better orgnization of the class.

packages/apple-pay-integration/e2e/ApplePaySessionCustomerStepMockObject.ts
```diff
@@ -0,0 +1,162 @@
...
+
+        onvalidatemerchant: (event: ApplePayJS.ApplePayValidateMerchantEvent) => void  = () => { 
+            console.log('onvalidatemerchant');
+        }
+
+        abort(): void {
+            console.log('abort')
+        }
+
+        STATUS_SUCCESS = 1;
```
bc-peng, line 49: Same here, plus:
```suggestion
        static STATUS_SUCCESS = 1;
```

packages/apple-pay-integration/e2e/ApplePaySessionCustomerStepMockObject.ts
```diff
@@ -0,0 +1,162 @@
...
+        addEventListener(_type: string, _callback: EventListenerOrEventListenerObject | null, _options?: boolean | AddEventListenerOptions | undefined): void {
+            console.log('addEventListener', _type, _callback, _options);
+        }
+        dispatchEvent(_event: Event): boolean {
+            console.log('dispatchEvent', _event);
+            return true
+        }
+        removeEventListener(_type: string, _callback: EventListenerOrEventListenerObject | null, _options?: boolean | EventListenerOptions | undefined): void {
+            console.log('removeEventListener', _type, _callback, _options);
+        }
```
bc-peng, line 19:
```suggestion
        constructor(version, paymentRequest) {
            this.version = version;
            this.paymentRequest = paymentRequest;
        }
                
        addEventListener(_type: string, _callback: EventListenerOrEventListenerObject | null, _options?: boolean | AddEventListenerOptions | undefined): void {
            console.log('addEventListener', _type, _callback, _options);
        }
        
        dispatchEvent(_event: Event): boolean {
            console.log('dispatchEvent', _event);
            return true
        }
        
        removeEventListener(_type: string, _callback: EventListenerOrEventListenerObject | null, _options?: boolean | EventListenerOptions | undefined): void {
            console.log('removeEventListener', _type, _callback, _options);
        }
```

packages/apple-pay-integration/tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "extends": "../../tsconfig.base.json",
   "include": [
     "src/**/*.ts",
-    "src/**/*.tsx"
-  ]
-}
+    "src/**/*.tsx",
+    "types/*.d.ts"
+  ],
+}
```
animesh1987, line 8: EOL?

packages/apple-pay-integration/types/ApplePaySession.d.ts
```diff
@@ -0,0 +1,7 @@
+export {};
+
+declare global {
+  interface Window {
+    ApplePaySession: ApplePaySession
+  }
+}
```
animesh1987, line 7: EOL?

packages/apple-pay-integration/project.json
DeveloperAlan: Was a leftover, will remove

packages/apple-pay-integration/e2e/ApplePayTestMockResponse.ts
DeveloperAlan: Removed the extra mockup

packages/apple-pay-integration/e2e/ApplePaySessionPaymentStepMockObject.ts
DeveloperAlan: moved to the top

packages/apple-pay-integration/e2e/ApplePaySessionCustomerStepMockObject.ts
DeveloperAlan: move to the top as well as added as a static variable

packages/apple-pay-integration/e2e/ApplePaySessionCustomerStepMockObject.ts
DeveloperAlan: committed the spacing

packages/apple-pay-integration/tsconfig.json
DeveloperAlan: added

packages/apple-pay-integration/types/ApplePaySession.d.ts
DeveloperAlan: added

packages/apple-pay-integration/e2e/ApplePaySessionCustomerStepMockObject.ts
```diff
@@ -0,0 +1,165 @@
...
+            console.log('supportsVersion', _versionNumber)
+            return true;
+        }
+
+        constructor(version, paymentRequest) {
+            this.version = version;
+            this.paymentRequest = paymentRequest;
+        }
+
+        addEventListener(_type: string, _callback: EventListenerOrEventListenerObject | null, _options?: boolean | AddEventListenerOptions | undefined): void {
```
davidchin, line 19: Don't prefix the parameter with an underscore if you're actually using it.

packages/apple-pay-integration/types/ApplePaySession.d.ts
```diff
@@ -0,0 +1,7 @@
+export {};
+
+declare global {
+  interface Window {
+    ApplePaySession: ApplePaySession
```
davidchin, line 5: I believe ApplePay isn't defined for every browser so it is more accurate to put a question mark in front.
```suggestion
    ApplePaySession?: ApplePaySession
```

packages/apple-pay-integration/e2e/ApplePaySessionPaymentStepMockObject.ts
```diff
@@ -0,0 +1,137 @@
...
+                    },
+                } as unknown as ApplePayJS.ApplePayPaymentAuthorizedEvent;
+
+                this.onpaymentauthorized(event);
+            }, 1000)
+        }
+    }
+
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
```
davidchin, line 133: Do you still need `ts-ignore`?

packages/apple-pay-integration/e2e/ApplePaySessionPaymentStepMockObject.ts
```diff
@@ -0,0 +1,137 @@
+const addApplePaySessionToChrome = () => {
+    class ApplePaySessionPaymentStep implements ApplePaySession {
+        version: number;
+        paymentRequest: ApplePayJS.ApplePayPaymentRequest;
+
+        static STATUS_SUCCESS = 1;
+        static STATUS_FAILURE = 2;
+
+        static supportsVersion(_versionNumber) {
```
davidchin, line 9: Need type definition for all parameters in this file.

packages/apple-pay-integration/types/ApplePaySession.d.ts
```diff
@@ -0,0 +1,7 @@
+export {};
```
davidchin, line 1: Is there a reason for this? Could you do `export declare global` instead if you want to export something?

DeveloperAlan: Good point, I can update it

packages/apple-pay-integration/e2e/ApplePaySessionPaymentStepMockObject.ts
DeveloperAlan: Sure thing I can update that

packages/apple-pay-integration/e2e/ApplePaySessionPaymentStepMockObject.ts
DeveloperAlan: Not needed anymore, will remove for this file

packages/apple-pay-integration/types/ApplePaySession.d.ts
DeveloperAlan: good point, will update

packages/apple-pay-integration/e2e/ApplePaySessionCustomerStepMockObject.ts
DeveloperAlan: Removed the prefix

davidchin (approved):
Thanks @DeveloperAlan for your patience in working through the changes. Much appreciated!

PR #950, opened 2022-08-01
feat(checkout): CHECKOUT-6426 Refactor ApplePay as Monorepo Package

bc-peng:
## What?
Refactor the existing `ApplePay` component to use the new `checkout-js` payment integration JS API.

## Why?
Lay the foundation work for restructuring `checkout-js` into a monorepo.

### Unit testing:
Add two new unit tests.

### Manual testing:
I. Mock `/api/storefront/payments?cartId=*` to return `ApplePay` payment method only.
Screen Shot 2022-08-01 at 4 54 27 pm

II. Use `<PaymentMethodV2/>` in `PaymentMethodList.tsx`.

III. With `Pineapple Pay` plug-in, Chrome executes `ApplePay` initialization.
Screen Shot 2022-07-29 at 11 14 54 pm


auto-export.config.json
```diff
@@ -1,9 +1,9 @@
 {
     "entries": [
         {
-            "inputPath": "packages/core/src/app/paymentIntegration/**/index.ts",
-            "outputPath": "packages/core/src/app/generated/paymentIntegrations/index.ts",
-            "memberPattern": ".+PaymentMethod$"
+            "inputPath"     : "packages/**/src/index.ts",
```
animesh1987, line 4: is there a reason for the spaces added?

bc-peng: After addition, it would be eaiser to read and compare those paths.

packages/apple-pay-integration/src/ApplePayPaymentMethod.tsx
```diff
@@ -0,0 +1,43 @@
+import { toResolvableComponent, PaymentMethodProps } from '@bigcommerce/checkout-js/payment-integration';
+import React, { FunctionComponent, useEffect } from 'react';
+
+const ApplePaymentMethod: FunctionComponent<PaymentMethodProps> = ({ method, checkoutService, language, onUnhandledError }) => {
+    useEffect(() => {
+        const componentDidMount = async ():Promise<void> => {
```
animesh1987, line 6: You should be able to rename the function to something more related like `initializePayment`?

bc-peng: Sure, I was going to ask about this. This naming is for nostalgia. 😂 So you can see two old friends.
Will rename them.

packages/apple-pay-integration/src/ApplePayPaymentMethod.tsx
```diff
@@ -0,0 +1,43 @@
...
+                    applepay: {
+                        shippingLabel: language.translate('cart.shipping_text'),
+                        subtotalLabel: language.translate('cart.subtotal_text'),
+                    },
+                });
+            } catch (error) {
+                onUnhandledError(error);
+            }
+        }
+        componentDidMount().then();
```
animesh1987, line 20: Do we need `.then` here, should be ok to do `componentDidMount()`

packages/apple-pay-integration/src/ApplePayPaymentMethod.tsx
```diff
@@ -0,0 +1,43 @@
...
+                    });
+                } catch (error) {
+                    onUnhandledError(error);
+                }
+            }
+            componentWillUnmount().then();
+        }
+    }, [checkoutService, language, method, onUnhandledError]);
+
+    return <React.Fragment />;
```
animesh1987, line 37: Should this return something other than a fragment?

packages/apple-pay-integration/src/ApplePayPaymentMethod.tsx
bc-peng: The functionality would stay the same, but it can help to get rid of the eslint warning, `Promise returned from componentDidMount is ignored`.

packages/apple-pay-integration/src/ApplePayPaymentMethod.tsx
bc-peng: In `<PaymentMethod/>` v1, we simply return `null`.
https://github.com/bigcommerce/checkout-js/blob/42ee27d679f81eaabc8b43771b08575525962d4a/packages/core/src/app/payment/paymentMethod/HostedPaymentMethod.tsx#L110

packages/apple-pay-integration/src/ApplePayPaymentMethod.tsx
DeveloperAlan: why not just do eslint ignore comment?

bc-peng: I have removed `.then()` as we actually intend to ignore that non-critical warning.

auto-export.config.json
davidchin: I prefer not aligning JSON keys because we don't want to re-align everything when we add a new key that is longer than the current longest key.

packages/apple-pay-integration/.eslintrc.json
```diff
@@ -0,0 +1,18 @@
+{
+  "extends": ["../../.eslintrc.json"],
+  "ignorePatterns": ["!**/*"],
+  "overrides": [
```
davidchin, line 4: Do we need `overrides` and `ignorePatterns` if nothing is different from the base config?

packages/apple-pay-integration/src/ApplePayPaymentMethod.spec.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+    it('initializes ApplePay with required props', () => {
+        const initializePayment = jest.spyOn(checkoutService,'initializePayment').mockResolvedValue(checkoutState);
+        const component = mount(<ApplePaymentMethod {...props} />);
+
+        expect(component.find(ApplePaymentMethod)).toHaveLength(1);
+        expect(initializePayment).toHaveBeenCalled()
+    });
+
+    it('deitializes ApplePay with required props', () => {
```
davidchin, line 36:
```suggestion
    it('deinitializes ApplePay with required props', () => {
```

packages/apple-pay-integration/src/ApplePayPaymentMethod.tsx
```diff
@@ -0,0 +1,43 @@
...
+                    });
+                } catch (error) {
+                    onUnhandledError(error);
+                }
+            };
+            deinitializePayment();
+        }
+    }, [checkoutService, language, method, onUnhandledError]);
+
+    return <React.Fragment />;
```
davidchin, line 37:
```suggestion
    return </>;
```

packages/samplePaymentMethod/project.json
```diff
@@ -0,0 +1,25 @@
+{
+  "root": "packages/samplePaymentMethod",
```
davidchin, line 2: Let's just remove the `samplePaymentMethod`. It only exists because I don't have sample. Now we have `ApplePay` so there's no need for it anymore.

packages/apple-pay-integration/project.json
```diff
@@ -0,0 +1,24 @@
...
+      "executor": "@nrwl/jest:jest",
+      "outputs": [
+        "coverage/packages/apple-pay-integration"
+      ],
+      "options": {
+        "jestConfig": "packages/apple-pay-integration/jest.config.js"
+      }
+    }
+  },
+  "tags": []
```
davidchin, line 23: Should we start introducing tags? Like `scope:integration`, `scope:core`, `scope:shared` etc... so we can impose [constraints](https://nx.dev/structure/monorepo-tags#imposing-constraints-on-the-project-graph) on what packages could be imported? cc @animesh1987

auto-export.config.json
bc-peng: oH, I see.

packages/apple-pay-integration/project.json
animesh1987: Yes, I think it's the right time to do it now 😄

bc-peng: On it.

packages/apple-pay-integration/.eslintrc.json
bc-peng: I'll remove `overrides` and keep `ignorePatterns`, because we set `"ignorePatterns": ["**/*"]` in `../../.eslintrc.json`, which is in line with `sdk`.

animesh1987: We should be adjusting that in sdk as well in that case. I will be cleaning that all up soon 😄

davidchin (approved):
Nice one, great work!

animesh1987 (approved):
Nice work @bc-peng

PR #954, opened 2022-08-03
CHECKOUT-6781: Load checkout buttons from integration packages

davidchin:
## What?
Load checkout buttons (a.k.a. wallet buttons) from various integration packages.

## Why?
So that they can be consumed and we can gradually extract members from `core` to individual integration packages.

## Testing / Proof
CircleCI

@bigcommerce/checkout


packages/core/src/app/customer/CheckoutButtonListV2.tsx
```diff
@@ -0,0 +1,54 @@
...
+> = ({
+    checkoutService,
+    checkoutState,
+    language,
+    onUnhandledError
+}) => {
+    const { 
+        statuses: { isInitializingCustomer },
+        data: { getConfig },
+    } = checkoutState;
```
animesh1987, line 25: Just out of curiosity, we usually do the pattern from other examples that we do the `mapToXYZComponentProps`. Is that something we will avoiding going ahead? 

Does it make sense going ahead to move this setup into a custom hook/context and use that?

animesh1987 (approved):

packages/core/src/app/customer/CheckoutButtonV2.spec.tsx
```diff
@@ -0,0 +1,71 @@
...
+            .toHaveBeenCalledWith({
+                methodId: 'foobar',
+                foobar: {
+                    container: 'button-container',
+                    onUnhandledError,
+                },
+            });
+    });
+
+    it('deinitializes when component is unmounted', () => {
```
bc-peng, line 42: Great work. The resolving mechanism is now better organized.

Just a question about wallet button itself: when should we expect a wallet button to be deinitialized or unmounted?

I can see Bolt deinitializes the previous Bolt customer strategy each time before initialization.

https://github.com/bigcommerce/checkout-js/blob/42ee27d679f81eaabc8b43771b08575525962d4a/packages/core/src/app/customer/checkoutSuggestion/BoltCheckoutSuggestion.tsx#L30

packages/core/src/app/customer/CheckoutButtonListV2.tsx
davidchin: We spoke offline. I'm fine with switching to using custom hooks rather than HOC. But currently we don't have an equivalent hook to use.

Regarding the props mapper, if there's nothing to filter, there's no need to write one. In this case, it's a simple passthrough because we need to propagate both the `checkoutService` and `checkoutState` to the inner component.

packages/core/src/app/customer/CheckoutButtonV2.spec.tsx
davidchin: `deinitialize` should only need to be called for components that have been initialized. In other words, when a component gets unmounted. We'll need to revisit Bolt once when get to creating an integration point for checkout suggestions. The scope of this PR is limited to just wallet buttons (which Bolt doesn't have).

bc-peng (approved):

packages/core/src/app/customer/CheckoutButtonV2.spec.tsx
animesh1987: @bc-peng Bolt doesn't work as a wallet button at the moment, it is encapsulated in it's own component called `<CheckoutSuggestion />` which is snowflake and should be fixed going forward separately.

PR #955, opened 2022-08-04
CHECKOUT-6781: Set up local plugin and generator for workspace

davidchin:
## What?
* Convert `auto-export` build script into a Nx generator.
* Create `workspace-tools`, local Nx plugin, to host `auto-export` generator, as well other generators or executors required by this workspace in the future.

## Why?
Currently the `auto-export` build script is executed as a part of Webpack build process. Considering Nx has now been set up, a better way to execute the task is via Nx instead, as it can better leverage build cache.

## Testing / Proof
CircleCI

@bigcommerce/checkout


package.json
```diff
@@ -144,7 +145,7 @@
     "ts-jest": "^25.5.1",
     "ts-loader": "^9.3.0",
     "tslib": "^1.10.0",
-    "typescript": "^3.9.7",
+    "typescript": "4.1.2",
```
bc-peng, line 148: 😃 Thanks.

packages/workspace-tools/.eslintrc.json
```diff
@@ -0,0 +1,33 @@
...
+    },
+    {
+      "files": [
+        "*.js",
+        "*.jsx"
+      ],
+      "rules": {}
+    }
+  ]
+}
```
bc-peng, line 33:
```suggestion
{
  "extends": ["../../.eslintrc.json"]
}
```
Think we can simplify it as a result of my last week's update to `../../.eslintrc.json`.

davidchin: Ah nice one, thanks! 👍

bc-peng (approved):
LGTM, thanks.

PR #956, opened 2022-08-04
feat(checkout): CHECKOUT-0000 Optimize Payment Integration JS API

bc-peng:
## What?
Move the following interfaces from `payment-integration` back to `core`:

- `PaymentMethodId`
- `PaymentMethodProviderType`
- `PaymentMethodType`
- `HostedPaymentMethodProps`

## Why?
A payment integration package (payment methods such as `ApplePay`) depends on `PaymentMethodProps` to automatically surface on the checkout page. Therefore, it is not necessary to include the above interfaces in the JS API.

By design, the packages owned by the checkout team do not need to know specific payment method IDs, types, and so forth.

## Testing / Proof
CI.

@bigcommerce/checkout


davidchin (approved):
LGTM! 👍

animesh1987 (approved):

PR #961, opened 2022-08-09
CHECKOUT-6781: Move default checkout button into its own package and allow GooglePay to override

davidchin:
## What?
* Move the default checkout button implementation (`CheckoutButton`) into its own package.
* Create a new package for Google Pay and implement `GooglePayButton` by extending from `CheckoutButton` and layering additional logic to handle embedded checkout.

## Why?
* We can load all the checkout buttons (default, Google Pay, Apple Pay etc...) in a consistent manner.
* Integrators can extend from the default button variant to create new variants.

## Testing / Proof
Unit

@bigcommerce/checkout

### DIFF
--- a/packages/checkout-button-integration/jest.config.js
+++ b/packages/checkout-button-integration/jest.config.js
@@ -1,0 +1,12 @@
+module.exports = {
+    displayName: 'checkout-button-integration',
+    preset: '../../jest.preset.js',
+    globals: {
+        'ts-jest': {
+            tsconfig: '<rootDir>/tsconfig.spec.json',
+            diagnostics: false,
+        }
+    },
+    setupFilesAfterEnv: ['../../jest-setup.ts'],
+    coverageDirectory: '../../coverage/packages/checkout-button-integration'
+};
--- a/packages/checkout-button-integration/project.json
+++ b/packages/checkout-button-integration/project.json
@@ -1,0 +1,27 @@
+{
+  "root": "packages/checkout-button-integration",
+  "sourceRoot": "packages/checkout-button-integration/src",
+  "projectType": "library",
+  "targets": {
+    "lint": {
+      "executor": "@nrwl/linter:eslint",
+      "outputs": ["{options.outputFile}"],
+      "options": {
+        "lintFilePatterns": ["packages/checkout-button-integration/**/*.{ts,tsx}"]
+      }
+    },
+    "test": {
+      "executor": "@nrwl/jest:jest",
+      "outputs": [
+        "coverage/packages/checkout-button-integration"
+      ],
+      "options": {
+        "jestConfig": "packages/checkout-button-integration/jest.config.js"
+      }
+    }
+  },
+  "tags": [
+    "scope:integration",
+    "scope:shared"
+  ]
+}
--- a/packages/checkout-button-integration/src/index.ts
+++ b/packages/checkout-button-integration/src/index.ts
@@ -1,0 +1,1 @@
+export { default as CheckoutButton } from './CheckoutButton';
--- a/packages/google-pay-integration/jest.config.js
+++ b/packages/google-pay-integration/jest.config.js
@@ -1,0 +1,12 @@
+module.exports = {
+    displayName: 'google-pay-integration',
+    preset: '../../jest.preset.js',
+    globals: {
+        'ts-jest': {
+            tsconfig: '<rootDir>/tsconfig.spec.json',
+            diagnostics: false,
+        }
+    },
+    setupFilesAfterEnv: ['../../jest-setup.ts'],
+    coverageDirectory: '../../coverage/packages/google-pay-integration'
+};
--- a/packages/google-pay-integration/project.json
+++ b/packages/google-pay-integration/project.json
@@ -1,0 +1,24 @@
+{
+  "root": "packages/google-pay-integration",
+  "sourceRoot": "packages/google-pay-integration/src",
+  "projectType": "library",
+  "targets": {
+    "lint": {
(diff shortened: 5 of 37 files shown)


packages/checkout-button-integration/project.json
```diff
@@ -0,0 +1,27 @@
...
+      "outputs": [
+        "coverage/packages/checkout-button-integration"
+      ],
+      "options": {
+        "jestConfig": "packages/checkout-button-integration/jest.config.js"
+      }
+    }
+  },
+  "tags": [
+    "scope:integration",
```
bc-peng, line 24: I didn't tag `payment-integration` with `scope:integration`. I assumed it was for a payment method not API. Do I need to add it to `payment-integration`?

packages/google-pay-integration/project.json
```diff
@@ -0,0 +1,27 @@
...
+        "coverage/packages/google-pay-integration"
+      ],
+      "options": {
+        "jestConfig": "packages/google-pay-integration/jest.config.js"
+      }
+    }
+  },
+  "tags": [
+    "scope:integration",
+    "scope:shared"
```
bc-peng, line 25: Same question, just want to confirm the tag usage.

animesh1987: Do we need to add multiple tags here?

davidchin: No, not for `google-pay-integration`, this is not intentional. But for `checkout-button-integration`, it is intentional because I want `checkout-button-integration` to be importable by `google-pay-integration` so it can be extended.

packages/checkout-button-integration/project.json
davidchin: No I think it's fine. As you said, `scope:integration` is for packages, when loaded, would integrate a payment method into checkout. Whereas `payment-integration-api` is a package consumed by the aforementioned packages to make the integration happen.

packages/google-pay-integration/project.json
animesh1987: Yes, it makes sense for checkout-button-integration :D

package-lock.json
```diff
@@ -1204,6 +1204,11 @@
             }
           }
         },
+        "tslib": {
```
animesh1987, line 1207: Wondering howcome there is a change here? Don't think anything was changed in package.json 🤔

packages/checkout-button-integration/jest.config.js
```diff
@@ -0,0 +1,17 @@
+module.exports = {
+    displayName: 'checkout-button-integration',
+    preset: '../../jest.preset.js',
+    globals: {
+        'ts-jest': {
+            tsconfig: '<rootDir>/tsconfig.spec.json',
+            diagnostics: false,
+        }
+    },
+    transform: {
```
animesh1987, line 10: Do we need to setup these transforms here? We can just have them defined in preset file, unless they are specific per app?

packages/checkout-button-integration/src/index.ts
```diff
@@ -0,0 +1 @@
+export { default as CheckoutButton } from './CheckoutButton';
```
animesh1987, line 1: EOL?

packages/checkout-button-integration/jest.config.js
davidchin: It's created by the generator. I'll remove 🙂

bc-peng (approved):
Nice job!!

packages/google-pay-integration/jest.config.js
```diff
@@ -0,0 +1,17 @@
+module.exports = {
+    displayName: 'google-pay-integration',
+    preset: '../../jest.preset.js',
+    globals: {
+        'ts-jest': {
+            tsconfig: '<rootDir>/tsconfig.spec.json',
+            diagnostics: false,
+        }
+    },
+    transform: {
```
animesh1987, line 10: Sorry missed this one in the previous one. We will need to document this part till we have out own generator to this work.

packages/payment-integration/src/errors/index.ts
```diff
@@ -0,0 +1,2 @@
+export { default as CustomError } from './CustomError';
+export { default as EmbeddedCheckoutUnsupportedError } from './EmbeddedCheckoutUnsupportedError';
```
animesh1987, line 2: EOL

tsconfig.base.json
```diff
@@ -21,15 +21,21 @@
     "noUnusedLocals": true,
     "noUnusedParameters": true,
     "paths": {
+      "@bigcommerce/checkout-js/apple-pay-integration": [
```
animesh1987, line 24: Maybe one for later, we discussed package names for SDK. Maybe we need to apply similar logic here?

animesh1987 (approved):
Nice work @davidchin  👍

packages/google-pay-integration/jest.config.js
davidchin: Sorry I forgot to take it out as well. As you suggested, it's now in the `preset` so it doesn't need to be redefined each time.

packages/payment-integration/src/errors/index.ts
davidchin: Argh, got to properly set up my editor on my new laptop hehe

tsconfig.base.json
davidchin: I think we end up settling with the `/` because these packages are never meant to be published via npm.

On a separate note, the parent package should be `@bigcommerce/checkout` rather than `@bigcommerce/checkout-js`. We rename all of them together in a separate PR.

PR #966, opened 2022-08-11
feat(checkout): CHECKOUT-6817 Update Sentry version and config

bc-peng:
## What?

- Apple Sentry sample rate, 10%.
- Upgrade Sentry to the latest version.

## Why?
We want to capture just enough error events to give us visibility on what’s broken without capturing too many.

## Testing / Proof
CircleCI.


packages/core/src/app/common/error/SentryErrorLogger.spec.ts
```diff
@@ -125,21 +128,25 @@ describe('SentryErrorLogger', () => {
...
+                    stacktrace: {
+                        frames: [
+                            {
+                                colno: 1234,
+                                filename: 'https://cdn.foo.bar/js/app-123.js',
+                                function: 't.<anonymous>',
+                                in_app: true,
+                                lineno: 1,
+                            },
+                        ],
```
bc-peng, line 142: Data structure changed here.

packages/core/src/app/common/error/SentryErrorLogger.spec.ts
```diff
@@ -154,45 +161,13 @@ describe('SentryErrorLogger', () => {
...
-                    },
-                ],
-            },
-        });
-
-        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-        const frame = output.stacktrace!.frames![0];
-
-        expect(frame)
-            .toEqual(frame);
```
bc-peng, line 193: Based on Sentry source code, they will rewrite Filename even if it does match with public path.
https://github.com/getsentry/sentry-javascript/blob/487810a32d6d6c5356386ea0b55b84a85675cb04/packages/integrations/src/rewriteframes.ts#L70

packages/core/src/app/common/error/SentryErrorLogger.ts
```diff
@@ -1,4 +1,4 @@
-import { captureException, init, withScope, BrowserOptions, Event, Integrations, Severity, StackFrame } from '@sentry/browser';
+import { captureException, init, withScope, BrowserOptions, Event, Integrations, SeverityLevel, StackFrame } from '@sentry/browser';
```
bc-peng, line 1: `Severity` is deprecated. 
https://github.com/getsentry/sentry-javascript/blob/master/packages/types/src/severity.ts

packages/core/src/app/common/error/SentryErrorLogger.spec.ts
```diff
@@ -154,45 +161,13 @@ describe('SentryErrorLogger', () => {
...
             .toHaveBeenCalledWith(expect.objectContaining({
-                blacklistUrls: [
+                denyUrls: [
                     'polyfill~checkout',
                     'sentry~checkout',
                 ],
             }));
     });
 
-    it('does not rewrite filename of error frames if it does match with public path', () => {
```
bc-peng, line 164: The rewriting will only happen when "the frame filename begins with `/` or a Windows-style prefix such as `C:\`", hence this test is no longer relevant.
https://github.com/getsentry/sentry-javascript/blob/487810a32d6d6c5356386ea0b55b84a85675cb04/packages/integrations/src/rewriteframes.ts#L70

davidchin: I think we still need it, as the same check exists in [5.27](https://github.com/getsentry/sentry-javascript/blob/250063170e0c871d52a83a414aeff7229857a28c/packages/integrations/src/rewriteframes.ts#L64). The test is intended to verify that this handler works as expected - https://github.com/bigcommerce/checkout-js/blob/3484d2aa1ab7ac131f089b90185dc4aff53d956e/packages/core/src/app/common/error/SentryErrorLogger.ts#L138 IIRC, due to the check mentioned above, the Sentry client doesn't automatically rewrite the frames and add the `app://` prefix so we got to do it manually ourselves.

DeveloperAlan (approved):

packages/core/src/app/common/error/SentryErrorLogger.spec.ts
bc-peng: Thanks @davidchin . I have restored it.
Just one more concern, the assertion of this test seems like always true. 
Did I miss anything?
```
const frame = output.exception!.values![0].stacktrace!.frames![0];

        expect(frame)
            .toEqual(frame);
```

davidchin: Hrm good catch @bc-peng, could you please change 
```ts
        expect(frame)
            .toEqual(frame);
```
to
```ts
        expect(frame)
            .toEqual({
                ...frame,
                filename: 'https://cdn.hello.world/js/app-123.js',
            });
```

packages/core/src/app/common/error/SentryErrorLogger.ts
```diff
@@ -78,20 +80,20 @@ export default class SentryErrorLogger implements ErrorLogger {
         });
     }
 
-    private mapToSentryLevel(level: ErrorLevelType): Severity {
+    private mapToSentryLevel(level: ErrorLevelType): SeverityLevel {
         switch (level) {
-        case ErrorLevelType.Info:
-            return Severity.Info;
+            case ErrorLevelType.Info:
+                return 'info';
```
animesh1987, line 86: This looks good, just out of curiosity should we have an enum to set all these values 
```
enum SeverityLevelEnum {
    INFO = 'info',
    ERROR = 'error',
    WARNING = 'warning',
    LOG = 'log',
    DEBUG = 'debug',
}
```

bc-peng: What is the concern here not to use Sentry's [type defintion](https://github.com/getsentry/sentry-javascript/blob/9b7f43246e75934c5b689f04147be788d3589fc2/packages/types/src/severity.ts#L22)?

packages/core/src/app/common/error/SentryErrorLogger.spec.ts
bc-peng: Hi @davidchin, I have updated `handleRewriteFrame` as well in [the new commit](https://github.com/bigcommerce/checkout-js/pull/966/commits/5100ee0f896361c4f78ff0968600dad1f4a41232).

In this test, `does not rewrite filename of error frames if it does match with public path`, the filename shares the same host as the public path.
In another test  few lines up, `configures client to rewrite filename of error frames`, the filename is on a different host.

davidchin: For that test, you need to keep the file name to be `https://cdn.hello.world/js/app-123.js` for both the input and the expected output.

bc-peng: Done. Did I get it right? 🙇🏻‍♂️
https://github.com/bigcommerce/checkout-js/pull/966/commits/d201e17579004f8a3a24357da51f2d5b2317ccad

packages/core/src/app/common/error/SentryErrorLogger.ts
animesh1987: We are still using Sentry type definition, since the method will still return a type `SeverityLevel`. This is just so we don't have strings in code and they are managed in one place.

bc-peng: Updated. I didn't include `LOG` cause we only have four `ErrorLevelType`.
https://github.com/bigcommerce/checkout-js/blob/42ee27d679f81eaabc8b43771b08575525962d4a/packages/core/src/app/common/error/ErrorLogger.ts#L40

packages/core/src/app/common/error/SentryErrorLogger.spec.ts
bc-peng: Because we changed the name of this test, I reverted the changes to `handleRewriteFrame` and the other test.

animesh1987 (approved):
Nice work @bc-peng 👏

davidchin (approved):
Nice one, LGTM! 👍

PR #967, opened 2022-08-11
fix(payment): CHECKOUT-5889 Fix issue with element detach modal

animesh1987:
## What?
Fix issue with an error modal being displayed if an element is detached

## Why?
At the moment we have an error modal displayed while a user transitions while hosted payment form is being loaded step to another step in checkout. 
Instead of showing the error modal and blocking the UX we can log the error using errorLogger.

## Testing / Proof
- Circle

https://user-images.githubusercontent.com/7134802/184576337-236008ac-62aa-4906-8f43-7e9ef6b94763.mov

@bigcommerce/checkout

### DIFF
--- a/packages/core/src/app/payment/Payment.tsx
+++ b/packages/core/src/app/payment/Payment.tsx
@@ -6,7 +6,7 @@ import React, { Component, ReactNode } from 'react';
 import { ObjectSchema } from 'yup';
 
 import { withCheckout, CheckoutContextProps } from '../checkout';
-import { isCartChangedError, isRequestError, ErrorModal, ErrorModalOnCloseProps } from '../common/error';
+import { isCartChangedError, isRequestError, ErrorModal, ErrorModalOnCloseProps, ErrorLogger } from '../common/error';
 import { EMPTY_ARRAY } from '../common/utility';
 import { withLanguage, WithLanguageProps } from '../locale';
 import { TermsConditionsType } from '../termsConditions';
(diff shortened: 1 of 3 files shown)


packages/core/src/app/payment/Payment.tsx
```diff
@@ -377,6 +379,23 @@ class Payment extends Component<PaymentProps & WithCheckoutPaymentProps & WithLa
...
+
+        const { type } = error as any;
+
+        if (type === 'unexpected_detachment') {
+            errorLogger.log(error);
+
+            return;
+        }
+        
+        return onUnhandledError(error);
```
bc-peng, line 396: All look good.
Error handling normally executes in a catch block. Is it necessary to have a `return` at the end?
```
try {
       ...
        } catch (error) {
            onUnhandledError(error);
        }
```

animesh1987: Thanks Peng for the feedback, here the error is already caught in SDK and passed to JS. Previously this error will be passed here and set in state of Checkout component.

Now we can handle it in child component i.e. `<Payment />` and hence the return. The rest of the errors are still passed to Checkout component.

bc-peng: > Now we can handle it in child component i.e. <Payment /> and hence the return.

I understand this is the reason for line 393. Do we need a `return` on line 396 to pass an error to Checkout component?

animesh1987: The return is to exit the function, otherwise the code continues and the error is passed to Checkout component. Which results in the error modal being displayed.

davidchin (approved):
LGTM 👍

packages/core/src/app/payment/Payment.tsx
bc-peng: > Which results in the error modal being displayed

Isn't that what we want for error types other than `unexpected_detachment`?

animesh1987: Yes, the error modal is still displayed for other error types. Since we pass the execution for the error to `<Checkout />` component.

bc-peng (approved):

PR #968, opened 2022-08-12
feat(checkout): CHECKOUT-6896 Add STOREURL as env variable

bc-peng:
## What?

- Add `STOREURL` as an env variable, eliminating the need to hardcode `storeUrl` in a test file.
- Add a `getStoreUrl()` helper.
- All HAR files now record the same generic store URL.
- As a benefit of removing `storeUrl` from a test, the template becomes lighter.
```
test.describe('xxxxxxxxxxxxxxxxxxxxxxx', () => {
    test('xxxxxxxxxxxxxxxxxxxxxxx', async ({assertions, checkout, page}) => {
        // Testing environment setup
        await checkout.start('name of the HAR file, which can be reused');

        // Playwright actions
        await checkout.goto();
        // await page.locator('text=Test Payment ProviderVisaAmexMaster').click();
        await checkout.placeOrder();

        // Assertions
        await assertions.shouldSeeOrderConfirmation();
    });
});
```

## Why?
`storeUrl` is unnecessary data stored in a test file.
Additionally, we do not want a test that has a strong tie to a certain dev store.

## Testing / Proof
CI.

### DIFF
--- a/tests/fixture/pageObject/Checkout.ts
+++ b/tests/fixture/pageObject/Checkout.ts
@@ -13,26 +13,27 @@ export class Checkout {
         this.playwright = new PlaywrightHelper(page);
     }
 
-    // This is for a developer to debug PollyJS replay quickly.
-    // Should not be used in a final test file.
-    log(): void {
-        this.playwright.enableDevMode();
-    }
-
+    // CheckoutFixtures helper, only used in ../CheckoutFixtures.ts
     async close(): Promise<void> {
         await this.playwright.stopAll();
     }
 
-    async create(har: string, storeUrl: string): Promise<void> {
-        await this.playwright.createCheckout(har, storeUrl);
+    // Dev helper
+    log(): void {
+        this.playwright.enableDevMode();
+    }
+
+    // Testing environment setup helpers
+    async use(preset: CheckoutPagePreset): Promise<void> {
+        await this.playwright.usePreset(preset);
+    }
+
+    async start(HAR: string): Promise<void> {
+        await this.playwright.createHAR(HAR);
     }
 
     async route(url: string | RegExp | ((url: URL) => boolean), filePath: string, data?: {}): Promise<void> {
         await this.playwright.renderAndRoute(url, filePath, data);
-    }
-
-    async use(preset: CheckoutPagePreset): Promise<void> {
-        await this.playwright.usePreset(preset);
     }
 
     // Abstract low-level HTML identifiers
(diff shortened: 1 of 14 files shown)


tests/fixture/pageObject/Checkout.ts
```diff
@@ -13,28 +13,29 @@ export class Checkout {
...
-    async close(): Promise<void> {
-        await this.playwright.stopAll();
+    // Testing environment setup helpers
+    async use(preset: CheckoutPagePreset): Promise<void> {
+        await this.playwright.usePreset(preset);
     }
 
-    async create(har: string, storeUrl: string): Promise<void> {
-        await this.playwright.createCheckout(har, storeUrl);
+    async start(HAR: string): Promise<void> {
```
bc-peng, line 31: Rename an essential step from `checkout.create()` to `checkout.start()` as we only need a HAR file name now, `start` also sounds more mandatory. We do `checkout.close()` automatically for all tests.

bc-peng, line 31: Rename an essential step from `checkout.create()` to `checkout.start()` as we only need a HAR file name now, `start` also sounds more mandatory. We do `checkout.close()` automatically for all tests.

tests/fixture/pageObject/playwright/PollyObject.ts
```diff
@@ -24,6 +26,7 @@ export class PollyObject {
     private readonly mode: MODE;
     private readonly baseUrl: string;
     private readonly bigpayBaseUrlIdentifier: string = '/api/public/v1/orders/payments';
+    private readonly genericStoreUrl:string = 'https://4241.project';
```
bc-peng, line 29: In memory of the payment integration project. 🤠

tests/fixture/pageObject/playwright/PlaywrightHelper.ts
```diff
@@ -12,16 +13,17 @@ export class PlaywrightHelper {
...
     private har: string = '';
-    private storeUrl: string = '';
 
     constructor(page: Page) {
         this.page = page;
         this.mode = process.env.MODE?.toLowerCase() === 'record'  ? 'record' : 'replay';
         this.isReplay = this.mode === 'replay';
         this.polly = new PollyObject(this.mode);
         this.server = new ServerSideRender();
+        this.storeUrl = (this.isReplay) ? '' : getStoreUrl();
```
davidchin, line 26: Probably don't need this conditional.

davidchin (approved):
LGTM, thanks for looking into this @bc-peng! 👏

tests/fixture/pageObject/playwright/PlaywrightHelper.ts
bc-peng: I have removed this control flow here and made `getStoreUrl()` mode-aware. 
As a result, it will complain about `STOREURL` undefined in `record` mode only.

PR #972, opened 2022-08-15
CHECKOUT-6899 Move method-specific interfaces out of payment integration API package

davidchin:
## What?
Move method-specific interfaces (e.g.: `IdealCustomFormFieldsetValues`, `SepaCustomFormFieldsetValues` etc...) out of payment integration API package

## Why?
They should belong to the method-specific integration package rather than the shared API package.

## Testing / Proof
CircleCI

@bigcommerce/checkout


packages/core/src/app/payment/paymentMethod/PaymentMethodTitle.tsx
```diff
@@ -211,13 +213,13 @@ const PaymentMethodTitle: FunctionComponent<PaymentMethodTitleProps & WithLangua
     const getSelectedCardType = () => {
         if (!isSelected) {
             return;
-        }
+        }
```
animesh1987, line 216: Is there extra spacing here?

packages/core/src/app/payment/paymentMethod/HostedCreditCardFieldsetValues.tsx
```diff
@@ -0,0 +1,34 @@
+import { CardInstrumentFieldsetValues } from "../storedInstrument";
```
animesh1987, line 1: Just out of curiosity does this need to be a `.tsx` file?

animesh1987:
Will we be adding a generator/executor in future to lift the payment method specific interfaces, if they are declared in payment method packages?

packages/core/src/app/payment/paymentMethod/HostedCreditCardFieldsetValues.tsx
davidchin: No reason. I just updated the file extension to be `.ts`

davidchin (commented):
> Will we be adding a generator/executor in future to lift the payment method specific interfaces, if they are declared in payment method packages?

No, there isn't because there's no need. Therefore, they could just live inside the payment method packages.

bc-peng (approved):

animesh1987 (approved):

PR #974, opened 2022-08-15
CHECKOUT-6899: Move mocks out of payment integration API package

davidchin:
## What?
Move mocks out of payment integration API package.

## Why?
Mocks used in tests should either belong to method-specific packages or exported from a "test utils" package.

## Testing / Proof
CircleCI

@bigcommerce/checkout


bc-peng (approved):
Thanks for the explanations.
It would be wonderful to have a "test utils" package later.

animesh1987 (approved):

DeveloperAlan (approved):

PR #977, opened 2022-08-15
feat(payment): PAYPAL-1613 removed unnecessary Braintree Venmo logic from PaymentSubmitButton component

serhii-tkachenko:
## What?
- Removed unnecessary Braintree Venmo logic from PaymentSubmitButton component;
- added `paypal_venmo_continue_action` translations for other languages;

## Why?
Because we don't need it anymore.

## Testing / Proof
Unit tests

### DIFF
--- a/packages/core/src/app/locale/translations/en.json
+++ b/packages/core/src/app/locale/translations/en.json
@@ -260,7 +260,6 @@
             "paypal_continue_action": "Continue with PayPal",
             "paypal_pay_later_continue_action": "Continue with Pay Later",
             "paypal_venmo_continue_action": "Continue with Venmo",
-            "braintreevenmo_continue_action": "Continue with Venmo",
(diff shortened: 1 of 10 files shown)


packages/core/src/app/locale/translations/en.json
```diff
@@ -260,7 +260,6 @@
             "paypal_continue_action": "Continue with PayPal",
             "paypal_pay_later_continue_action": "Continue with Pay Later",
             "paypal_venmo_continue_action": "Continue with Venmo",
-            "braintreevenmo_continue_action": "Continue with Venmo",
```
animesh1987, line 263: Do we need to do this for other language files as well?

serhii-tkachenko: Yes we do. Thanks for catch up!
I have found that we don't have `paypal_venmo_continue_action` translations for other languages..

animesh1987 (approved):

PR #978, opened 2022-08-15
fix(payment): INT-6404 AmazonPayV2: Prevent button flickering

mauricio-sg:
## What? [INT-6404](https://bigcommercecloud.atlassian.net/browse/INT-6404)
Wrap the Amazon Pay button in a container with a specific size to prevent it from flickering.

## Why?
The size of the button's container is calculated by Amazon's JS every time the window resizes, that's why it flickers. We prevent the script from doing that by providing a parent container with a specific size. Therefore, the button's container inherits this size instead of doing all the math.

#### Before:
https://user-images.githubusercontent.com/4843328/184449774-27043fe6-9bf8-4ffe-a2cd-33c5adf16040.mov
#### After:
https://user-images.githubusercontent.com/4843328/184711992-a61f3880-b2d0-4eaa-8c4b-8eb5ce1b90ec.mov

### DIFF
--- a/packages/core/src/scss/components/checkout/checkoutRemote/_checkoutRemote.scss
+++ b/packages/core/src/scss/components/checkout/checkoutRemote/_checkoutRemote.scss
@@ -11,6 +11,10 @@
     > div {
         margin-right: spacing("single");
     }
+
+    > .AmazonPayContainer {
+        width: remCalc(200px);
+    }
 }
 
 .checkoutRemote-button {
--- a/packages/core/src/app/customer/CheckoutButtonList.tsx
+++ b/packages/core/src/app/customer/CheckoutButtonList.tsx
@@ -4,7 +4,7 @@ import React, { memo, Fragment, FunctionComponent } from 'react';
 import { isApplePayWindow } from '../common/utility';
 import { TranslatedString } from '../locale';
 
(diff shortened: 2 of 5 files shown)


ggarcia-bc (approved):
LGTM 👍🏼

bc-lessgarmi:
:green_heart: Tested!
**QE** details:

The bug for Amazon Pay button has been validated without issue.
The following was tested:

- Amazon Pay, Google Pay and Apple Pay buttons in Checkout page.

Please see the attachment word below with the evidence:

https://docs.google.com/document/d/16p0_1upVtb_596SdPb00LohHBT_8VoDK/edit?usp=sharing&ouid=115824247328847092590&rtpof=true&sd=true 

Please let me know any comment or doubt

packages/core/src/scss/components/checkout/checkoutRemote/_checkoutRemote.scss
```diff
@@ -11,6 +11,10 @@
     > div {
         margin-right: spacing("single");
     }
+
+    > .AmazonPayContainer {
+        width: 200px;
```
animesh1987, line 16: Can we use rem or a built in utility for this?

PR #979, opened 2022-08-15
refactor(checkout): CHECKOUT-6896 Running e2e tests in CircleCI

bc-peng:
## What?
- Build test framework as a package.
Devs can utilize fixtures, presets by importing `@bigcommerce/checkout/payment-integration-test-framework`.

- ~~Build `e2e` package to host all e2e tests.~~
~~Because`nx` is not interested in supporting `playwright` (see `nx` issue [Introduce Playwright as option for E2E testing](https://github.com/nrwl/nx/issues/3390)), I believe it is better to host e2e tests in a single package for the time being.~~

- Put `e2e` test files into payment packages. The rules are:

   -  `playwright` will run all tests inside a payment package's `e2e` folder. 

   - `jest` will ignore all `e2e` folders.

- Establish `e2e` job in CircleCI.
   - If `e2e` job fails on CI, it will look like this: job [link](https://app.circleci.com/pipelines/github/bigcommerce/checkout-js/5048/workflows/9b97619d-ad00-411a-bb77-4f18e2473996/jobs/12965), artifacts [link](https://app.circleci.com/pipelines/github/bigcommerce/checkout-js/5048/workflows/9b97619d-ad00-411a-bb77-4f18e2473996/jobs/12965/artifacts), HTML test report [viewable link](https://output.circle-artifacts.com/output/job/1e155c0d-6389-4014-8ace-8cbcac68afa5/artifacts/0/artifact-file/index.html).
   - Also configured `playwright` to generate comprehansive trace-recordings([doc](https://playwright.dev/docs/trace-viewer)) for failed tests, which can be found on the artifacts page.

- Add `e2e` npm command.

- Update all previous HAR files. They all use a generic store URL now.

- Update many places in the test framework to eliminate `eslint` complaints.

## Why?

- We want to have integration tests to ensure that packages work together.
- We can better enforce test-related package boundaries. Checkout team owns the test-framework, while the integration team owns the e2e tests.


package.json
```diff
@@ -15,6 +15,8 @@
     "build:server": "http-server dist",
     "dev": "nx run core:dev",
     "dev:server": "http-server build",
+    "pree2e": "npx playwright install",
+    "e2e": "PORT=5005 MODE=REPLAY npx playwright test",
```
bc-peng, line 19: `e2e` command is for devs to run e2e tests locally, hence I put `PORT`, `MODE` and `npx playwright install` here.

playwright.config.ts
```diff
@@ -27,7 +27,7 @@ const config: PlaywrightTestConfig = {
   workers: process.env.CI ? 1 : undefined,
```
bc-peng, line 27: I have only set below env vars in CI now. So `workers` is `undefined`, which means `playwright` manages it automatically. We can always modify this config with a PR.
```
MODE = replay
PORT = 3000
```

package.json
```diff
@@ -15,6 +15,8 @@
     "build:server": "http-server dist",
     "dev": "nx run core:dev",
     "dev:server": "http-server build",
+    "pree2e": "npx playwright install",
```
animesh1987, line 18: Do we still need this command?

bc-peng: Otherwise, devs will always see browser packages are missing on the first run, and this command will come out on the terminal.

There is a way to avoid this by adding `playwright` to `package.json`, but it will increase the cache image size to 1.5G on CI and still need to run `npx playwright install` separately on CI. 😂

packages/e2e/project.json
```diff
@@ -0,0 +1,15 @@
+{
+  "root": "packages/e2e",
+  "sourceRoot": "packages/e2e/src",
```
bc-peng, line 3: Each payment method can place its e2e tests here as a folder. Test framework automatically generates HAR files in this package as well.

playwright.config.ts
animesh1987: Is it a big change for that? If not can we do it in this PR itself?

packages/payment-integration-test-framework/src/fixture/pageObject/playwright/ServerSideRender.ts
```diff
@@ -4,7 +4,7 @@ import ejs from 'ejs';
  * @internal
  */
 export class ServerSideRender {
-    async renderFile(filePath: string, data?: {}): Promise<string> {
+    async renderFile(filePath: string, data?: any): Promise<string> {
```
animesh1987, line 7: Is it possible to type this `data` property?

packages/payment-integration-test-framework/src/fixture/pageObject/playwright/PlaywrightHelper.ts
```diff
@@ -24,6 +25,7 @@ export class PlaywrightHelper {
         this.polly = new PollyObject(this.mode);
         this.server = new ServerSideRender();
         this.storeUrl = getStoreUrl();
+        this.srcPath = './packages/payment-integration-test-framework/src';
```
animesh1987, line 28: Do we need to do this in the constructor?

packages/payment-integration-test-framework/src/fixture/pageObject/Checkout.ts
```diff
@@ -32,7 +32,7 @@ export class Checkout {
         await this.playwright.createHAR(HAR);
     }
 
-    async route(url: string | RegExp | ((url: URL) => boolean), filePath: string, data?: {}): Promise<void> {
+    async route(url: string | RegExp | ((url: URL) => boolean), filePath: string, data?: any): Promise<void> {
```
animesh1987, line 35: Same question about `any`

package.json
animesh1987: 1.5G 😱

playwright.config.ts
bc-peng: Not a big change. I think the situation is we only have few tests to run. The workflow always finishes in less than 10 seconds. Hence not sure what to set.

package.json
bc-peng: When run `npm run e2e`, `npm` will automatically execute `pree2e` before `e2e`.
On another thought, we can combine `pree2e` and `e2e` into one command. Does this look better?

packages/payment-integration-test-framework/src/fixture/pageObject/playwright/ServerSideRender.ts
bc-peng: How about `Record<string, unknown>`?

packages/payment-integration-test-framework/src/fixture/pageObject/playwright/PlaywrightHelper.ts
bc-peng: `playwright` helper uses this prop three times for locating support files, such as HTML template.

What do you think about the package name, `payment-integration-test-framework`? I wanted to put it in line with `payment-integration-api`, but made it quite long😂.

playwright.config.ts
animesh1987: hmm maybe we can have 1 regardless in that case?

package.json
animesh1987: yes possibly better to combine them, less overhead.

packages/payment-integration-test-framework/src/fixture/pageObject/playwright/ServerSideRender.ts
animesh1987: Seems good to me, and then you can use typegurad to validate it.

packages/payment-integration-test-framework/src/fixture/pageObject/playwright/PlaywrightHelper.ts
animesh1987: Should be able to declare this variable after line 19, just need to declare it as public.

animesh1987: The package name seems fine to me, the `api` keyword doesn't feel right here since its not the api package. Its named correctly I think as of now :D

playwright.config.ts
bc-peng: All 3 sample tests run simultaneously on CI. Should be safe to set 3?

.circleci/config.yml
```diff
@@ -49,6 +49,21 @@ jobs:
...
+          name: "Build checkout-js"
+          command: npm run build
+      - run:
+          name: "Install playwright"
+          command: npx playwright install --with-deps
+      - run:
+          name: "Run e2e tests"
+          command: npx playwright test
+
   release:
```
bc-peng, line 67: I didn't put `e2e` as part of the release workflow. Do we want it there?

animesh1987: I guess we should put it there, since we do run tests as part of that too.

playwright.config.ts
bc-peng: Add `IS_CI` env var now.
```
workers: process.env.IS_CI ? 3 : undefined,
reporter: process.env.IS_CI ? 'line' : [ ['html', { outputFolder: './packages/payment-integration-test-framework/report' }] ],
```

.circleci/config.yml
```diff
@@ -49,6 +49,21 @@ jobs:
...
+      - node/npm-install
+      - run:
+          name: "Build checkout-js"
+          command: npm run build
+      - run:
+          name: "Install playwright"
+          command: npx playwright install --with-deps
+      - run:
+          name: "Run e2e tests"
+          command: npx playwright test
```
davidchin, line 65: After this step, could consider uploading the artifacts to CircleCI (screenshots for failed tests etc...) - see https://circleci.com/docs/artifacts#uploading-artifacts

workspace.json
```diff
@@ -4,8 +4,10 @@
     "apple-pay-integration": "packages/apple-pay-integration",
     "core": "packages/core",
     "checkout-button-integration": "packages/checkout-button-integration",
+    "e2e": "packages/e2e",
```
davidchin, line 7: The test files should live in the payment integration packages, e.g.: `google-pay`, not in a centralised package, so we can easily assign code owners.

.circleci/config.yml
bc-peng: Hi @davidchin , CircleCI is configured as suggested. 

Here is a demo for e2e job failure scenario: job [link](https://app.circleci.com/pipelines/github/bigcommerce/checkout-js/5048/workflows/9b97619d-ad00-411a-bb77-4f18e2473996/jobs/12965), artifacts [link](https://app.circleci.com/pipelines/github/bigcommerce/checkout-js/5048/workflows/9b97619d-ad00-411a-bb77-4f18e2473996/jobs/12965/artifacts), HTML test report [viewable link](https://output.circle-artifacts.com/output/job/1e155c0d-6389-4014-8ace-8cbcac68afa5/artifacts/0/artifact-file/index.html).

I've also configured `playwright` to generate comprehansive trace-recordings([doc](https://playwright.dev/docs/trace-viewer)) for failed tests, which can be found on the artifact page.

workspace.json
bc-peng: Moved test files into payment packages. The rules are set as below:

- `playwright` will run all tests inside a payment package's `e2e` folder.
- `jest` will ignore all `e2e` folders.

packages/google-pay-integration/e2e/Afterpay.spec.ts
```diff
@@ -1,4 +1,4 @@
-import { test, PaymentStepAsGuestPreset, UseAUDPreset } from '../';
+import { test, PaymentStepAsGuestPreset, UseAUDPreset } from '@bigcommerce/checkout/payment-integration-test-framework';
```
davidchin, line 1: Could you create a new `afterpay` package and move the test under there?

packages/google-pay-integration/e2e/BigpayTestPaymentProvider.spec.ts
```diff
@@ -1,4 +1,5 @@
-import { test, PaymentStepAsGuestPreset } from '../';
+import { test, PaymentStepAsGuestPreset } from '@bigcommerce/checkout/payment-integration-test-framework';
```
davidchin, line 1: Likewise, create a `hosted-credit-card` package and move the test in there.

packages/google-pay-integration/e2e/template.ts
```diff
@@ -1,20 +1,22 @@
-import { test, PaymentStepAsGuestPreset } from '../';
+import { test, PaymentStepAsGuestPreset } from '@bigcommerce/checkout/payment-integration-test-framework';
```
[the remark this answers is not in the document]
bc-peng: I moved the file there and commented out all code because `ts` forbids using `@bigcommerce/checkout/payment-integration-test-framework` within `payment-integration-test-framework`, and `eslint` also complains a lot.

davidchin (approved):
Great work @bc-peng, LGTM! 👍

PR #987, opened 2022-08-22
Feature/remove floating cart summary

macanhhuy:
## What?
Remove floating cart summary

## Why?
A/B TEST VARIATION HAS WON
UNHIDE AND EXPOSE TO ALL VISITORS
## Testing / Proof
...

@bigcommerce/checkout


animesh1987:
Hey @macanhhuy can you please explain what's the goal of this PR, this seems like a fairly substantial change by the looks of it. Which I don't think we want to do in the app as of now.

macanhhuy:
The goal of this PR is display the cart summary on top of checkout page on mobile by default. Because the AB Testing for this part is approved. So we don't need to hide it by default and display when we add class like before @animesh1987

PR #990, opened 2022-08-24
chore(payment): CHECKOUT-6904 Update resolver component to return method v1 as fallback

animesh1987:
## What?
As above

## Why?
Add fallback to v1 payment method component if a component cannot be resolved via the `resolveComponent` method

## Testing / Proof
- tests 
- manual

@bigcommerce/checkout

### DIFF
--- a/packages/core/src/app/payment/paymentMethod/PaymentMethodV2.tsx
+++ b/packages/core/src/app/payment/paymentMethod/PaymentMethodV2.tsx
@@ -15,7 +15,7 @@ export interface PaymentMethodProps {
     method: PaymentMethod;
     isEmbedded?: boolean;
     isUsingMultiShipping?: boolean;
-    resolveComponent?(query: PaymentMethodResolveId): ComponentType<ResolvedPaymentMethodProps>;
+    resolveComponent?(query: PaymentMethodResolveId): ComponentType<ResolvedPaymentMethodProps> | undefined;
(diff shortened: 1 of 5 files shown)


packages/core/src/app/payment/paymentMethod/PaymentMethodV2.tsx
```diff
@@ -8,11 +8,14 @@ import { withLanguage, WithLanguageProps } from '../../locale';
 import { withForm, WithFormProps } from '../../ui/form';
 import createPaymentFormService from '../createPaymentFormService';
 import resolvePaymentMethod from '../resolvePaymentMethod';
+import { default as PaymentMethodComponent } from './PaymentMethod';
```
davidchin, line 11:
```suggestion
import { default as PaymentMethodV1 } from './PaymentMethod';
```

packages/core/src/app/payment/paymentMethod/PaymentMethodV2.tsx
```diff
@@ -57,7 +62,13 @@ const PaymentMethodContainer: ComponentType<
     });
 
     if (!ResolvedPaymentMethod) {
-        return null;
+        console.log('Does this run');
```
davidchin, line 65: Can you remove the `console.log`?

davidchin (approved):
LGTM, just two minor comments 🙂

PR #991, opened 2022-08-24
feat(checkout): CHECKOUT-6851 Separate shipping option from address

bc-peng:
## What?
Make the shipping address and the shipping method separate.

## Why?
Currently, shipping method is appended at the end of the address. This makes it hard for shoppers to quickly identify it. We want the checkout flow to be less cluttered.

## Testing / Proof
Before:
Before
After:
After

### DIFF
--- a/packages/core/src/app/shipping/StaticConsignment.tsx
+++ b/packages/core/src/app/shipping/StaticConsignment.tsx
@@ -48,7 +48,7 @@ const StaticConsignment: FunctionComponent<StaticConsignmentProps> = ({
                         <strong>
                             <TranslatedString id="shipping.shipping_method_label" />
                         </strong> }
-                    <div className="shippingOption shippingOption--alt">
+                    <div className="shippingOption shippingOption--alt shippingOption--state-selected">
                         <StaticShippingOption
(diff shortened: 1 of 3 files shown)


DeveloperAlan (approved):

packages/core/src/app/shipping/StaticConsignment.tsx
```diff
@@ -48,7 +48,7 @@ const StaticConsignment: FunctionComponent<StaticConsignmentProps> = ({
                         <strong>
                             <TranslatedString id="shipping.shipping_method_label" />
                         </strong> }
-                    <div className="shippingOption shippingOption--alt">
+                    <div className="shippingOption shippingOption--alt shippingOption--state-selected">
```
animesh1987, line 51: This class name gives the impression that `state` is selected, can we just call it `shippingOption--selected`? Can't think of a better name at the moment :|

bc-peng: Renamed.

animesh1987 (approved):
Nice work @bc-peng

PR #996, opened 2022-08-28
feat(checkout): CHECKOUT-6875 Improve accessibility part I

bc-peng:
## What?

Orignial requirements doc ([link](https://docs.google.com/spreadsheets/d/19NjngW7dapPcaqFcgBXF0q2ofjqVESdX/edit?usp=sharing&ouid=105120289126066958944&rtpof=true&sd=true)).

This PR contains 4 out of 9 sub tasks that will break `BCapp` UI tests.

PR 1

## Related PR
Will update `BCapp` UI tests and bump `checkout-js` in this PR.
https://github.com/bigcommerce/bigcommerce/pull/48253

## Why?
BC merchants can benefit from the ADA-compliant native checkout and order confirmation pages.

## Testing / Proof
- BCapp UI tests after update:
   - [integration_bcapp_release_blocking_UI_test](https://app.circleci.com/pipelines/github/bigcommerce/bigcommerce/229818/workflows/e8a5401f-5a33-447a-af36-8ee488adcb7f): success
   - [integration_bcapp_checkout_test](https://app.circleci.com/pipelines/github/bigcommerce/bigcommerce/229820/workflows/3abfe03b-ed06-47de-bae7-ee12e095443c): success
- Manual testing:
   - Wrap `continue shopping` in a `form` for compliance.
   
   https://user-images.githubusercontent.com/88361607/187099466-c1bd047f-2af6-4c5c-8ffc-797a1186bb28.mov

   - Convert step header from `a` to `div`, no visual changes introduced.
   
   aToDiv

### DIFF
--- a/packages/core/src/app/checkout/CheckoutStepHeader.tsx
+++ b/packages/core/src/app/checkout/CheckoutStepHeader.tsx
@@ -29,10 +29,13 @@ const CheckoutStepHeader: FunctionComponent<CheckoutStepHeaderProps> = ({
     type,
 }) => {
     return (
-        <a
+        <div
             className={ classNames(
                 'stepHeader',
-                { 'is-readonly': !isEditable }
+                {
+                    'is-readonly': !isEditable,
+                    'is-clickable': isEditable && !isActive,
+                }
             ) }
             onClick={ preventDefault(isEditable && onEdit ? () => onEdit(type) : noop) }
         >
@@ -66,7 +69,7 @@ const CheckoutStepHeader: FunctionComponent<CheckoutStepHeaderProps> = ({
                     <TranslatedString id="common.edit_action" />
                 </Button>
             </div> }
-        </a>
+        </div>
     );
 };
 
--- a/packages/core/src/app/order/OrderConfirmation.tsx
+++ b/packages/core/src/app/order/OrderConfirmation.tsx
@@ -143,11 +143,11 @@ class OrderConfirmation extends Component<
                         }) }
 
                         <div className="continueButtonContainer">
-                            <a href={ siteLink } target="_top">
-                                <Button variant={ ButtonVariant.Secondary }>
+                            <form action={ siteLink } method="get" target="_top">
+                                <Button type="submit" variant={ ButtonVariant.Secondary }>
                                     <TranslatedString id="order_confirmation.continue_shopping" />
                                 </Button>
-                            </a>
+                            </form>
                         </div>
                     </div>
                 </div>
--- a/packages/core/src/app/order/OrderSummaryItem.tsx
+++ b/packages/core/src/app/order/OrderSummaryItem.tsx
@@ -35,26 +35,27 @@ const OrderSummaryItem: FunctionComponent<OrderSummaryItemProps> = ({
         </figure>
 
         <div className="product-column product-body">
-            <h5
+            <h4
                 className="product-title optimizedCheckout-contentPrimary"
                 data-test="cart-item-product-title"
             >
                 { `${quantity} x ${name}` }
-            </h5>
-            <ul
-                className="product-options optimizedCheckout-contentSecondary"
-                data-test="cart-item-product-options"
-            >
-                { (productOptions || []).map((option, index) =>
-                    <li
-                        className="product-option"
-                        data-test={ option.testId }
-                        key={ index }
-                    >
-                        { option.content }
-                    </li>
-                ) }
-            </ul>
+            </h4>
+            { (productOptions && productOptions.length>0) && <ul
+                    className="product-options optimizedCheckout-contentSecondary"
+                    data-test="cart-item-product-options"
+                >
+                    { productOptions.map((option, index) =>
+                        <li
+                            className="product-option"
+                            data-test={ option.testId }
+                            key={ index }
+                        >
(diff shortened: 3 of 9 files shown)


packages/core/src/app/order/OrderSummaryItem.tsx
```diff
@@ -35,26 +35,27 @@ const OrderSummaryItem: FunctionComponent<OrderSummaryItemProps> = ({
         </figure>
 
         <div className="product-column product-body">
-            <h5
+            <h4
```
bc-peng, line 38: Before: we use `h1`, `h2`, `h3`, `h5`. So it was complained that `checkout-js` does not provide appropriate heading levels in the page and maintain hierarchy of the headings.

After: we use  `h1`, `h2`, `h3`, `h4`.

packages/core/src/app/order/OrderSummaryItem.tsx
```diff
@@ -35,26 +35,27 @@ const OrderSummaryItem: FunctionComponent<OrderSummaryItemProps> = ({
...
-                        key={ index }
-                    >
-                        { option.content }
-                    </li>
-                ) }
-            </ul>
+            </h4>
+            { (productOptions && productOptions.length>0) && <ul
+                    className="product-options optimizedCheckout-contentSecondary"
+                    data-test="cart-item-product-options"
```
bc-peng, line 46: Hiding `cart-item-product-options` when no `produceOptions` will break current UI tests.

packages/core/src/app/order/OrderConfirmation.tsx
```diff
@@ -143,11 +143,11 @@ class OrderConfirmation extends Component<
...
 
                         <div className="continueButtonContainer">
-                            <a href={ siteLink } target="_top">
-                                <Button variant={ ButtonVariant.Secondary }>
+                            <form action={ siteLink } method="get" target="_top">
+                                <Button type="submit" variant={ ButtonVariant.Secondary }>
                                     <TranslatedString id="order_confirmation.continue_shopping" />
                                 </Button>
-                            </a>
+                            </form>
```
bc-peng, line 150: Changing `<button>` to `<a>` will make its text colour to be blue, even with the same class names. Hence I went for the `form` solution.

packages/core/src/app/checkout/CheckoutStepHeader.tsx
```diff
@@ -29,10 +29,10 @@ const CheckoutStepHeader: FunctionComponent<CheckoutStepHeaderProps> = ({
     type,
 }) => {
     return (
-        <a
```
DeveloperAlan, line 32: For accessibility, wouldn't screen readers prefer having the `a` tag as opposed to a clickable `div` tag

bc-peng: The current issue is that a screen reader might mislead a shopper by announcing our step header as a clickable link.
Regarding your concern, all editable (collapsed) steps have delicate `edit` buttons, it should be good to use them.

DeveloperAlan (approved):

packages/core/src/app/checkout/CheckoutStepHeader.tsx
animesh1987: Do we need to change the html element here?

packages/core/src/app/checkout/CheckoutStepHeader.tsx
```diff
@@ -29,10 +29,10 @@ const CheckoutStepHeader: FunctionComponent<CheckoutStepHeaderProps> = ({
...
+        <div
             className={ classNames(
                 'stepHeader',
-                { 'is-readonly': !isEditable }
+                { 'is-clickable': isEditable && !isActive }
```
animesh1987, line 35: Are we going to change the class name?

packages/core/src/app/order/OrderConfirmation.tsx
animesh1987: Again what was the issue with `a` before?

bc-peng: According to the audit doc, A `<Button>` should not not be in `a`. 
> "HTML does not allow an interactive element to be nested in a link. This will cause significant issues for some assistive technology.

> Remove the button markup and keep the link and the text. "

packages/core/src/app/checkout/CheckoutStepHeader.tsx
bc-peng: Think we have to get rid of `a` so that a screen reader won't annouce each step header as a link.

I'll add an expanded or collapsed state aria attribute in the next PR.

packages/core/src/app/checkout/CheckoutStepHeader.tsx
bc-peng: @animesh1987 I think we need to.

Previously, a step header is `a`, which has a pointer cursor by default. `is-readonly` specifies a default cursor for customer step, which cannot be edited, `sign out` only.

Now, a step header is a div, so `is-clickable` provides a pointer cursor for editable shipping and billing steps, and leaves customer step as non-editable like before.

animesh1987 (approved):

PR #999, opened 2022-09-02
feat(checkout): CHECKOUT-0000 New test framework features

bc-peng:
## What?
- Add playwright video recording option in the config. (turned off by default)
- Add check for essential helper functions usage.
- Add check for essential env variables.
- Add `checkout.selectPaymentMethod()` helper function.
- Add displays of active API requests when applying presets.
- Update afterpay, bigpay, googlepay sample tests.

## Why?
With better error-prevention mechanisms and helper functions, developers will find it eaiser to use the test framework.

## Testing / Proof
CI.


packages/afterpay-integration/e2e/Afterpay.spec.ts
```diff
@@ -1,16 +1,18 @@
...
         await checkout.use(new PaymentStepAsGuestPreset());
         await checkout.start('sample Afterpay AUD');
 
         // Playwright actions
         await checkout.goto();
+        await checkout.selectPaymentMethod('pay_by_installment');
 
         // Assertions
         await assertions.shouldSeePaymentStep();
+        await expect(await page.locator('#radio-afterpay-pay_by_installment').isChecked()).toBeTruthy()
```
bc-peng, line 16: Checking `checkout.selectPaymentMethod('pay_by_installment')` is actually working.

packages/core/src/app/ui/accordion/AccordionItem.tsx
```diff
@@ -46,7 +46,6 @@ const AccordionItem: FunctionComponent<AccordionItemProps> = ({
             className,
             { [classNameSelected]: isSelected }
         ) }
-        data-test={ `accordion-item_${itemId}` }
```
bc-peng, line 49: This is not working. `playwright` will not select a payment method after clicking this li.
It was working in Apple Pay e2e test because Apple Pay is the only option in the payment list.

Apple Pay test recording:

[e004646597fb864c31553020e45eae64.webm](https://user-images.githubusercontent.com/88361607/188043436-0cdc071b-21a1-4e0b-8536-be8375b04d82.webm)

playwright.config.ts
```diff
@@ -19,19 +19,23 @@ const config: PlaywrightTestConfig = {
   forbidOnly: true,
   retries: 0,
   workers: process.env.IS_CI ? 3 : undefined,
-  // reporter: 'line',
   reporter: [['html', { outputFolder: './packages/payment-integration-test-framework/report', open: 'never' }]],
   use: {
     baseURL: 'http://localhost:' + process.env.PORT,
     screenshot: 'only-on-failure',
     trace: 'retain-on-failure',
-    // actionTimeout: 30 * 1000, // Maximum time each action such as `click()` can take. Defaults to 0 (no limit). */
```
bc-peng, line 28: We have set 60s global timeout.

packages/afterpay-integration/e2e/Afterpay.spec.ts
animesh1987, line 16:
```suggestion
        await expect(await page.locator('#radio-afterpay-pay_by_installment').isChecked()).toBe(true)
```

package.json
```diff
@@ -15,7 +15,7 @@
     "build:server": "http-server dist",
     "dev": "nx run core:dev",
     "dev:server": "http-server build",
-    "e2e": "npx playwright install && PORT=5005 MODE=REPLAY npx playwright test",
+    "e2e": "npm run build && npx playwright install && PORT=5005 MODE=REPLAY npx playwright test",
```
bc-peng, line 18: @animesh1987 Could you please do another review?
I just added everything I can think of for the test framework.

packages/payment-integration-test-framework/src/fixture/pagePreset/ApiRequestsSender.ts
```diff
@@ -103,55 +117,76 @@ export class ApiRequestsSender {
...
+        console.log(`Disposing APIRequestContext`);
+
         const apiContext = await this.apiContextFactory.create(this.page, this.storeUrl);
 
         await apiContext.dispose();
+
+        const time = Math.ceil((Date.now() - this.startTime) / 1000);
+
+        console.log(`\x1b[32m\n  ${message} \x1b[0m\x1b[2m(${time}s)\x1b[0m`);
     }
```
bc-peng, line 175: Updated look of recording Bigpay testing payment method.
image

packages/payment-integration-test-framework/src/fixture/pageObject/Checkout.ts
```diff
@@ -32,8 +34,10 @@ export class Checkout {
     }
 
     async close(): Promise<void> {
```
animesh1987, line 36: Do we need to have a way to reset all the variables to `false`?

bc-peng: `!this.harFileName || !this.usedPreset || !this.openedCheckout` serves as a reminder for a developer to set a proper checkout envirnment and record a test. They should be always`true` at the end of every test.

Since I don't expect a complicated usa case, such as changing payment configuration in the middle of a test, it should be sufficient to check the essential functions have run at least once for now.

animesh1987 (approved):
Nice work @bc-peng

PR #1009, opened 2022-09-09
fix(checkout): CHECKOUT-6365 Add scrollIntoView to embedded checkout

bc-peng:
## What?
Add `scrollIntoView` to embedded checkout, so a shopper can always see the content of a modal.

## Why?
A shopper cannot see the content of a modal without scrolling to the top of the iframe when an embedded checkout's iframe is too long.

## Testing / Proof
Before:

https://user-images.githubusercontent.com/88361607/189245953-1258db75-7244-4999-96b5-674ae1b393c3.mov

After:

https://user-images.githubusercontent.com/88361607/189245959-1027d53c-e962-4d27-9872-30de307b1a31.mov

### DIFF
--- a/packages/core/src/app/ui/modal/ModalTrigger.tsx
+++ b/packages/core/src/app/ui/modal/ModalTrigger.tsx
@@ -49,6 +49,8 @@ export default class ModalTrigger extends Component<ModalTriggerProps, ModalTrig
     }
 
     private handleOpen: () => void = () => {
+        document.body.scrollIntoView();
+
         if (!this.canHandleEvent) {
             return;
         }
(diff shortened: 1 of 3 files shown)


packages/core/src/app/ui/modal/ModalTrigger.tsx
```diff
@@ -49,6 +49,8 @@ export default class ModalTrigger extends Component<ModalTriggerProps, ModalTrig
     }
 
     private handleOpen: () => void = () => {
+        document.body.scrollIntoView();
```
bc-peng, line 52: `<ModalTrigger>` is being used in four components.

This bug does not apply to the normal checkout because all modals have a fixed top position due to CSS. Only embedded checkout requires a method to scroll the top of the iframe into the current view.

animesh1987 (approved):
Nice and easy fix, thanks @bc-peng 😄

PR #1010, opened 2022-09-11
feat(checkout): STRF-9858 Integrate Bodl Service with Checkout Begin and Order Purchased

jairo-bc:
## What?

- Called checkout begin event for BODL on checkout page load
- Called order purchsed event for BODL on order confirmation page load 

## Why?

It's required for BODL phase 1 

## Testing / Proof
Screenshot 2022-09-12 at 10 00 56
Screenshot 2022-09-12 at 10 07 29

@bigcommerce/checkout


packages/core/src/app/checkout/CheckoutApp.tsx
```diff
@@ -67,4 +68,8 @@ export default class CheckoutApp extends Component<CheckoutAppProps> {
     private createStepTracker: () => StepTracker = () => {
         return createStepTracker(this.checkoutService);
     };
+
+    private createBodlService: (subscriber: (state: CheckoutSelectors) => void) => BodlService = (subscriber) => {
+        return createBodlService(subscriber);
```
animesh1987, line 73: You should be able to do this here
```suggestion
        return createBodlService(this.checkoutService.subscribe);
```

animesh1987 (approved):

PR #1015, opened 2022-09-13
fix(checkout): CHECKOUT-6860 Use numeric keypad to type phone number

bc-peng:
## What?
Using a numeric keypad to enter tel numbers.

## Why?
Speed up the checkout.

## Testing / Proof

@bigcommerce/checkout

### DIFF
--- a/packages/core/src/app/ui/form/DynamicFormField.spec.tsx
+++ b/packages/core/src/app/ui/form/DynamicFormField.spec.tsx
@@ -44,6 +44,31 @@ describe('DynamicFormField Component', () => {
             }));
     });
 
+    it('renders telephone field', () => {
+        const telField = {
+            custom: false,
+            fieldType: "text",
+            id: 'field_17',
+            label: 'Phone Number',
+            name: 'phone',
+            required: true,
+            type: "string",
+        };
+        const component = mount(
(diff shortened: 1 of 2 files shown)


icatalina:
Number is not a good keyboard for telephone, can we get something like `type="tel"` instead?

bc-peng:
> Number is not a good keyboard for telephone, can we get something like `type="tel"` instead?

@icatalina  We do use `type="tel"`, the final html code on the shipping form is:
```
<input aria-labelledby="phoneInput-label phoneInput-field-error-message" autocomplete="tel" id="phoneInput" name="shippingAddress.phone" class="form-input optimizedCheckout-form-input" type="tel" data-test="phoneInput-text" value="">
```

icatalina:
Why is it showing the `,` and `.` then? I'd expect the `tel` type to look a bit more like this: 

IMAGE 2022-09-13 12:48:25

icatalina:
Is that an emulator thing? 🤔

bc-peng:
> Is that an emulator thing? 🤔

It turns out the kepad in the description is a custom form field with `type="number"`, the telephone number's keypad in the shipping looks like this on my Pixel. I have updated the description. Thanks for the catch.

animesh1987 (approved):

PR #1017, opened 2022-09-13
feat(payment): STRIPE-25 create a custom guest form for Stripe link

AlanMunozN:
## What? [STRIPE-25](https://bigcommercecloud.atlassian.net/browse/STRIPE-25)

- Create logic to render Stripeupe customer component
- Create auto-step logic
- Look and feel
- logic to show only credit card in case StripeLink is activated and is an authenticated Stripe customer

## Why?
As a Guest Shopper, I want to use the Stripe link experience.

## Testing / Proof
**HOW TO SETUP**

## Depends on
https://github.com/bigcommerce/checkout-js/pull/1018
https://github.com/bigcommerce/checkout-sdk-js/pull/1616

### DIFF
--- a/packages/core/src/app/checkout/Checkout.spec.tsx
+++ b/packages/core/src/app/checkout/Checkout.spec.tsx
@@ -95,11 +95,8 @@ describe('Checkout', () => {
         );
 
         jest.spyOn(checkoutService, 'getState').mockImplementation(() => checkoutState);
-        jest.spyOn(checkoutService, 'loadPaymentMethods')
-            .mockResolvedValue(checkoutService.getState());
 
-        jest.spyOn(checkoutService, 'getState')
-            .mockImplementation(() => checkoutState);
+        jest.spyOn(checkoutService, 'loadPaymentMethods').mockResolvedValue(checkoutService.getState());
 
         jest.spyOn(checkoutService, 'subscribe').mockImplementation((subscriber) => {
             subscribeEventEmitter.on('change', () => subscriber(checkoutService.getState()));
--- a/packages/core/src/app/checkout/Checkout.spec.tsx
+++ b/packages/core/src/app/checkout/Checkout.spec.tsx
@@ -95,6 +95,11 @@ describe('Checkout', () => {
         );
 
         jest.spyOn(checkoutService, 'getState').mockImplementation(() => checkoutState);
+        jest.spyOn(checkoutService, 'loadPaymentMethods')
+            .mockResolvedValue(checkoutService.getState());
+
+        jest.spyOn(checkoutService, 'getState')
+            .mockImplementation(() => checkoutState);
 
         jest.spyOn(checkoutService, 'subscribe').mockImplementation((subscriber) => {
             subscribeEventEmitter.on('change', () => subscriber(checkoutService.getState()));
--- a/packages/core/src/app/checkout/Checkout.tsx
+++ b/packages/core/src/app/checkout/Checkout.tsx
@@ -126,7 +126,6 @@ export interface CheckoutState {
     isRedirecting: boolean;
     hasSelectedShippingOptions: boolean;
     isBuyNowCartEnabled: boolean;
-    isStripeLinkAuthenticated: boolean;
 }
 
 export interface WithCheckoutProps {
@@ -165,7 +164,6 @@ class Checkout extends Component<
         isMultiShippingMode: false,
         hasSelectedShippingOptions: false,
         isBuyNowCartEnabled: false,
-        isStripeLinkAuthenticated: false,
     };
 
     private embeddedMessenger?: EmbeddedCheckoutMessenger;
@@ -351,11 +349,6 @@ class Checkout extends Component<
 
     private renderCustomerStep(step: CheckoutStepStatus): ReactNode {
         const { isGuestEnabled } = this.props;
-
-        const updateStripeLinkAuthenticated = (auth: boolean = false) => {
-            this.setState({ isStripeLinkAuthenticated: auth });
-        }
-
         const {
             customerViewType = isGuestEnabled ? CustomerViewType.Guest : CustomerViewType.Login,
         } = this.state;
@@ -388,7 +381,6 @@ class Checkout extends Component<
                         onSignInError={this.handleError}
                         onUnhandledError={this.handleUnhandledError}
                         step={step}
-                        updateStripeLinkAuthenticated={updateStripeLinkAuthenticated}
                         viewType={customerViewType}
                     />
                 </LazyContainer>
@@ -478,7 +470,6 @@ class Checkout extends Component<
                         checkEmbeddedSupport={this.checkEmbeddedSupport}
                         errorLogger={errorLogger}
                         isEmbedded={isEmbedded()}
-                        isStripeLinkAuthenticated={this.state.isStripeLinkAuthenticated}
                         isUsingMultiShipping={
                             cart && consignments
                                 ? isUsingMultiShipping(consignments, cart.lineItems)
--- a/packages/core/src/app/checkout/Checkout.tsx
+++ b/packages/core/src/app/checkout/Checkout.tsx
@@ -126,6 +126,7 @@ export interface CheckoutState {
     isRedirecting: boolean;
     hasSelectedShippingOptions: boolean;
     isBuyNowCartEnabled: boolean;
+    isStripeLinkAuthenticated: boolean;
 }
 
 export interface WithCheckoutProps {
@@ -164,6 +165,7 @@ class Checkout extends Component<
         isMultiShippingMode: false,
         hasSelectedShippingOptions: false,
         isBuyNowCartEnabled: false,
+        isStripeLinkAuthenticated: false,
     };
 
     private embeddedMessenger?: EmbeddedCheckoutMessenger;
@@ -350,6 +352,10 @@ class Checkout extends Component<
     private renderCustomerStep(step: CheckoutStepStatus): ReactNode {
         const { isGuestEnabled } = this.props;
 
+        const updateStripeLinkAuthenticated = (auth: boolean = false) => {
+            this.setState({ isStripeLinkAuthenticated: auth });
+        }
+
         const {
             customerViewType = isGuestEnabled ? CustomerViewType.Guest : CustomerViewType.Login,
         } = this.state;
@@ -381,6 +387,8 @@ class Checkout extends Component<
                         onSignIn={this.navigateToNextIncompleteStep}
                         onSignInError={this.handleError}
                         onUnhandledError={this.handleUnhandledError}
+                        step={step}
+                        updateStripeLinkAuthenticated={updateStripeLinkAuthenticated}
                         viewType={customerViewType}
                     />
                 </LazyContainer>
@@ -470,6 +478,7 @@ class Checkout extends Component<
                         checkEmbeddedSupport={this.checkEmbeddedSupport}
                         errorLogger={errorLogger}
                         isEmbedded={isEmbedded()}
+                        isStripeLinkAuthenticated={this.state.isStripeLinkAuthenticated}
                         isUsingMultiShipping={
                             cart && consignments
                                 ? isUsingMultiShipping(consignments, cart.lineItems)
--- a/packages/core/src/app/customer/Customer.spec.tsx
+++ b/packages/core/src/app/customer/Customer.spec.tsx
@@ -13,6 +13,7 @@ import React, { FunctionComponent } from 'react';
 import { getBillingAddress } from '../billing/billingAddresses.mock';
 import { CheckoutProvider } from '../checkout';
 import { getCheckout } from '../checkout/checkouts.mock';
+import CheckoutStepType from '../checkout/CheckoutStepType';
 import { getStoreConfig } from '../config/config.mock';
 import { createLocaleContext, LocaleContext, LocaleContextType } from '../locale';
 
@@ -23,6 +24,7 @@ import CustomerViewType from './CustomerViewType';
 import EmailLoginForm from './EmailLoginForm';
 import GuestForm, { GuestFormProps } from './GuestForm';
 import LoginForm, { LoginFormProps } from './LoginForm';
+import StripeGuestForm from './StripeGuestForm';
 
 describe('Customer', () => {
     let CustomerTest: FunctionComponent<CustomerProps & Partial<WithCheckoutCustomerProps>>;
@@ -90,6 +92,46 @@ describe('Customer', () => {
 
             expect(component.find(GuestForm).exists()).toBe(true);
         });
+
+        it('renders stripe guest form if enable', async () => {
+            const steps = { isActive: true,
+                isComplete: true,
+                isEditable: true,
+                isRequired: true,
+                type: CheckoutStepType.Customer };
+
+            const component = mount(
+                <CustomerTest isStripeLinkEnabled={ true } step={ steps } viewType={ CustomerViewType.Guest } />
+            );
+
+            await new Promise(resolve => process.nextTick(resolve));
+            component.update();
+
+            expect(component.find(StripeGuestForm).exists()).toEqual(true);
+        });
+
+        it('calls onUnhandledError if initialize was failed', async () => {
+            jest.spyOn(checkoutService, 'initializeCustomer').mockRejectedValue(new Error());
+            const unhandledError = jest.fn();
+
+            mount(<CustomerTest onUnhandledError={ unhandledError } viewType={ CustomerViewType.Guest } />);
+            await new Promise(resolve => process.nextTick(resolve));
+
+            expect(unhandledError).toHaveBeenCalledWith(expect.any(Error));
+        });
+
+        it('calls onUnhandledError if deinitialize was failed', async () => {
+            jest.spyOn(checkoutService, 'deinitializeCustomer').mockRejectedValue(new Error());
+            const unhandledError = jest.fn();
+
+            const component = mount(<CustomerTest onUnhandledError={ unhandledError } viewType={ CustomerViewType.Guest }/>);
+            await new Promise(resolve => process.nextTick(resolve));
+            component.unmount();
+            await new Promise(resolve => process.nextTick(resolve));
+
+            expect(unhandledError).toHaveBeenCalled();
+        });
+
 
         it('renders guest form if billing address is undefined', async () => {
             jest.spyOn(checkoutService.getState().data, 'getBillingAddress').mockReturnValue(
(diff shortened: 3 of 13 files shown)


packages/core/src/app/customer/Customer.spec.tsx
```diff
@@ -90,6 +92,46 @@ describe('Customer', () => {
                 .toEqual(true);
         });
 
+        it('renders stripe guest form if enable', async () => {
```
FranciscoOnt, line 95: small typo

```suggestion
        it('renders stripe guest form if enabled', async () => {
```

packages/core/src/app/customer/Customer.tsx
```diff
@@ -433,16 +475,26 @@ export function mapToWithCheckoutCustomerProps(
...
     const checkout = getCheckout();
     const customer = getCustomer();
     const signInEmail = getSignInEmail();
     const config = getConfig();
+    let stripeUpeLinkEnabled = false;
+
+    if (cart) {
+        const stripeUpe = getPaymentMethod('card', PaymentMethodId.StripeUPE);
+        const linkEnabled = stripeUpe?.initializationData.enableLink || false;
+        const stripeUpeSupportedCurrency = cart?.currency.code === 'USD' || false;
```
FranciscoOnt, line 494: since this block is inside if (cart), the optional chaining is not required
```suggestion
        const stripeUpeSupportedCurrency = cart.currency.code === 'USD' || false;
```

packages/core/src/app/customer/StripeGuestForm.spec.tsx
```diff
@@ -0,0 +1,199 @@
...
+        );
+        const button = component.find('[data-test="stripe-customer-continue-as-guest-button"]');
+
+        expect(button.prop('disabled')).toBeTruthy();
+    });
+
+    it('executes a function when on click action is triggered', () => {
+        const component = mount(
+            <TestComponent />
+        );
```
FranciscoOnt, line 72: this can be made in one line 🍹

packages/core/src/app/customer/StripeGuestForm.tsx
```diff
@@ -0,0 +1,249 @@
...
+}) => {
+
+    const [continueAsAGuestButton, setContinueAsAGuestButton] = useState(true);
+    const [emailValue, setEmailValue] = useState('');
+    const [authentication, setAuthentication] = useState(false);
+    const [isStripeLoading, setIsStripeLoading] = useState(true);
+    const [isNewAuth, setIsNewAuth] = useState(false);
+    const handleOnClickSubmitButton = () => {
+        onContinueAsGuest({
+            email: emailValue || '',
```
FranciscoOnt, line 57: is`|| ''` required? the state is already initialized as an empty string 🍹

packages/core/src/app/payment/Payment.tsx
```diff
@@ -489,10 +490,10 @@ class Payment extends Component<PaymentProps & WithCheckoutPaymentProps & WithLa
     };
 }
 
-export function mapToPaymentProps({
-    checkoutService,
-    checkoutState,
-}: CheckoutContextProps): WithCheckoutPaymentProps | null {
+export function mapToPaymentProps(
+    { checkoutService, checkoutState,}: CheckoutContextProps,
```
FranciscoOnt, line 494: you can remove the trailing comma in the second element 🍹

FranciscoOnt (approved):
Just tome small style changes, but looks good overall 👍

packages/core/src/app/payment/Payment.tsx
```diff
@@ -519,7 +520,9 @@ export function mapToPaymentProps({
     const customer = getCustomer();
     const consignments = getConsignments();
     const { isComplete = false } = getOrder() || {};
-    const methods = getPaymentMethods() || EMPTY_ARRAY;
+    const stripeUpePaymentMethod = getPaymentMethod('card', PaymentMethodId.StripeUPE);
+    const methods = props.isStripeLinkAuthenticated && stripeUpePaymentMethod ?
```
animesh1987, line 524: Should this happen on payments endpoint? Or we want to do this in frontend?

packages/core/src/app/payment/paymentMethod/HostedPaymentMethod.spec.tsx
```diff
@@ -93,6 +94,24 @@ describe('HostedPaymentMethod', () => {
...
+    it('calls onUnhandledError if deinitialize was failed', async () => {
+        defaultProps.deinitializePayment = jest.fn(() => {
+            throw new Error();
+        });
+
+        mount(<HostedPaymentMethodTest { ...defaultProps } />).unmount();
+
+        expect(defaultProps.onUnhandledError).toHaveBeenCalledWith(expect.any(Error));
+    });
+
```
bc-peng, line 114: I can see we are testing error cases here. Could we put this file in the payment PR?

packages/core/src/app/checkout/Checkout.tsx
```diff
@@ -295,6 +297,10 @@ class Checkout extends Component<CheckoutProps & WithCheckoutProps & WithLanguag
     private renderCustomerStep(step: CheckoutStepStatus): ReactNode {
         const { isGuestEnabled } = this.props;
 
+        const updateStripeLinkAuthenticated = (auth: boolean = false) => {
+            this.setState({ isStripeLinkAuthenticated: auth });
```
animesh1987, line 301: hey @AlanMunozN this seems ok, but after another look this feels like can be moved to sdk state. Since I can see we use this in payments step below. And possibly in shipping step in next PR as well.

packages/core/src/app/customer/StripeGuestForm.tsx
```diff
@@ -0,0 +1,249 @@
...
+                                    disabled={ continueAsAGuestButton }
+                                    id="stripe-checkout-customer-continue"
+                                    isLoading={ isLoading }
+                                    onClick={ handleOnClickSubmitButton }
+                                    testId="stripe-customer-continue-as-guest-button"
+                                    type="submit"
+                                    variant={ ButtonVariant.Primary }
+                                >
+                                    <TranslatedString id={ buttonText }/>
+                                </Button> }
```
bc-peng, line 210: Once we are able to get `stripeUpeLinkEnabled` and `authentication` from `checkout-sdk-js`, we can add this area to the current `<GuestForm/>`. In this way, we can avoid having a custom guest form for Stripe.

packages/core/src/app/checkout/Checkout.tsx
AlanMunozN: Hi @animesh1987, do you mean something like this? https://github.com/bigcommerce/checkout-sdk-js/compare/master...AlanMunozN:checkout-sdk-js:STRIPE-52-peng

packages/core/src/app/payment/paymentMethod/HostedPaymentMethod.spec.tsx
AlanMunozN: Hey @bc-peng, sorry for the late answer, I think this Pr contains the Payment and customer code, the other one only contains the shipping code, I don't know if you added this comment in the wrong PR

BC-AOrtiz: hey @bc-peng what do you mean about the payment PR?, we only have two PRs, one with the customer changes and  other wiht the shipping changes

packages/core/src/app/payment/Payment.tsx
BC-AOrtiz: hey @animesh1987  We do it in the front because in the back we don't have the way to know when the shopper is authenticated with his Link account

packages/core/src/app/payment/paymentMethod/HostedPaymentMethod.spec.tsx
bc-peng: @AlanMunozN @BC-AOrtiz I was assuming we had a payment PR for this ticket. Apology for the misunderstanding. Please focus on other comments.

packages/core/src/app/customer/StripeGuestForm.tsx
AlanMunozN: Hey @bc-peng, we followed your suggestions and got rid of the stripeGuestForm https://github.com/bigcommerce/checkout-js/pull/1017/commits/c247ba279c5aac133034f67164b350c12162f779, please let us know if you agree with the approach

packages/core/src/app/checkout/Checkout.tsx
bc-peng: @AlanMunozN Appreciate your effort. Now we can use checkout context to get `isStripeLinkAuthenticated` from SDK directly, no need to set an addtional prop here. You simply need:
1. import context:
`import checkoutContext from "../../checkout/CheckoutContext";`
2. Use context to get checkout state, for example, getting the value of feature flag `CHECKOUT-3790.upgrade_checkout_payment_step`:
```
const context = useContext(checkoutContext);
const flagValue = context?.checkoutState.data.getConfig()?.checkoutSettings?.features['CHECKOUT-3790.upgrade_checkout_payment_step']
```

packages/core/src/app/customer/GuestForm.tsx
```diff
@@ -38,7 +49,213 @@ const GuestForm: FunctionComponent<GuestFormProps & WithLanguageProps & FormikPr
...
+    const [emailValue, setEmailValue] = useState('');
+    const [isStripeLoading, setIsStripeLoading] = useState(true);
+    const [isNewAuth, setIsNewAuth] = useState(false);
+    const handleOnClickSubmitButton = () => {
+        onContinueAsGuest({
+            email: emailValue,
+            shouldSubscribe: false,
+        });
+    };
+    const buttonText = isStripeLinkAuthenticated && !isNewAuth? 'customer.continue_as_stripe_customer_action' : continueAsGuestButtonLabelId;
```
bc-peng, line 71: What if we were to add another `StripeLink`-like product to checkout,  current `buttonText`. `setEmailCallback` and style solutions would make things not sustainable.

We can discuss a proper `GuestFormProps` interface to support `StripeLink` on the following meeting calls.
cc @animesh1987

packages/core/src/app/payment/Payment.tsx
animesh1987: Sorry I haven't had time to test the paypal flow, but as far as I know if we click paypal button on cart page. We filter other providers on checkout payment step. 
@serhii-tkachenko do you know how that logic is setup?

serhii-tkachenko: As far as I know, provider setup step triggers on bcapp side.
When customer submits paypal flow we are posting the form through formPoster here:
https://github.com/bigcommerce/checkout-sdk-js/blob/master/packages/core/src/checkout-buttons/strategies/paypal-commerce/paypal-commerce-button-strategy.ts#L87

Then, on the backend side we are selecting specific provider (and returning only paypal in array of payment providers for payment step). You can double check this code on bcapp:
https://github.com/bigcommerce/bigcommerce/blob/01f0b9e7b92e3cdf75cd03bda50d838d1de3a219/modules/checkout/paypalcommerce/module.paypalcommerce.php#L727

AlanMunozN: Hey @animesh1987 , we tried to do different things to support this in the backend but we’re detecting multiple issues with this approach:
	1. Currently the use of setRemoteProvider (through set_external_checkout) as other payment providers are using, is not prepared for MultiOption Providers, currently, this flow only accepts the gateway, and the APM it's not.
	2. Now, Stripe Link doesn't support the authentication after a page reload (because it is not preserved on the stripe side, that part is in their scope but not worked yet). So checking the flow that PayPal is following, requires a page reload to load the remote provider.
	3. To support remoteProvider for MultiOption requires a bigger effort because it touches multiple repos (Bcapp, Checkout-SDK & Bigpay) and before we start to code something we need to discuss together with the payments team because apparently, this approach touches core files.
Saying this, we are in a very complicated position because the timeline is super tight and there is a contractual date and financial agreement that is a big pressure for us to deliver this. So I propose the following actions:
	1. Send the code that works good (adding a validation into Payment.tsx) and add a TODO comment that we're gonna change that approach to get the provider by the backend after the certification and as fast follow, also adding the ticket number (this ticket will be created if you agree of this :pray)  where it will be addressed.
    2. Create a thread to discuss these problematics related to RemoteProvider and the flow that we're trying to implement with Stripe Link, to include multiple BC teams that maybe can give us a better context and ideas to solve the problem/concern you're describing and after the certification from Stripe.

I hope you understand the complex situation we have related to the deadline (Oct 14th in Prod).

Let me know if you have any other questions about it, 

CC: @clopezh, @vmparra, @FranciscoOnt

packages/core/src/app/checkout/Checkout.tsx
AlanMunozN: @bc-peng, we'll wait for the resolution of the preview comment because could affect the complete approach

packages/core/src/app/customer/GuestForm.tsx
animesh1987: Can we move all the stripe related logic to its own component? `<StripeGuestForm />` with the props that are only required by that component.

packages/core/src/app/checkout/Checkout.spec.tsx
```diff
@@ -95,6 +95,11 @@ describe('Checkout', () => {
         );
 
         jest.spyOn(checkoutService, 'getState').mockImplementation(() => checkoutState);
+        jest.spyOn(checkoutService, 'loadPaymentMethods')
+            .mockResolvedValue(checkoutService.getState());
+
+        jest.spyOn(checkoutService, 'getState')
+            .mockImplementation(() => checkoutState);
```
FranciscoOnt, line 102: Isn't this a duplicate of line 97?

packages/core/src/app/customer/Customer.tsx
```diff
@@ -158,10 +167,43 @@ class Customer extends Component<CustomerProps & WithCheckoutCustomerProps, Cust
             isInitializing = false,
             privacyPolicyUrl,
             requiresMarketingConsent,
+            isStripeLinkEnabled,
             onUnhandledError = noop,
+            step,
+            updateStripeLinkAuthenticated,
+
```
FranciscoOnt, line 174: empty line

FranciscoOnt (changes requested):

packages/core/src/app/customer/Customer.tsx
```diff
@@ -158,10 +166,39 @@ class Customer extends Component<CustomerProps & WithCheckoutCustomerProps, Cust
...
             onUnhandledError = noop,
+            step,
         } = this.props;
 
         return (
+            isStripeLinkEnabled ?
+                <StripeGuestForm
+                    canSubscribe={ canSubscribe }
+                    checkoutButtons={
+                        <CheckoutButtonList
```
animesh1987, line 179: We can remove the checkout buttons here?

packages/core/src/app/customer/StripeGuestForm.tsx
```diff
@@ -0,0 +1,246 @@
...
+import { TranslatedString } from '../locale';
+import { PrivacyPolicyField } from '../privacyPolicy';
+import { Button, ButtonVariant } from '../ui/button';
+import { BasicFormField, Fieldset, Legend } from '../ui/form';
+import { LoadingOverlay } from '../ui/loading';
+
+import { GuestFormValues } from './GuestForm';
+import SubscribeField from './SubscribeField';
+
+export interface StripeGuestFormProps {
```
animesh1987, line 16: Please can we review the interface here and only pass what's needed for stripe link component.

packages/core/src/app/payment/Payment.tsx
```diff
@@ -524,15 +524,22 @@ export function mapToPaymentProps({
...
     const { isComplete = false } = getOrder() || {};
-    const methods = getPaymentMethods() || EMPTY_ARRAY;
+    let methods = getPaymentMethods() || EMPTY_ARRAY;
+    const stripeUpePaymentMethod = methods.filter(method =>
+        method.id === 'card' && method.gateway === PaymentMethodId.StripeUPE
+    );
+
+    if (isStripeLinkAuthenticated() && !!stripeUpePaymentMethod.length){
+        methods = stripeUpePaymentMethod
+    }
```
animesh1987, line 542: Why don't `isStripeLinkAuthenticated` is authenticated first? And then go down the flow of filtering cards.

packages/core/src/app/customer/StripeGuestForm.tsx
```diff
@@ -0,0 +1,246 @@
...
+        properties: string[]) => {
+        const parentContainer = document.getElementById(id);
+        if (parentContainer) {
+            return getAppliedStyles(parentContainer, properties);
+        } else {
+            return undefined;
+        }
+    };
+
+    const getStripeStyles: any = useCallback( () => {
```
animesh1987, line 91: Can we avoid `any`? Maybe something like`Record<string, string>`

packages/core/src/app/customer/StripeGuestForm.tsx
```diff
@@ -0,0 +1,246 @@
...
+            fieldPlaceholderText: formInput.color,
+            fieldErrorText: formError.color,
+            fieldBackground: formInput['background-color'],
+            fieldInnerShadow: formInput['box-shadow'],
+            fieldBorder: formInput['border-color'],
+        } : undefined;
+    }, [])
+
+    const renderCheckoutThemeStylesForStripeUPE = () => {
+        const containerId = 'stripe-card-component-field';
```
animesh1987, line 108: Can possibly move this to a constant in the file above the component declaration.

packages/core/src/app/customer/StripeGuestForm.tsx
```diff
@@ -0,0 +1,246 @@
...
+        );
+    };
+
+    const stripeDeinitialize = () => {
+        deinitialize({
+            methodId: 'stripeupe',
+        });
+    };
+
+    const stripeInitialize = () => {
```
animesh1987, line 138: Move this to the top of the file after line 77

packages/core/src/app/customer/Customer.tsx
```diff
@@ -458,11 +487,21 @@ export function mapToWithCheckoutCustomerProps({
...
     const config = getConfig();
+    let stripeUpeLinkEnabled = false;
+
+    if (cart) {
+        const stripeUpe = getPaymentMethod('card', PaymentMethodId.StripeUPE);
+        const linkEnabled = stripeUpe?.initializationData.enableLink || false;
+        const stripeUpeSupportedCurrency = cart.currency.code === 'USD' || false;
+
+        stripeUpeLinkEnabled = linkEnabled && stripeUpeSupportedCurrency;
+    }
```
animesh1987, line 504: @BC-AOrtiz we need to add another TODO here, this should be driven by backend, same as other wallet buttons.

animesh1987 (approved):

BCHMercado:
💚  testing done !

PR #1018, opened 2022-09-13
feat(payment): STRIPE-52 Auto-Select Address and mount shipping component for Stripe Link

AlanMunozN:
## What? [STRIPE-52](https://bigcommercecloud.atlassian.net/browse/STRIPE-52)

- Create logic to render Stripeupe shipping component
- Create auto-step logic
- Look and feel

## Why?
As a Guest Shopper, I want to use the Stripe link experience.

## Testing / Proof
**HOW TO SETUP**

## Depends on
https://github.com/bigcommerce/checkout-js/pull/1017


FranciscoOnt (approved):

packages/core/src/app/checkout/Checkout.tsx
```diff
@@ -425,6 +425,7 @@ class Checkout extends Component<
                         onSignIn={this.handleShippingSignIn}
                         onToggleMultiShipping={this.handleToggleMultiShipping}
                         onUnhandledError={this.handleUnhandledError}
+                        step={step}
```
animesh1987, line 428: Same I think we need to avoid passing step here.

packages/core/src/app/shipping/Shipping.tsx
```diff
@@ -119,38 +133,106 @@ class Shipping extends Component<ShippingProps & WithCheckoutShippingProps, Ship
...
-                        updateAddress={updateShippingAddress}
+        const {
+            isInitializing,
+            isStripeLoading,
+            isStripeAutoStep,
+        } = this.state;
+
+        const renderShipping = () => {
+            if (isStripeLinkEnabled && !customer.email) {
+                return <div className="checkout-form">
```
animesh1987, line 149: Create a separate component, same as the customer part.

packages/core/src/app/shipping/Shipping.tsx
```diff
@@ -119,38 +133,106 @@ class Shipping extends Component<ShippingProps & WithCheckoutShippingProps, Ship
...
+                            isMultiShippingMode={ isMultiShippingMode }
+                            onMultiShippingChange={ this.handleMultiShippingModeSwitch }
+                            shouldShowMultiShipping={ shouldShowMultiShipping }
+                        />
+
+                        <LoadingOverlay
+                            isLoading={ isInitializing }
+                            unmountContentWhenLoading
+                        >
+                            <ShippingForm
```
animesh1987, line 166: This can be it's own component with props, and can be inside the main component body.

packages/core/src/app/shipping/Shipping.tsx
```diff
@@ -119,38 +133,106 @@ class Shipping extends Component<ShippingProps & WithCheckoutShippingProps, Ship
...
+                    </LoadingOverlay>
             </div>
-        );
+
+        }
+
+        return ( renderShipping() );
+    }
+
+    private handleIsStripeLoading: () => void = () => {
```
animesh1987, line 230:
```suggestion
    private stripeLoadedCallback: () => void = () => {
```

packages/core/src/app/shipping/ShippingAddress.tsx
```diff
@@ -59,9 +75,105 @@ const ShippingAddress: FunctionComponent<ShippingAddressProps> = (props) => {
...
+        postalCode: '',
+        phone: '',
+        customFields: [],
+    });
+
+    const handleLoading = useCallback(() => {
+        if (isStripeLoading) {
+            isStripeLoading();
+        }
+    }, []);
```
animesh1987, line 110: Need to pass the dependency here

packages/core/src/app/shipping/Shipping.tsx
```diff
@@ -85,13 +94,17 @@ class Shipping extends Component<ShippingProps & WithCheckoutShippingProps, Ship
...
+            isStripeAutoStep: false,
+
         };
     }
 
     async componentDidMount(): Promise<void> {
         const {
             loadShippingAddressFields,
             loadShippingOptions,
+            loadPaymentMethods,
```
animesh1987, line 107: We should not be loading payment methods as part of shipping step.

packages/core/src/app/shipping/StripeStateMapper.ts
```diff
@@ -0,0 +1,232 @@
+export default function StripeStateMapper(country: string, state: string): string {
```
animesh1987, line 1: what's the purpose of this file? Can't we use the geographies data?

packages/core/src/app/shipping/StripeupeShippingAddress.tsx
```diff
@@ -0,0 +1,59 @@
+import { CheckoutSelectors, FormField, ShippingInitializeOptions, ShippingRequestOptions } from '@bigcommerce/checkout-sdk';
+import { noop } from 'lodash';
+import React, { PureComponent, ReactNode } from 'react';
+
+import './StaticAddressEditable.scss';
+
+export interface StripeupeShippingAddressProps {
+    formFields: FormField[];
```
animesh1987, line 8: We don't use this in this file, so do we even need this?

animesh1987: Can we please have a look at interfaces and only use what is needed.

animesh1987:
I have a question for this PR, can we make a stripe shipping component? 

Since right now we pass the stripe change throughout the component tree where as from looking at the demo video, for a stripe link authenticated shopper we done even show the shipping step. So it should be a fairly small change in my opinion as compared to making all shipping components stripe aware.

packages/core/src/app/shipping/StripeStateMapper.ts
AlanMunozN: Hey @animesh1987, the purpose of this file is to map the state that Stripe is sending us because it returns a non-standard format example: For the country MX, if the user wants to select "Ciudad de Mexico" Stripe returns "CDMX" as the state code, on the other hand, BC uses "MEX" as state code

packages/core/src/app/shipping/Shipping.tsx
```diff
@@ -119,38 +134,96 @@ class Shipping extends Component<ShippingProps & WithCheckoutShippingProps, Ship
...
-
-        return (
-            <div className="checkout-form">
+        const {
+            isInitializing,
+            isStripeLoading,
+            isStripeAutoStep,
+        } = this.state;
+
+        const stripeShipping = () => {
```
animesh1987, line 148: Move this to a component as well

animesh1987 (approved):

BCHMercado:
💚  testing done !

PR #1036, opened 2022-09-27
fix(checkout): CHECKOUT-6365 Scroll to modal in WordPress

bc-peng:
## What?
For iOS, when a user clicks on "show details" button, the parent page will scroll to the top of the checkout iframe.

## Why?
`ScrollIntoView` does not work in iOS due to CSS conflicts with the Bigcommerce WordPress plug-in. Therefore, adding a method specific to the Bigcommerce WordPress plug-in.

## Testing / Proof
iOS Safari:

https://user-images.githubusercontent.com/88361607/192404635-4e7d93d5-9eec-43b9-bb7e-38b422c2ae47.mov

iOS Chrome:

https://user-images.githubusercontent.com/88361607/192404642-8fb877ec-b795-4d44-8aad-7f7c106ed538.mov

### DIFF
--- a/packages/core/src/app/ui/modal/ModalTrigger.tsx
+++ b/packages/core/src/app/ui/modal/ModalTrigger.tsx
@@ -14,6 +14,8 @@ export interface ModalTriggerState {
     isOpen: boolean;
 }
 
+const wordpressCheckout = parent.document.getElementById('bc-embedded-checkout');
+
 export default class ModalTrigger extends Component<ModalTriggerProps, ModalTriggerState> {
     state = {
         isOpen: false,
@@ -49,7 +51,6 @@ export default class ModalTrigger extends Component<ModalTriggerProps, ModalTrig
     }
 
     private handleOpen: () => void = () => {
-        const wordpressCheckout = parent.document.getElementById('bc-embedded-checkout');
         if (wordpressCheckout) {
             const top = wordpressCheckout.getBoundingClientRect().top;
             window.parent.scrollBy(0,top);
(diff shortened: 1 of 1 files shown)


packages/core/src/app/ui/modal/ModalTrigger.tsx
```diff
@@ -49,7 +49,13 @@ export default class ModalTrigger extends Component<ModalTriggerProps, ModalTrig
     }
 
     private handleOpen: () => void = () => {
-        document.body.scrollIntoView();
+        const wordpressCheckout = parent.document.getElementById('bc-embedded-checkout');
```
bc-peng, line 52: `bc-embedded-checkout` is the name of the embedded checkout container ([sample embedded checkout in Wordpress](https://bcsandbox.jabctools.com/checkout/)).

packages/core/src/app/ui/modal/ModalTrigger.tsx
```diff
@@ -49,7 +49,13 @@ export default class ModalTrigger extends Component<ModalTriggerProps, ModalTrig
...
+        if (wordpressCheckout) {
+            const top = wordpressCheckout.getBoundingClientRect().top;
+            window.parent.scrollBy(0,top);
```
bc-peng, line 55: These two lines currently uncovered by any unit test. Not sure how to test it.😂

packages/core/src/app/ui/modal/ModalTrigger.tsx
animesh1987: Do you think move this to a const outside? Since if we do change the container id will be easier to track.

animesh1987 (approved):

bc-peng:
Will release this PR under the ticket CHECKOUT-6969, then come back to add unit tests.

PR #1041, opened 2022-09-28
fix(checkout): CHECKOUT-6969 Display cart summary at bottom

bc-peng:
## What?
Revert previous PR https://github.com/bigcommerce/checkout-js/pull/1036
Not scroll any more, display cart summary at bottom.

## Why?
By default, modern browsers prevent cross-domain iframes from accessing anything in the parent window. 

The previous PR is unable to work in a cross domain scenario, which is the most common use case for the embedded checkout.

As a result, this PR provides a pure CSS solution to the problem that a shopper cannot see the cart summary when clicking on the cart button.

## Testing / Proof

Android - Chrome

https://user-images.githubusercontent.com/88361607/192906250-e95d8b48-2015-4ff8-a6e3-e010e4159a60.MP4

iOS - Safari

https://user-images.githubusercontent.com/88361607/192906254-eb203e66-109b-4ba6-ac61-d6d87c562059.mov

iOS - Chrome

https://user-images.githubusercontent.com/88361607/192906258-a5ea1d78-27a2-4762-8ec5-758841ecf846.mov

mac - Safari

### DIFF
--- a/packages/core/src/app/order/OrderSummaryDrawer.spec.tsx
+++ b/packages/core/src/app/order/OrderSummaryDrawer.spec.tsx
@@ -20,7 +20,6 @@ describe('OrderSummaryDrawer', () => {
     let localeContext: LocaleContextType;
 
     beforeEach(() => {
-        window.HTMLElement.prototype.scrollIntoView = jest.fn();
         localeContext = createLocaleContext(getStoreConfig());
 
         order = getOrder();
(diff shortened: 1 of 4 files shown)


snaderiBC (approved):
Not bad, not bad at all 👏 👏 👏 👏 
image

animesh1987 (approved):

animesh1987:
This fix is good, but we need to look at the white space at top for safari screens.

bc-peng:
> This fix is good, but we need to look at the white space at top for safari screens.

I tried to remove the white background and shadow of the container, then it looked like this.



However, I'm not sure if this will break anything. So I'll stick to my original plan and leave this option for the future.

PR #1049, opened 2022-10-06
feat(checkout): CHECKOUT-4944 Add floating labels

bc-peng:
## What?
Use floating labels on checkout when the experiment, `CHECKOUT-6879.enable_floating_labels`, is on.

## Why?
Speed up checkout, reduce the length of the page.

## Testing / Proof
Frame 1

### DIFF
--- a/packages/core/src/app/address/AddressForm.scss
+++ b/packages/core/src/app/address/AddressForm.scss
@@ -16,7 +16,7 @@
         display: block;
         flex: 1;
         flex-basis: 100%;
-        margin-bottom: $floating-form-field-spacing;
+        margin-bottom: spacing("half");
         max-width: 100%;
         padding: 0 spacing("half");
 
@@ -26,6 +26,7 @@
     }
 
     .dynamic-form-field.floating-form-field {
+        margin-bottom: $floating-form-field-spacing;
         padding: 0 0.5 * $floating-form-field-spacing;
     }
 
--- a/packages/core/src/app/address/AddressForm.scss
+++ b/packages/core/src/app/address/AddressForm.scss
@@ -1,5 +1,10 @@
 @import "../ui/Base";
 
+.checkout-address:has(.floating-form-field) {
+    margin-left: -0.23rem;
+    margin-right: -0.23rem;
+}
+
 .checkout-address {
     display: flex;
     flex-flow: row wrap;
@@ -18,6 +23,10 @@
         &:last-child {
             margin-bottom: 0;
         }
+    }
+
+    .dynamic-form-field.floating-form-field {
+        padding: 0 4px;
     }
 
     .dynamic-form-field--province,
--- a/packages/core/src/app/address/AddressForm.scss
+++ b/packages/core/src/app/address/AddressForm.scss
@@ -1,8 +1,8 @@
 @import "../ui/Base";
 
 .checkout-address:has(.floating-form-field) {
-    margin-left: -0.23rem;
-    margin-right: -0.23rem;
+    margin-left: -0.5 * $floating-form-field-spacing;
+    margin-right: -0.5 * $floating-form-field-spacing;
 }
 
 .checkout-address {
@@ -16,7 +16,7 @@
         display: block;
         flex: 1;
         flex-basis: 100%;
-        margin-bottom: spacing("half");
+        margin-bottom: $floating-form-field-spacing;
         max-width: 100%;
         padding: 0 spacing("half");
 
@@ -26,7 +26,7 @@
     }
 
     .dynamic-form-field.floating-form-field {
-        padding: 0 4px;
+        padding: 0 0.5 * $floating-form-field-spacing;
     }
 
     .dynamic-form-field--province,
--- a/packages/core/src/app/common/utility/isFloatingLabelEnabled.ts
+++ b/packages/core/src/app/common/utility/isFloatingLabelEnabled.ts
@@ -1,9 +1,5 @@
-import { CheckoutContextProps } from '../../checkout';
+import { CheckoutSettings } from '@bigcommerce/checkout-sdk';
 
-export default function isFloatingLabelEnabled(context: CheckoutContextProps): boolean {
-    return (
-        context.checkoutState.data.getConfig()?.checkoutSettings.features[
-            'CHECKOUT-6879.enable_floating_labels'
-        ] ?? false
-    );
+export default function isFloatingLabelEnabled(checkoutSettings: CheckoutSettings): boolean {
+    return !!checkoutSettings.features['CHECKOUT-6879.enable_floating_labels'];
 }
--- a/packages/core/src/app/common/utility/isFloatingLabelEnabled.ts
+++ b/packages/core/src/app/common/utility/isFloatingLabelEnabled.ts
@@ -1,0 +1,9 @@
+import { CheckoutContextProps } from '../../checkout';
+
+export default function isFloatingLabelEnabled(context: CheckoutContextProps): boolean {
+    return (
+        context.checkoutState.data.getConfig()?.checkoutSettings.features[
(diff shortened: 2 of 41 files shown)


packages/core/src/app/customer/Customer.spec.tsx
```diff
@@ -107,11 +107,12 @@ describe('Customer', () => {
             await new Promise(resolve => process.nextTick(resolve));
             component.update();
 
-            expect(component.find(StripeGuestForm).exists()).toEqual(true);
```
bc-peng, line 110: `Prettier` modified a lot of lines after running `npm run lint`.
Put them into the PR as these lines would be modified anyway.

packages/core/src/app/customer/EmailField.tsx
```diff
@@ -1,36 +1,42 @@
...
+        <TextInput
+            { ...props.field }
+            autoComplete={ props.field.name }
+            id={ props.field.name }
+            type="email"
+        />
+    ), []);
 
-    const labelContent = useMemo(() => <TranslatedString id="customer.email_label" />, []);
+    const useFloatingLabel = isFloatingLabelEnabled(useCheckoutContext());
```
bc-peng, line 26: I used `isFloatingLabelEnabled()` helper function to switch floating lables on/off.

theromulans:
🍹  Would be great to see a Before vs After view to demonstrate this great improvement even more so 🙏

packages/core/src/app/common/utility/isFloatingLabelEnabled.ts
```diff
@@ -0,0 +1,9 @@
+import { CheckoutContextProps } from '../../checkout';
+
+export default function isFloatingLabelEnabled(context: CheckoutContextProps): boolean {
+    return (
+        context.checkoutState.data.getConfig()?.checkoutSettings.features[
+            'CHECKOUT-6879.enable_floating_labels'
+        ] ?? false
+    );
+}
```
bc-peng, line 9: `isFloatingLabelEnabled()` is a helper function to switch floating lables on/off.

packages/core/src/app/ui/form/FormField.tsx
```diff
@@ -26,18 +27,25 @@ const FormField: FunctionComponent<FormFieldProps> = ({
...
+                        htmlFor={name}
+                        id={`${id ?? name}-label`}
+                        useFloatingLabel={useFloatingLabel}
+                    >
                         {labelContent}
                     </Label>
                 )}
 
-                {input(props)}
+                {!useFloatingLabel && input(props)}
```
bc-peng, line 48: The entire floating label implementation is based on the `+` CSS selector, which is used to apply floating styles to the label that follows a text field.

Because there is no CSS selector to do the opposite of `+` (see [stackoverflow](https://stackoverflow.com/questions/15449622/opposite-of-adjacent-sibling-selector)), we must change the placement of `<Label/>` according to the state of A/B testing.

packages/core/src/app/common/utility/isFloatingLabelEnabled.ts
```diff
@@ -0,0 +1,9 @@
+import { CheckoutContextProps } from '../../checkout';
+
+export default function isFloatingLabelEnabled(context: CheckoutContextProps): boolean {
```
davidchin, line 3: Instead of passing `CheckoutContextProps`, which contains both `CheckoutService` and `CheckoutState`, I'd just pass the minimum that you need, which is i.e.: `settings: CheckoutSettings`

packages/core/src/app/customer/__snapshots__/GuestForm.spec.tsx.snap
```diff
@@ -30,8 +30,6 @@ exports[`GuestForm matches snapshot 1`] = `
             >
               <label
                 class="form-label optimizedCheckout-form-label"
-                for="email"
```
davidchin, line 33: How come we're losing the `for/id`? Is this intentional?

packages/core/src/app/ui/form/DynamicInput.tsx
```diff
@@ -49,23 +52,40 @@ const DynamicInput: FunctionComponent<DynamicInputProps & WithDateProps> = ({
...
-                    {placeholder && <option value="">{placeholder}</option>}
-                    {options &&
-                        options.map(({ label, value: optionValue }) => (
-                            <option key={optionValue} value={optionValue}>
-                                {label}
-                            </option>
-                        ))}
-                </select>
+                <>
+                    <svg
```
davidchin, line 56: Could this be an `Icon` component?

packages/core/src/scss/components/foundation/forms/_forms.scss
```diff
@@ -28,12 +28,98 @@
...
+        }
+
+        .floating-textarea {
+            background-color: $floating-label-bg-color;
+            padding-top: $input-verticalPadding * 2.5;
+        }
+
+        .dropdown-arrow,
+        .floating-dropdown-arrow {
+            height: 6px;
```
animesh1987, line 73: Do we have utility functions that we can use here?

animesh1987:
Changes look great @bc-peng .

Something for later PRs dealing with changes like this, maybe we can setup a context like `ExperimentalFeaturesContext` and wrap `<CheckoutApp />` in the provider, so we can avoid prop drilling from parent to child components. And read value directly from the context for such features.

animesh1987 (approved):

packages/core/src/app/address/AddressForm.scss
```diff
@@ -1,5 +1,10 @@
 @import "../ui/Base";
 
+.checkout-address:has(.floating-form-field) {
+    margin-left: -0.23rem;
```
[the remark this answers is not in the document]
bc-peng: We want to set the spacing between fields to `8px`, which does not exist in our current design system. (`spacing("single")` is `13px`)

In the last commit, I used a variable instead and moved all hard-coded values into the `_settings.scss` file.

packages/core/src/app/address/AddressForm.scss
```diff
@@ -20,6 +25,10 @@
         }
     }
 
+    .dynamic-form-field.floating-form-field {
+        padding: 0 4px;
```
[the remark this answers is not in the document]
bc-peng: It has been replaced with `padding: 0 0.5 * $floating-form-field-spacing;`.

packages/core/src/app/common/utility/isFloatingLabelEnabled.ts
bc-peng: Sure, using `isFloatingLabelEnabled(checkoutSettings: CheckoutSettings): boolean` now.

packages/core/src/app/customer/__snapshots__/GuestForm.spec.tsx.snap
bc-peng: It was removed because I used a differnt way to create the label for `<EmailField/>` compoent.
All similar changes have been reverted.
All snapshots have been checked, no more removal like this.

packages/core/src/app/ui/form/DynamicInput.tsx
bc-peng: I didn't know we have a large collection of `Icon` components. 😅

This `svg` has been replaced by the existing `IconChevronDown`.

packages/core/src/scss/components/foundation/forms/_forms.scss
bc-peng: I removed this line after realising it was unnecessary to set `height` here.
Also, all hard-coded values have been moved to the `_settings.scss` file.
These hard-coded values are new design styles for floating labels.

davidchin (approved):
Nice one @bc-peng, really nice work! 🚢

PR #1052, opened 2022-10-10
fix(shipping): CHECKOUT-5034 Update autocomplete address set for NZ address

animesh1987:
## What?
Set `sublocality` as street 2 for NZ based addresses. 

## Why?
Set `sublocality` as street 2 for NZ based addresses, since NZ has a concept of a sublocality which is not captured in our shipping form as part of google autocomplete.

## Testing / Proof
- Tests
- Screenshot

Screen Shot 2022-10-12 at 5 29 17 pm

@bigcommerce/checkout


bc-peng (approved):
Nice one 👏

PR #1055, opened 2022-10-11
chore(common): CHECKOUT-6985 Move form ui components

animesh1987:
## What?
Move form UI components to separate package.

## Why?
As part of decomposing checkout JS, this is part 2 of moving form UI components.

Components in this PR are:
- DynamicFormField
- Form

Notes:
Have ignored coverage from three files
- Form
- MultiCheckboxControl
- CheckboxGroupFormField

Will add a JIRA to remove these.

## Testing / Proof
- Circle

@bigcommerce/checkout


packages/ui/src/form/FormField/FormField.tsx
```diff
@@ -0,0 +1,65 @@
...
+    additionalClassName,
+    labelContent,
+    label,
+    onChange,
+    footer,
+    input,
+    name,
+    id,
+}) => {
+    const renderField = useCallback(
```
bc-peng, line 30: I can tell we are missing floating label changes here.
Are we going to add this into the `index.ts` so that we can import `FormField` from here?

packages/ui/src/form/DynamicFormField/DynamicInput.tsx
```diff
@@ -0,0 +1,172 @@
...
+                target: {
+                    name,
+                    value: dateValue,
+                },
+            }),
+        [onChange, name],
+    );
+
+    switch (fieldType) {
+        case DynamicFormFieldType.DROPDOWM:
```
bc-peng, line 50: Same here. Should be a lot of `useFloatingLabel` props in this file.

animesh1987: Possibly, I only plan to use these for payment forms as of now. And since they don't have floating labels then should be ok.

animesh1987:
The build is failing here since we have the FormContext in ui package, and FormProvider in payment-integration package. Which causes a circular dependency 😮‍💨  So I am thinking to move FormProvider back in ui package? 

What do you think @davidchin @bc-peng ?

packages/locale/src/index.ts
```diff
@@ -1 +1,2 @@
 export { default as TranslatedString } from './TranslatedString/TranslatedString';
```
davidchin, line 1: Could we have an index file for each folder?

packages/locale/src/index.ts
```diff
@@ -1 +1,2 @@
...
+export { LocaleContext, LocaleContextType, useLocale } from './context';
```
davidchin, line 2: `contexts` if you intend for this folder to contain all the context components. Otherwise, name it after the component itself, e.g.: `LocaleContext`

animesh1987: I think that's the idea I was going for, to have a contexts folder. I will update that accordingly 😄 . Thanks

bc-peng (approved):
LGTM, nice work.

PR #1057, opened 2022-10-13
fix(payment): INT-6720 prevent calls on undefined error body

orecio-bc:
## What? [INT-6720](https://bigcommercecloud.atlassian.net/browse/INT-6720) [PROJECT-4622](https://bigcommercecloud.atlassian.net/browse/PROJECT-4622)
Prevent calling properties on undefined `error.body` when handling closing modals

## Why?
There are some errors (like `PaymentMethodFailedError` which is thrown by Cardinal) that include a `type` but not a `body`, which breaks the close modal functionality when accessing the property type of undefined.

## Testing / Proof
Without changes:
https://drive.google.com/file/d/1D2XU594smtrOOG6rtp-zVeOP5QM1UEG1/view?usp=sharing

After changes:
https://drive.google.com/file/d/1zgatOvL6gSxGSLLVT11IDLTewdK9N9Tf/view?usp=sharing

### DIFF
--- a/packages/core/src/app/payment/Payment.tsx
+++ b/packages/core/src/app/payment/Payment.tsx
@@ -370,7 +370,7 @@ class Payment extends Component<
                 window.location.reload();
             }
 
-            if (isErrorWithType(error)) {
+            if (isErrorWithType(error) && error.body) {
                 const { body, headers, status } = error;
 
                 if (body.type === 'provider_error' && headers.location) {
--- a/packages/core/src/app/payment/Payment.spec.tsx
+++ b/packages/core/src/app/payment/Payment.spec.tsx
@@ -662,6 +662,25 @@ describe('Payment', () => {
         expect(container.find(PaymentForm).prop('didExceedSpamLimit')).toBeTruthy();
     });
 
+    it('clears error when error has no body', async () => {
+        jest.spyOn(checkoutService, 'loadCheckout').mockResolvedValue(checkoutState);
+        jest.spyOn(checkoutService, 'clearError').mockResolvedValue(checkoutState);
+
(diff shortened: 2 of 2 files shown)


AlanMunozN (approved):

bc-peng (approved):

packages/core/src/app/payment/Payment.tsx
```diff
@@ -370,7 +370,7 @@ class Payment extends Component<
                 window.location.reload();
             }
 
-            if (isErrorWithType(error)) {
+            if (isErrorWithType(error) && error.body) {
```
bc-peng, line 373: On second thought, a unit test to cover this change would be helpful.
Could you please add one?

JelopezBC:
💚 Tested!
PR has been validated as a solution.
Under the following scenarios:

- Cancel a 3DS transaction.
- Complete purchase with 3DS.
- Vaulting a 3DS card and place an order.
- Place an order with vaulted 3DS Card.
- etc.
Please review the evidence.
https://drive.google.com/file/d/12ciAocRyLoyIUA6Ukk9IPFVh-8hx9E61/view?usp=sharing, https://drive.google.com/file/d/14ANbyZplYPLjaQcRpP3Pdnrz2Py9xvAT/view?usp=sharing, https://drive.google.com/file/d/1Ta9V21cGgq2x7FiNHUiFHtNk-T1hlixg/view?usp=sharing, https://drive.google.com/file/d/1d7EapFhAC15tKUrKKe69aQLduZNFcFRd/view?usp=sharing

BC-PabloRivera:
♻️ @bc-peng I added a test for this case

PR #1069, opened 2022-10-18
test(payment): INT-6517 Increase BluesnapV2 payment method coverage

FranciscoOnt:
## What? [INT-6517](https://bigcommercecloud.atlassian.net/browse/INT-6517)
Add modal and page content test case for BluesnapV2 payment method.

Also in the payment method, add listener first then append to page.

## Why?
Test missing case and increase the test coverage for the payment method.

If listener is added after appending content, 'load' would have been already fired and method can't be tested properly.

## Testing / Proof
Screen Shot 2022-10-18 at 2 57 59 PM


mauricio-sg (approved):

bc-peng (approved):
LGTM, thanks.

PR #1071, opened 2022-10-19
feat(checkout): CHECKOUT-6835 Address autocomplete for UK

bc-peng:
## What?
Allow checkout to not automatically fill in the `postcode` and `state/province` fields for UK shipping addresses.

## Why? ([CHECKOUT-6835](https://bigcommercecloud.atlassian.net/browse/CHECKOUT-6835))
We have discovered that Google does not return UK postcodes reliably.

Furthermore, UK addresses do not have “State/Province”. 

## Testing / Proof
Added two unit tests.

### DIFF
--- a/packages/core/src/app/address/googleAutocomplete/AddressSelectorUk.ts
+++ b/packages/core/src/app/address/googleAutocomplete/AddressSelectorUk.ts
@@ -2,10 +2,14 @@ import AddressSelector from './AddressSelector';
 
 export default class AddressSelectorUK extends AddressSelector {
     getState(): string {
-        return this._get('administrative_area_level_2', 'long_name');
+        return '';
     }
 
     getStreet2(): string {
         return this._get('locality', 'long_name');
+    }
+
+    getPostCode(): string {
+        return '';
     }
 }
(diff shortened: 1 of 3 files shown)


packages/core/src/app/address/googleAutocomplete/AddressSelectorUk.ts
```diff
@@ -2,10 +2,14 @@ import AddressSelector from './AddressSelector';
...
 export default class AddressSelectorUK extends AddressSelector {
     getState(): string {
-        return this._get('administrative_area_level_2', 'long_name');
+        return '';
     }
 
     getStreet2(): string {
         return this._get('locality', 'long_name');
     }
+
```
animesh1987, line 11: Should we implement methods for other fields as well?

bc-peng: I understand no product requirements for other fields right now, so it should be good to use the parent class' methods.

packages/core/src/app/address/googleAutocomplete/AddressSelectorUk.ts
```diff
@@ -2,10 +2,14 @@ import AddressSelector from './AddressSelector';
 
 export default class AddressSelectorUK extends AddressSelector {
     getState(): string {
-        return this._get('administrative_area_level_2', 'long_name');
+        return '';
```
animesh1987, line 5: Do we need to change this?

bc-peng: Yes, because the ticket wants no more auto fill for UK's `state/province`.

animesh1987 (approved):

PR #1073, opened 2022-10-24
fix(checkout): CHECKOUT-0000 Fix shipping component memory leak

bc-peng:
## What?
Fix `<Shipping />` component memory leak.

## Why?
The `finally` block inside `componentDidMount` will always run even after `<Shipping />` is unmounted.

Therefore, before running setState, we need to check the mount status.

## Testing / Proof
Before:
Screen Shot 2022-10-24 at 12 03 07 pm

After:
Screen Shot 2022-10-24 at 12 18 28 pm

@bigcommerce/checkout

### DIFF
--- a/packages/core/src/app/checkout/CheckoutStepStatus.ts
+++ b/packages/core/src/app/checkout/CheckoutStepStatus.ts
@@ -1,6 +1,7 @@
 import CheckoutStepType from './CheckoutStepType';
 
 export default interface CheckoutStepStatus {
+    activeStepType?: CheckoutStepType;
     isActive: boolean;
     isBusy: boolean;
     isComplete: boolean;
--- a/packages/core/src/app/shipping/Shipping.tsx
+++ b/packages/core/src/app/shipping/Shipping.tsx
@@ -103,6 +103,7 @@ class Shipping extends Component<ShippingProps & WithCheckoutShippingProps, Ship
             loadPaymentMethods,
             onReady = noop,
             onUnhandledError = noop,
+            step,
         } = this.props;
 
         try {
@@ -113,7 +114,9 @@ class Shipping extends Component<ShippingProps & WithCheckoutShippingProps, Ship
         } catch (error) {
             onUnhandledError(error);
         } finally {
-            this.setState({ isInitializing: false });
+            if (step.activeStepType === step.type) {
+                this.setState({ isInitializing: false });
+            }
         }
     }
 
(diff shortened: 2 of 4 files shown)


packages/core/src/app/shipping/Shipping.tsx
```diff
@@ -113,10 +117,16 @@ class Shipping extends Component<ShippingProps & WithCheckoutShippingProps, Ship
         } catch (error) {
             onUnhandledError(error);
         } finally {
-            this.setState({ isInitializing: false });
+            if (this.isShippingMounted) {
```
animesh1987, line 120: Do we need to add a test for this?

bc-peng: I tried...but I didn't figure out how to access `isShippingMounted` in our context? 😂 Also, I didn't find we were testing `setState` before.

packages/core/src/app/shipping/Shipping.tsx
```diff
@@ -113,10 +117,16 @@ class Shipping extends Component<ShippingProps & WithCheckoutShippingProps, Ship
...
+                this.setState({ isInitializing: false });
```
davidchin, line 121: 🍹 Instead of checking the mounted state, could we just cancel the XHR requests if the component is unmounted? e.g.:

```ts
import { createTimeout } from '@bigcommerce/checkout-sdk'; 

const timeout = createTimeout();

loadShippingOptions({ timeout });
```

And then in unmount function
```ts
timeout.complete();
```

And in the `catch` block, check if `error.status === 0` and not `setState` if the request is aborted.

bc-peng: @davidchin When I tried to implement this, I found out that all three requests could complete before invoking `componentWillUnmount`, so checking the request error status might not work sometimes.

I ended up with another strategy to enhance the `setState` condition. That was to only `setState` if the shipping step was the active step.

packages/core/src/app/checkout/CheckoutStepStatus.ts
```diff
@@ -1,6 +1,7 @@
 import CheckoutStepType from './CheckoutStepType';
 
 export default interface CheckoutStepStatus {
+    activeStepType?: CheckoutStepType;
     isActive: boolean;
```
bc-peng, line 5: A default step may have `isActive` as true, so we need to pass the `activeStepType` to determine the current active step.

packages/core/src/app/shipping/Shipping.tsx
bc-peng: Testing video for the issue. Also tested the whole checkout process, working as expected.

https://user-images.githubusercontent.com/88361607/197689589-264c0609-b54c-4fd6-b85d-a7c2f2857d55.mov

davidchin (approved):

animesh1987 (approved):

PR #1074, opened 2022-10-24
feat(checkout): CHECKOUT-0000 Merge prettier auto fixes

bc-peng:
## What?
Automatic code formatting improvement by `prettier`.

## Why?
These auto fixes constantly show up when we run `npm run lint` because we never commit them into the repo.

## Testing / Proof
CI.

@bigcommerce/checkout

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "release:alpha": "npm run lint && npm run test -- --coverage && npm run build -- --env prerelease=prerelease && npm run release:version",
     "release:version": "git add dist && standard-version -a",
     "generate": "npx nx run core:generate",
-    "lint": "npx nx run-many --target=lint --all --fix",
+    "lint": "npx nx run-many --target=lint --all",
     "test": "npx nx run-many --target=test --all",
     "test:watch": "npx nx run-many --target=test --all --watch"
   },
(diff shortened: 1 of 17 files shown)


davidchin (approved):
LGTM. Separate question, why aren't these raised as an error in the first place?

bc-peng:
> LGTM. Separate question, why aren't these raised as an error in the first place?

@davidchin I think because they are all auto-fixable, and we run `npm run lint` with `--fix`.

davidchin:
@bc-peng we should put in something to fail the build if the autofix changes are not committed. This is to prevent similar things from happening again in the future.

animesh1987:
@bc-peng correct me if I am wrong, these are fixes that are suggested but we haven't actioned on till now. Since most of the rules are turned off in core repo it could be a reason why the build doesn't fail.

bc-peng:
> @bc-peng we should put in something to fail the build if the autofix changes are not committed. This is to prevent similar things from happening again in the future.

@davidchin as we will commit all auto fixes in this PR, maybe we can remove `--fix` from now on.
https://github.com/bigcommerce/checkout-js/blob/d8a6c79870877788bf21ca6ef6bc9e25a56c7df3/package.json#L23

bc-peng:
> @bc-peng correct me if I am wrong, these are fixes that are suggested but we haven't actioned on till now. Since most of the rules are turned off in core repo it could be a reason why the build doesn't fail.

@animesh1987 Some of them are errors, some are warnings.
If we run `npx nx run-many --target=lint --all` without `--fix` on master branch now, the command will fail.

PR #1075, opened 2022-10-25
chore(common): CHECKOUT-6986 Move components to ui package

animesh1987:
## What?
As above

## Why?
As ongoing effort we can move more components across.

## Testing / Proof
- Circle

@bigcommerce/checkout


bc-peng (approved):
LGTM.
Just one question, are we going to remove those components in `core` later?

animesh1987:
> Just one question, are we going to remove those components in core later?

Yes that's the idea, we will slowly remove all the references and replace with the package implementation.

PR #1079, opened 2022-10-26
fix(checkout): CHECKOUT-7048 Allow user to clear cart changed error

huntario:
## What?
...
Check to see if body is defined within the handleCloseModal method 

## Why?
...
When it it undefined the user cannot clear the error to complete checkout - https://bigcommercecloud.atlassian.net/browse/CHECKOUT-7048 

## Testing / Proof
...
Before:
0:25 error cannot be cleared - TypeError: Cannot read properties of undefined (reading 'type')

https://user-images.githubusercontent.com/5630999/198082107-7c2234dd-e92a-4806-befa-a87882be4fc4.mp4

After:
0:24 error can be cleared to complete checkout

https://user-images.githubusercontent.com/5630999/198083790-1dfa816b-3af5-4bda-b717-0ccf54b5431e.mp4

@bigcommerce/checkout

### DIFF
--- a/packages/core/src/app/payment/Payment.tsx
+++ b/packages/core/src/app/payment/Payment.tsx
@@ -373,7 +373,7 @@ class Payment extends Component<
             if (isErrorWithType(error)) {
                 const { body, headers, status } = error;
 
-                if (body.type === 'provider_error' && headers.location) {
+                if (body?.type === 'provider_error' && headers.location) {
                     window.top?.location.assign(headers.location);
                 }
 
@@ -384,8 +384,8 @@ class Payment extends Component<
                 // HTML page instead of JSON response when there is a 429 error.
                 if (
                     status === 429 ||
-                    body.type === 'spam_protection_expired' ||
-                    body.type === 'spam_protection_failed'
+                    body?.type === 'spam_protection_expired' ||
+                    body?.type === 'spam_protection_failed'
                 ) {
                     this.setState({ didExceedSpamLimit: true });
 


packages/core/src/app/payment/Payment.tsx
```diff
@@ -372,24 +372,25 @@ class Payment extends Component<
...
-                // HTML page instead of JSON response when there is a 429 error.
-                if (
-                    status === 429 ||
-                    body.type === 'spam_protection_expired' ||
-                    body.type === 'spam_protection_failed'
-                ) {
-                    this.setState({ didExceedSpamLimit: true });
-
-                    await loadCheckout();
+                if (body || status === 429){
```
davidchin, line 375: If `body` is `null` and `status` is 429, the line below `if (body.type === 'provider_error' && headers.location) {` will still break?

I think the quickest fix is to just put a question mark after `body`, like `body?.type` within this method.

bc-peng, line 375: As a recent bugfix involved checking `error.body` on line 373, we might not need to recheck on line 375.
https://github.com/bigcommerce/checkout-js/pull/1057/files

huntario: Agh good point, thanks for the advice. Looks like it was fixed in a separate PR but went ahead and made this change for practice.

huntario: Nice catch. Thanks for taking a look at this. I've tested this out again with the latest checkout-js and the error can be cleared with that change.  I'll keep an eye out for the bump in BCApp and close [CHECKOUT-7048](https://bigcommercecloud.atlassian.net/browse/CHECKOUT-7048) accordingly.

huntario:
This was also fixed by https://github.com/bigcommerce/checkout-js/pull/1057/files so we'll just go with that one.

davidchin (approved):

packages/core/src/app/payment/Payment.tsx
animesh1987: Thanks @bc-peng and @huntario

PR #1082, opened 2022-10-31
fix(checkout): CHECKOUT-6840 Fix payment step counter in Safari

bc-peng:
## What?
Fix the problem where Safari always displays 1 as the payment step number instead of 4.

## Why?
Safari's CSS counter function differs from Chrome's.

### After
iOS Safari

Also tested on iOS Chrome, Android Chrome, and Firefox, all of which behaved as expected.

### DIFF
--- a/packages/core/src/scss/components/checkout/checkoutSteps/_checkoutSteps.scss
+++ b/packages/core/src/scss/components/checkout/checkoutSteps/_checkoutSteps.scss
@@ -19,6 +19,7 @@
 
 .checkout-step {
     border-bottom: container("border");
+    counter-increment: $checkoutStep-counter;
     padding: spacing("single") 0;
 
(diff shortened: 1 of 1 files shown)


packages/core/src/scss/components/checkout/checkoutSteps/_checkoutSteps.scss
```diff
@@ -19,6 +19,7 @@
 
 .checkout-step {
     border-bottom: container("border");
+    counter-increment: $checkoutStep-counter;
```
bc-peng, line 22: It turns out that increasing the counter with `.checkout-step` creates a more consistent cross-browser counter result.

davidchin (approved):

animesh1987 (approved):

PR #1084, opened 2022-11-02
fix(checkout): CHECKOUT-5742 shouldSaveAddress is not saved

bc-peng:
## What?
Fix the bug that `shouldSaveAddress` was not saved when a shopper made an explicit change to it.

## Why? ([CHECKOUT-5742](https://bigcommercecloud.atlassian.net/browse/CHECKOUT-5742))
Prevously, we only updated `shouldSaveAddress` value on the server if an address change occurred. 

Now, whenever the value of `shouldSaveAddress` is changed, it is sent to the server.

## Testing / Proof
Manual testing:

https://user-images.githubusercontent.com/88361607/199401949-035be053-e0b0-453d-b50c-19b8311c08df.mov

### DIFF
--- a/packages/core/src/app/shipping/Shipping.tsx
+++ b/packages/core/src/app/shipping/Shipping.tsx
@@ -232,7 +232,7 @@ class Shipping extends Component<ShippingProps & WithCheckoutShippingProps, Ship
         const promises: Array<Promise<CheckoutSelectors>> = [];
         const hasRemoteBilling = this.hasRemoteBilling(methodId);
 
-        if (!isEqualAddress(updatedShippingAddress, shippingAddress)) {
+        if (!isEqualAddress(updatedShippingAddress, shippingAddress) || shippingAddress?.shouldSaveAddress !== updatedShippingAddress?.shouldSaveAddress) {
             promises.push(updateShippingAddress(updatedShippingAddress || {}));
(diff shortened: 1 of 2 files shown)


packages/core/src/app/shipping/Shipping.tsx
```diff
@@ -232,7 +232,7 @@ class Shipping extends Component<ShippingProps & WithCheckoutShippingProps, Ship
         const promises: Array<Promise<CheckoutSelectors>> = [];
         const hasRemoteBilling = this.hasRemoteBilling(methodId);
 
-        if (!isEqualAddress(updatedShippingAddress, shippingAddress)) {
+        if (!isEqualAddress(updatedShippingAddress, shippingAddress) || shippingAddress?.shouldSaveAddress !== updatedShippingAddress?.shouldSaveAddress) {
```
bc-peng, line 235: When we compare two addresses in `isEqualAddress`, we explicitly ignore `shouldSaveAddress`, so I put the check here.
https://github.com/bigcommerce/checkout-js/blob/5f25ba2e45ec698bddd842b639f8a8b0b6cd92fa/packages/core/src/app/address/isEqualAddress.ts#L45

animesh1987 (approved):

PR #1086, opened 2022-11-02
feat(payment): INT-5953-T Force to use Payment Method V1 when Mollie and ApplePay

clopezh:
## What?
Add a validation to force to use PaymentMethod V1 when the gateway is mollie.
This to use the MolliePaymentStrategy.

## Why?
As a momentarily solution to support Apple Pay on Mollie and using the MolliePaymentMethod, and meanwhile an enhancement to support Mollie as Payment Method V2.

## Testing / Proof
image

## Depends on
https://github.com/bigcommerce/checkout-sdk-js/pull/1671

@bigcommerce/checkout @bigcommerce/payments @bigcommerce/apex-integrations

### DIFF
--- a/packages/core/src/app/payment/paymentMethod/PaymentMethodV2.spec.tsx
+++ b/packages/core/src/app/payment/paymentMethod/PaymentMethodV2.spec.tsx
@@ -101,4 +101,28 @@ describe('PaymentMethod', () => {
 
         expect(componentFallback.find(Foo)).toBeDefined();
     });
+
+    it('returns payment method v1 if gateway is mollie', () => {
+        const Foo: FunctionComponent<PaymentMethodProps> = ({ method }) => <div>{method.id}</div>;
+
+        const resolver = (method: PaymentMethod) => {
+            return method.id === 'applepay' ? Foo : undefined;
+        };
+
+        const componentFallback = mount(
+            <ContextProvider>
+                <PaymentMethodComponent
+                    method={{
+                        ...getPaymentMethod(),
+                        gateway: 'mollie',
+                        id: 'applepay',
+                    }}
+                    onUnhandledError={jest.fn()}
+                    resolveComponent={resolver}
+                />
+            </ContextProvider>,
+        );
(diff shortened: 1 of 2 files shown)


packages/core/src/app/payment/paymentMethod/PaymentMethodV2.spec.tsx
```diff
@@ -101,4 +101,28 @@ describe('PaymentMethod', () => {
...
+            return method.id === 'applepay' ? Foo : undefined;
+        };
+
+        const componentFallback = mount(
+            <ContextProvider>
+                <PaymentMethodComponent
+                    method={{
+                        ...getPaymentMethod(),
+                        gateway: 'mollie',
+                        id: 'test',
```
animesh1987, line 118: Should the id be `applepay` just to keep the test aligned with the requirement?

clopezh: No, because if I changed it to `applepay` then the resolver: 
```
 const ResolvedPaymentMethod = resolveComponent({
        id: method.id,
        gateway: method.gateway,
        type: method.type,
    });

will get applepay payment method (from V2), and the validation of: https://github.com/bigcommerce/checkout-js/pull/1086/files#diff-0ad410b99a894433eb62773917d0c31ce5e3878ffdb6f22bd228f1fad9793e0eR73 will be false.

animesh1987: From what I understand we want to return the V1 component if id is ApplePay and gateway is Mollie? So in that case even if the component is resolved we should return the mollie component?

clopezh: well you're right, let me change it

animesh1987 (approved):

animesh1987 (approved):

animesh1987 (approved):

PR #1090, opened 2022-11-08
fix(checkout): CHECKOUT-6672 Maintain subscription checkbox state

bc-peng:
## What?
Fix the bugs:
- A shopper cannot actually opt out marketing newsletter on the checkout page.
- The newsletter option always displays its default state rather than a shopper's actual choice.

The limitation of this bugfix is when a shopper reloads the page, the state will lose.

## Why? [CHECKOUT-6672](https://bigcommercecloud.atlassian.net/browse/CHECKOUT-6672)

The subscription state of a shopper's email address is currently unavailable through our API endpoints. 

There will be a server-side ticket made as a follow-up. (update: [CHECKOUT-7095](https://bigcommercecloud.atlassian.net/browse/CHECKOUT-7095))

## Testing / Proof

Manual testing:

https://user-images.githubusercontent.com/88361607/200476203-8c21e6e7-4d47-41dd-95d0-b2cb2232576d.mov

@bigcommerce/checkout

### DIFF
--- a/packages/core/src/app/checkout/Checkout.tsx
+++ b/packages/core/src/app/checkout/Checkout.tsx
@@ -128,6 +128,7 @@ export interface CheckoutState {
     isRedirecting: boolean;
     hasSelectedShippingOptions: boolean;
     isBuyNowCartEnabled: boolean;
+    isSubscribed: boolean;
 }
 
 export interface WithCheckoutProps {
@@ -166,6 +167,7 @@ class Checkout extends Component<
         isMultiShippingMode: false,
         hasSelectedShippingOptions: false,
         isBuyNowCartEnabled: false,
+        isSubscribed: false,
     };
 
     private embeddedMessenger?: EmbeddedCheckoutMessenger;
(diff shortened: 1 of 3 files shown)


packages/core/src/app/checkout/Checkout.tsx
```diff
@@ -128,6 +128,7 @@ export interface CheckoutState {
     isRedirecting: boolean;
     hasSelectedShippingOptions: boolean;
     isBuyNowCartEnabled: boolean;
+    isSubscribed: boolean;
```
bc-peng, line 131: When the server-side implementation is done, we can get rid of everything in `checkout.tsx` from this PR.
Show the subscription choice of a shopper based on something like this:
```
const { isSubscribed } = getBillingAddress();
```

packages/core/src/app/customer/Customer.tsx
```diff
@@ -58,6 +60,7 @@ export interface WithCheckoutCustomerProps {
     isCreatingAccount: boolean;
     isExecutingPaymentMethodCheckout: boolean;
     isGuestEnabled: boolean;
+    isHavingBillingId: boolean;
```
animesh1987, line 63:
```suggestion
    hasBillingId: boolean;
```

animesh1987 (approved):

PR #1092, opened 2022-11-09
feat(checkout): CHECKOUT-7036 Filter out convertcart-related issues on Sentry.

bc-peng:
## What?
Filter out convertcart-related issues on Sentry.

## Why? [CHECKOUT-7036](https://bigcommercecloud.atlassian.net/browse/CHECKOUT-7036)
Reduce the amount of noisy errors we receive, such as [3505225940](https://sentry.io/organizations/bigcommerce/issues/3505225940/?project=1542560).

## Testing / Proof
- Added one unit test.
- CI

### DIFF
--- a/packages/core/src/app/common/error/SentryErrorLogger.ts
+++ b/packages/core/src/app/common/error/SentryErrorLogger.ts
@@ -45,7 +45,7 @@ export default class SentryErrorLogger implements ErrorLogger {
         init({
             sampleRate: SAMPLE_RATE,
             beforeSend: this.handleBeforeSend,
-            denyUrls: [...(config.denyUrls || []), 'polyfill~checkout', 'sentry~checkout'],
+            denyUrls: [...(config.denyUrls || []), 'polyfill~checkout', 'sentry~checkout', 'convertcart'],
             integrations: [
                 new Integrations.GlobalHandlers({
                     onerror: false,
--- a/packages/core/src/app/common/error/SentryErrorLogger.ts
+++ b/packages/core/src/app/common/error/SentryErrorLogger.ts
@@ -135,6 +135,12 @@ export default class SentryErrorLogger implements ErrorLogger {
     }
 
     private handleBeforeSend: (event: Event, hint?: EventHint) => Event | null = (event, hint) => {
+        if (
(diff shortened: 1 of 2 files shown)


packages/core/src/app/common/error/SentryErrorLogger.ts
```diff
@@ -135,6 +135,12 @@ export default class SentryErrorLogger implements ErrorLogger {
     }
 
     private handleBeforeSend: (event: Event, hint?: EventHint) => Event | null = (event, hint) => {
+        if (
+            event.breadcrumbs?.filter((breadcrumb) => breadcrumb.data?.url.includes('convertcart'))
```
animesh1987, line 139: Do you for see us having a list of works like `convertcart` that we need to ignore in future?

bc-peng: For other issues, we might have to use other signatures to identify them.
Right now, I think we just aim to remove `convertcart`.

snaderiBC, line 139: Is it possible to use and take advantage of https://docs.sentry.io/platforms/javascript/guides/capacitor/configuration/options/#deny-urls

bc-peng: Yesterday, the same Sentry issue's URL was https://www.extremekartz.com/checkout, today it is https://mcneelamusic.com/checkout/order-confirmation.
Therefore, I think it is better to identify `convertcart` in breadcrumbs.

animesh1987: Found this in github issues https://github.com/getsentry/sentry-javascript/issues/3147, seems similar. It also links a PR for denyUrls.

bc-peng: I have added `convertcart` to `denyUrls`.
Maybe we can keep the filter in `beforeSend` as it is the simlar solution mentioned in https://github.com/getsentry/sentry-javascript/issues/3147 .

animesh1987 (approved):

PR #883, opened 2022-05-27
feat(checkout): CHECKOUT-6723 Add Norwegian Language

bc-peng:
## What?
Add Norwegian Language.

## Why?
BC’s expansion into international markets has generated the need to support further languages. The new translations are included in the attached zip file.

Language files source: [CHECKOUT-6723](https://bigcommercecloud.atlassian.net/browse/CHECKOUT-6723).

## Testing / Proof
Screen Shot 2022-05-27 at 10 21 48 am

PR #1093, opened 2022-11-09
Update translations

bc-svc-local:
ping @bigcommerce/localization


animesh1987 (approved):

funivan:
@animesh1987 can you check this PR and merge it? Thanks.

packages/core/src/app/locale/translations/da.json
```diff
@@ -43,14 +44,14 @@
         },
         "billing": {
             "billing_address_heading": "Faktureringsadresse",
-            "billing_heading": "Fakturering",
+            "billing_heading": "fakturering",
```
animesh1987, line 47: @bc-peng I remember you had some questions around this? Is that supposed to be capitalized?

animesh1987:
Thanks for the reminder @funivan , we need to rebase this PR. And I remember @bc-peng pointed out some issue earlier.

packages/core/src/app/locale/translations/da.json
bc-peng: @funivan Could you please check is this better to be capitalized as you can see currently we are using `Fakturering` in production? This issue was first pointed out [here](https://github.com/bigcommerce/checkout-js/pull/883#discussion_r883218922).

funivan: 👋 I can change manually this string once in our Single Source of Truth. But unfortunately, there is no way to control if translators write words in a specific case.
__Changing manually:-_
pros - fast, time-consuming if we need to do this multiple times.
cons - we can have the same error tomorrow and it will be hard to spot it. 


Can we have a function that capitalizes the first letter and use it?

bc-peng: I understand there are many ways to force capitalization on FE. However, I believe it is better to use the translation json files as the single source of truth to display any string on the checkout page.

bc-peng: As an alternative mechanism, we can use a unit test to ensure that strings like this one are capitalised. 
@funivan What do you think?

funivan: > However, I believe it is better to use the translation json files as the single source of truth to display any string on the checkout page

are there some patterns in what words we should capitalize first letter?

> As an alternative mechanism, we can use a unit test to ensure that strings like this one are capitalised.

Definitely, it will work to catch some wrong word formats. But what we will do next:
1. Change phrases in localization-dictionaries <- this is the source of truth for all translations that we have in the platform. 
2. Recreate the PR (this should be manually triggered)
This can cost time and effort every time we add a new language & phrases.

funivan:
@animesh1987 We create PRs automatically and there is no easy way to rebase changes automatically. 
We can close this one and create a new one.


The source of truth for these translations is https://github.com/bigcommerce/localization-dictionaries/tree/main/dictionaries/checkout-js In case if you need to change translations it is better to change phrases in this repo.

bc-peng:
I have created a PR at https://github.com/bigcommerce/localization-dictionaries/pull/229 to capitalise `fakturering` for checkout related area. @funivan Could you please approve it?
cc @animesh1987

funivan:
@animesh1987 can u close this PR in favor https://github.com/bigcommerce/checkout-js/pull/1210

bc-svc-local:
new one https://github.com/bigcommerce/checkout-js/pull/1215

PR #1087, opened 2022-11-04
chore(common): CHECKOUT-6440 Add hosted widget component package

animesh1987:
## What?
As above

## Why?
In order to support payment methods monorepo packages that compose with `<HostedWidgetPaymentComponent />` we need to move it as a shared package.

Working on a separate PR to add unit tests.

## Testing / Proof
- circle

@bigcommerce/checkout

PR #1101, opened 2022-11-11
chore(common): CHECKOUT-6440 Add tests hosted widget component package

animesh1987:
## What?
As above

## Why?
Follow up from https://github.com/bigcommerce/checkout-js/pull/1087. This PR adds the tests for the `HostedWidgetPaymentComponent` package.

## Testing / Proof
- Circle

@bigcommerce/checkout


jest-setup.ts
```diff
@@ -3,32 +3,45 @@ import * as Adapter from 'enzyme-adapter-react-16';
...
+    'userAgent',
+    ((value) => ({
+        get() {
+            return value;
+        },
+        set(v) {
+            value = v;
+        },
+    }))(window.navigator.userAgent),
+);
```
bc-peng, line 44: This file appears to only have formatting changes.
Could you please confirm that it is all done by code rather than manual editing so that it does not change in the future?

packages/locale/src/contexts/index.ts
```diff
@@ -1 +1,2 @@
 export { default as LocaleContext, LocaleContextType, useLocale } from './LocaleContext';
```
bc-peng, line 1: I'm just curious why we explicitly list all properties in the export instead of using `*`?
Also ,the `default` usage seems redundant since we never use default import...

jest-setup.ts
animesh1987: This seems to be done by auto saving.

packages/locale/src/contexts/index.ts
animesh1987: We can do that too, just feel explicit is better. The default will be used once we need to wrap components in the context. 

So we can do something like `<LocaContext.Provider value={value}>{/* children */}</LocaContext.Provider>` instead of the `withTranslations` HOC pattern.

bc-peng: Perhaps I didn't express myself correctly. 
If we don't set `default` export in`./LocaleContext`, we can get rid of `default`.
```
export { LocaleContext, LocaleContextType, useLocale } from './LocaleContext';
```
It should be used in the same way with or without `default`.

packages/test-utils/src/checkout.mock.ts
```diff
@@ -0,0 +1,78 @@
...
+        subtotal: 190,
+        grandTotal: 190,
+        giftWrappingCostTotal: 0,
+        outstandingBalance: 190,
+        giftCertificates: [],
+        balanceDue: 0,
+        createdTime: '2018-03-06T04:41:49+00:00',
+        updatedTime: '2018-03-07T03:44:51+00:00',
+        promotions: [],
+    };
```
bc-peng, line 40: We added a mandatory field 8 days ago.
https://github.com/bigcommerce/checkout-sdk-js/blob/ee42eb2573ef1e8ab9ee34f8304f5d76b5c8585f/packages/core/src/checkout/checkout.ts#L43
```suggestion
        channelId: 123456,
    };
```

packages/test-utils/tsconfig.lib.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "outDir": "../../dist/out-tsc",
```
davidchin, line 4: Do you need `outDir`?

packages/test-utils/src/checkout.mock.ts
animesh1987: Yes, will update the mocks.

packages/test-utils/tsconfig.lib.json
animesh1987: Actually don't think so, this is just the generic template by Nx.

davidchin (approved):

PR #1102, opened 2022-11-13
Revert "feat(checkout): CHECKOUT-7036 Filter out convertcart-related issues on Sentry."

TomA-R:
Reverts bigcommerce/checkout-js#1092

This seems to be causing:

```
TypeError: undefined is not an object (evaluating 't.url.includes')
  at ? (./packages/core/src/app/common/error/SentryErrorLogger.ts:139:72)
  at filter([native code])
  at ? (./packages/core/src/app/common/error/SentryErrorLogger.ts:139:32)
```

ping @bigcommerce/checkout


animesh1987 (approved):
Thanks @TomA-R 🙇‍♂️

PR #1103, opened 2022-11-13
feat(checkout): CHECKOUT-7036 Filter out convertcart-related issues

bc-peng:
## What?
The previous PR did help reduce the number of the convertcart-related issues, it also introduced a new one because it failed to realise that the `url` property is optional.

This PR expands the optional chaining to include `url`.

The updated unit test captures the case where `data` and `url` are optional.

## Why?
Reduce the amount of noisy errors we receive, such as [3505225940](https://sentry.io/organizations/bigcommerce/issues/3505225940/?project=1542560).
## Testing / Proof
- CI.

@bigcommerce/checkout

### DIFF
--- a/packages/core/src/app/common/error/SentryErrorLogger.ts
+++ b/packages/core/src/app/common/error/SentryErrorLogger.ts
@@ -45,7 +45,7 @@ export default class SentryErrorLogger implements ErrorLogger {
         init({
             sampleRate: SAMPLE_RATE,
             beforeSend: this.handleBeforeSend,
-            denyUrls: [...(config.denyUrls || []), 'polyfill~checkout', 'sentry~checkout'],
+            denyUrls: [...(config.denyUrls || []), 'polyfill~checkout', 'sentry~checkout', 'convertcart'],
             integrations: [
                 new Integrations.GlobalHandlers({
                     onerror: false,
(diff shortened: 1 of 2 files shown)


animesh1987 (approved):

packages/core/src/app/common/error/SentryErrorLogger.ts
```diff
@@ -135,6 +135,12 @@ export default class SentryErrorLogger implements ErrorLogger {
     }
 
     private handleBeforeSend: (event: Event, hint?: EventHint) => Event | null = (event, hint) => {
+        if (
+            event.breadcrumbs?.filter((breadcrumb) => breadcrumb.data?.url?.includes('convertcart'))
```
davidchin, line 139: I don't think we should filter out errors by hardcoding third party URLs like this, because it's impossible to filter out all the third party URLs defined by custom checkouts.

I think we have the following filter function that only tracks exceptions if they are thrown by the code that we own:
https://github.com/bigcommerce/checkout-js/blob/68a899bc56c59b298681d9872c7ff462f107af4b/packages/core/src/app/common/error/SentryErrorLogger.ts#L114-L135

It seems it's not working as expected if the exception only has a single frame in the stacktrace, and the frame is the Sentry logger itself. i.e.:

Screen Shot 2022-11-14 at 9 22 38 am

I think a better fix is to actually update the filter function to also cater for that particular scenario. Thoughts?

davidchin (commented):
Sorry I wasn't able to submit my comment in time 😅

packages/core/src/app/common/error/SentryErrorLogger.ts
bc-peng: Thanks DC for the feedback.
I have created [CHECKOUT-7100](https://bigcommercecloud.atlassian.net/browse/CHECKOUT-7100) to follow up on this.
cc @animesh1987

PR #1112, opened 2022-11-26
style(checkout): STRIPE-192 Add labels as floating

er-contreras:
## What? [STRIPE-192](https://bigcommercecloud.atlassian.net/jira/software/c/projects/STRIPE/boards/622?modal=detail&selectedIssue=STRIPE-192)

Add labels as floating

## Why?
in order to change the style of the label in checkout customer, billing and payment inputs to be inside.

## Testing / Proof

https://user-images.githubusercontent.com/67211919/208209385-97c2ae58-df3d-4c23-a7bc-4c61c7f5633c.mov

### DIFF
--- a/packages/core/src/app/ui/loading/LoadingOverlay.tsx
+++ b/packages/core/src/app/ui/loading/LoadingOverlay.tsx
@@ -21,7 +21,7 @@ const LoadingOverlay: FunctionComponent<LoadingOverlayProps> = ({
                 {unmountContentWhenLoading && isLoading ? null : (
                     <div
                         style={{
-                            display: hideContentWhenLoading && isLoading ? 'none' : undefined,
+                            visibility: hideContentWhenLoading && isLoading ? 'hidden' : undefined,
                         }}
                     >
                         {children}
--- a/packages/core/src/app/customer/StripeGuestForm.tsx
+++ b/packages/core/src/app/customer/StripeGuestForm.tsx
@@ -110,9 +110,9 @@ const StripeGuestForm: FunctionComponent<StripeGuestFormProps & FormikProps<Gues
         if (parentContainer) {
             return getAppliedStyles(parentContainer, properties);
         }
- 
+
             return undefined;
-        
+
     };
 
     const containerId = 'stripe-card-component-field';
@@ -181,7 +181,7 @@ const StripeGuestForm: FunctionComponent<StripeGuestFormProps & FormikProps<Gues
                             </Legend>
                         }
                     >
-                        <div className="customerEmail-container">
+                        <div className="customerEmail-container" style={{paddingTop: '4px'}}>
                             <div className="customerEmail-body">
                                 <div id="stripeupeLink"/>
                                 <br/>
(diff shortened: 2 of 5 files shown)


packages/core/src/scss/components/checkout/customer/_customer.scss
```diff
@@ -41,6 +41,8 @@
     @include breakpoint("small") {
         width: grid-calc(7, $total-columns);
     }
+
+    margin-top: 20px;
```
bc-peng, line 45: image
It looks like the text fields are larger than the normal (non-stripe) form.
Also, CSS changes like this will apply to all interface with or without floating labels. Could you please provide a preview on those non-stripe forms. Thanks.

er-contreras: Hi @bc-peng, thanks for review this PR.  

https://user-images.githubusercontent.com/67211919/205139101-4d7cae49-2be5-40e6-a0c3-ff09092db9f2.mov

BC-AOrtiz: Hey @er-contreras could you add a screend shot or demo whit the css chagnes where we could see the billing field (non-stripe fields)

er-contreras: Sure,

bc-peng: This line and the one in `_forms.css` appear to be breaking current checkout design.
Please refer to the screenshots.
Screen Shot 2022-12-02 at 10 23 17 pm
Screen Shot 2022-12-02 at 10 26 07 pm

cc @animesh1987

bc-peng:
I think the CI check failed because of the code coverage, which should be fixed after rebase.
Is this PR achieving what you intended?

er-contreras:
> I think the CI check failed because of the code coverage, which should be fixed after rebase. Is this PR achieving what you intended?
Hi, @bc-peng 
It is, there were some changes that fix the issue that affect BC front-end.

packages/core/src/app/ui/loading/LoadingOverlay.tsx
```diff
@@ -21,7 +21,7 @@ const LoadingOverlay: FunctionComponent<LoadingOverlayProps> = ({
                 {unmountContentWhenLoading && isLoading ? null : (
                     <div
                         style={{
-                            display: hideContentWhenLoading && isLoading ? 'none' : undefined,
+                            visibility: hideContentWhenLoading && isLoading ? 'hidden' : undefined,
```
bc-peng, line 24: Could you please explain the reason of changing this?
This component is everywhere, I understand the difference it will make, I'm afarid a thorough testing of the whole checkout is required.

er-contreras: By now it is the only way to make work the changes. Somehow `display: none` affect the the Stripe API appearance classes that are in the DOM, and don't pass through some values. Those values appears as 'NaNpx' so while this issue  I don't know where exactly is coming that bug when we set `Labels: floating` in the Stripe appearance. I thought it could be an option but if is a problem it would be better to find another way to make it work properly.

bc-peng: I have tested this change, and found no noticeable visual changes.
Think we can go ahead with this, and please apply changes to `packages/ui/src/loading/LoadingOverlay.tsx`, too.

PR #1116, opened 2022-12-01
fix(payment): STRIPE-269 exclude wallets from the custom customer step status for stripeLink

AlanMunozN:
## What? [STRIPE-269](https://bigcommercecloud.atlassian.net/browse/STRIPE-269)
Exclude wallet from custom stripeLinkFlow

## Why?
Could affect functionality and UX 

## Testing / Proof
BUG:
https://drive.google.com/file/d/1cw6iNp2HL-StziDoiNc3ZaVAHqZgm7K2/view?usp=share_link

FIX:
https://drive.google.com/file/d/1BuaMZEVnCQKhpOlNlN9TuodBNZf8L1cl/view?usp=share_link

### DIFF
--- a/packages/core/src/app/checkout/getCheckoutStepStatuses.ts
+++ b/packages/core/src/app/checkout/getCheckoutStepStatuses.ts
@@ -32,9 +32,15 @@ const getCustomerStepStatus = createSelector(
                 : false;
         const isGuest = !!(customer && customer.isGuest);
         const isComplete = hasEmail || isUsingWallet;
-        const isEditable = isComplete && !isUsingWallet && isGuest
+        const isEditable = isComplete && !isUsingWallet && isGuest;
 
-        if (config?.checkoutSettings.providerWithCustomCheckout === PaymentMethodId.StripeUPE && hasEmail && isGuest) {
+        // StripeLink is a UX that is only available with StripeUpe and will only be displayed for BC guest users,
+        // it uses its own components in the customer and shipping steps, unfortunately in order to preserve the UX
+        // when reloading the checkout page it's necessary to refill the stripe components with the information saved.
+        // In this step, we require that the customer strategy be reloaded the first time.
+        const isUsingStripeLinkAndCheckoutPageIsReloaded = !isUsingWallet &&
+            config?.checkoutSettings.providerWithCustomCheckout === PaymentMethodId.StripeUPE && hasEmail && isGuest;
+        if (isUsingStripeLinkAndCheckoutPageIsReloaded) {
             return {
                 type: CheckoutStepType.Customer,
                 isActive: false,
--- a/packages/core/src/app/checkout/getCheckoutStepStatuses.spec.ts
+++ b/packages/core/src/app/checkout/getCheckoutStepStatuses.spec.ts
@@ -11,9 +11,11 @@ import {
 } from '../address/formField.mock';
 import { getBillingAddress, getEmptyBillingAddress } from '../billing/billingAddresses.mock';
 import { getCart } from '../cart/carts.mock';
+import { getStoreConfig } from '../config/config.mock';
 import { getCustomer, getGuestCustomer } from '../customer/customers.mock';
 import { getOrder } from '../order/orders.mock';
 import { getPaymentMethod } from '../payment/payment-methods.mock';
+import { PaymentMethodId } from '../payment/paymentMethod';
 import { getConsignment } from '../shipping/consignment.mock';
 import { getShippingAddress } from '../shipping/shipping-addresses.mock';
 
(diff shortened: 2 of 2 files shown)


BC-AOrtiz (approved):
LGTM

packages/core/src/app/checkout/getCheckoutStepStatuses.ts
```diff
@@ -34,7 +34,7 @@ const getCustomerStepStatus = createSelector(
         const isComplete = hasEmail || isUsingWallet;
         const isEditable = isComplete && !isUsingWallet && isGuest
 
-        if (config?.checkoutSettings.providerWithCustomCheckout === PaymentMethodId.StripeUPE && hasEmail && isGuest) {
+        if (!isUsingWallet && config?.checkoutSettings.providerWithCustomCheckout === PaymentMethodId.StripeUPE && hasEmail && isGuest) {
```
animesh1987, line 37: Can we please add tests to cover this? Also can we work on this control flow, feels like we are just adding conditions one after the other. 

Have you tested this with other wallet buttons?

BCHMercado:
💚 Testing done !
https://bigcommercecloud.atlassian.net/jira/software/c/projects/STRIPE/boards/622?modal=detail&selectedIssue=STRIPE-269

packages/core/src/app/checkout/getCheckoutStepStatuses.ts
AlanMunozN: Hello @animesh1987, I already added unit testing, also, QA already validated the wallets available for StripeUPE, I added a video of the missing wallet https://drive.google.com/file/d/1N7JxCWImcA960SqjC6eaX1N9zLEmBb8u/view?usp=sharing, and about the approach let me explain a little each validation inside the If validation for stripeLink.

The validation is the following:
`if (!isUsingWallet && config?.checkoutSettings.providerWithCustomCheckout === PaymentMethodId.StripeUPE && hasEmail && isGuest)`

**`config?.checkoutSettings.providerWithCustomCheckout === PaymentMethodId.StripeUPE`**: this validation was suggested by you, it comes from BCAPP and tells us when StripeLink is active.

**`hasEmail`**: this validation was added to make sure that the flow is only used if an email has already been saved, it could be deleted but we prefer to exclude it from the custom flow when it is not necessary.

**`isGuest`**: Stripelink only works with guest so in the case of using the BC customer flow we are not interested in having the custom flow.

and the last one, **`isUsingWallet`** : which is used to exclude wallets from the custom flow, since they don't need to reload stripe components in case of a reload.

please let me know any concern :D

bc-peng: Because this is part of the core logic of checkout-js, `!isUsingWallet && config?.checkoutSettings.providerWithCustomCheckout === PaymentMethodId.StripeUPE && hasEmail && isGuest` is simply too complex for future developers to comprehend.

Could we replace this long line with a function like `isUsingStripeLink()`? (This may not be the best name. please summarise the conditions' purpose.)

animesh1987: Hey Alan thanks for the explanation, it does feel a bit too many conditions still. 

On the condition `config?.checkoutSettings.providerWithCustomCheckout === PaymentMethodId.StripeUPE` what happens when we add another accelerated checkout in future?


On another note in the video above, I can see wallet buttons still active. Thought there was a decision made somewhere to remove wallet buttons if stripe link was active?

AlanMunozN: Hello @bc-peng, as we saw by message, I made a constant instead of a function so to avoid sending many parameters, let me know if It's ok and if not I can change it for a function :D, I also added a comment explaining the reason for the `constant` and the `if` 

In response to your questions @animesh1987,
Apparently, there can only be one accelerated checkout working. 
For example:
**Bolt** and **StripeLink** cannot coexist. I understand that the **Bolt team** is working on these validations, so if we add another accelerated checkout in the future, there will be a standard flow to avoid **StripeLink** as well as **Bolt**.

And in response to your last question, It's partially correct, we remove the wallets for **StripeLink** but as long as the user is authenticated in **Stripe** (In the demo, I am still a guest for both **BC** and **Stripe**)

Let me know any other questions, regards

packages/core/src/app/checkout/getCheckoutStepStatuses.ts
```diff
@@ -34,7 +34,13 @@ const getCustomerStepStatus = createSelector(
         const isComplete = hasEmail || isUsingWallet;
         const isEditable = isComplete && !isUsingWallet && isGuest
 
-        if (config?.checkoutSettings.providerWithCustomCheckout === PaymentMethodId.StripeUPE && hasEmail && isGuest) {
+        // StripeLink is a UX that is only available with StripeUpe and will only be displayed for BC guest users,
+        // it uses its own components in the customer and shipping steps, unfortunately in order to preserve the UX
+        // when reloading the checkout page it's necessary to refill the stripe components with the information saved.
+        // In this step, we require that the customer strategy be reloaded the first time.
+        const isUsingStripeLinkAndCheckoutPageIsReloaded = !isUsingWallet &&
+            config?.checkoutSettings.providerWithCustomCheckout === PaymentMethodId.StripeUPE && hasEmail && isGuest
```
bc-peng, line 42: Could you please add `;` to Line 42 and 35 ?
`checkout-js` linter should catch these automatically. 🤔

AlanMunozN: @bc-peng, I'm so sorry, I think Ruby has affected me hahaha

bc-peng (approved):
LGTM, thanks.

animesh1987 (approved):

PR #1125, opened 2022-12-14
fix(checkout): CHECKOUT-0000 Revert square to use previous hosted payment component

animesh1987:
## What?
As above

## Why?
While migrating square payment method to use the `HostedPaymentComponent`  from separate monorepo package. We realised that the translations provider needs to be built correctly, otherwise the component seems to be missing translation context. 

Once we update the locale package with correct translation setup, we can migrate square to use monorepo package component.

## Testing / Proof
- circle

@bigcommerce/checkout


bc-peng (approved):
LGTM.👍

PR #1127, opened 2022-12-15
feat(payment): PAYMENTS-8376 Adds ability to enable vaulting for multi-shipping

bc-amit:
## What?
Adds an experiment to the condition that determines if a payment instrument can be vaulted if the order has multi-shipping. 

## Why?
Because we want to enable vaulting for multiple shipping addresses. Which would mean that we won't need `isUsingMultiShipping` prop to determine if the payment method should allow vaulting or not. 

Instead of getting rid of the prop right-away, we are using an experiment to control the rollout.

## Testing / Proof
Additional specs

* Without experiment, we don't see the vaulting checkbox for the multi-shipping. This is the current behaviour.
Screen Shot 2022-12-15 at 2 27 15 pm

* With experiment we see the vaulting checkbox for the multi-shipping. 
 
Screen Shot 2022-12-15 at 2 21 31 pm

### DIFF
--- a/packages/core/src/app/payment/storedInstrument/isInstrumentFeatureAvailable.spec.ts
+++ b/packages/core/src/app/payment/storedInstrument/isInstrumentFeatureAvailable.spec.ts
@@ -68,7 +68,7 @@ describe('isInstrumentFeatureAvailable()', () => {
         ).toBe(false);
     });
 
-    it('returns false if shopper is checking out with multiple shipping address', () => {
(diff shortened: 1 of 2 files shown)


bc-alfreid (approved):
Looks good to me.

davidchin (approved):

bc-peng (approved):
LGTM. Please don't bump `checkout-js` version in BCapp now. We'll do it for you.

PR #1136, opened 2022-12-22
chore(common): CHECKOUT-6445 Move offline component to separate package

animesh1987:
## What?
As above

## Why?
As part of decomposing payment integrations, move `<OfflinePaymentMethod/>` to it's own package and use the new Payment integration API.

## Testing / Proof
- Circle

@bigcommerce/checkout


packages/offline-payment-integration/src/OfflinePaymentMethod.spec.tsx
```diff
@@ -0,0 +1,90 @@
...
+    });
+
+    it('catches error during ApplePay initialization', async () => {
+        jest.spyOn(checkoutService, 'initializePayment').mockRejectedValue(new Error('test error'));
+        mount(<OfflinePaymentMethod {...props} />);
+
+        await new Promise((resolve) => process.nextTick(resolve));
+
+        expect(props.onUnhandledError).toHaveBeenCalled();
+    });
```
bc-peng, line 61: Is this related to ApplePay?

animesh1987: copy paste 🤦

packages/offline-payment-integration/tsconfig.lib.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "./tsconfig.json",
+  "include": [
+    "src/**/*.ts",
+    "src/**/*.tsx"
+  ]
+}
```
bc-peng, line 7: I must miss this one before. Think we discussed removing `lib.json` before. 
The rest look good to me.

packages/offline-payment-integration/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "references": [
+    {
+      "path": "./tsconfig.lib.json"
```
bc-peng, line 5: Same here.

bc-peng (approved):

packages/offline-payment-integration/tsconfig.lib.json
animesh1987: I can't recall correctly the discussion, but can remove this since don't think it's used.

PR #1146, opened 2023-01-04
refactor(checkout): CHECKOUT-6434 Build credit card component package

bc-peng:
## What?
Refactor the existing `CreditCard` component into a shared monorepo package.

## Why?
Lay the foundation work for restructuring `checkout-js` into a monorepo.

## Testing / Proof
- CI checks.


packages/hosted-widget-integration/.eslintrc.json
```diff
@@ -1,3 +1,7 @@
 {
-  "extends": ["../../.eslintrc.json"]
+  "extends": ["../../.eslintrc.json"],
+  "rules": {
+    "@typescript-eslint/no-unsafe-call": "off",
+    "@typescript-eslint/no-unnecessary-condition": "off"
```
bc-peng, line 5: Another side effect of upgrading `card-validator` to 8.1.1

packages/hosted-credit-card-integration/src/HostedCreditCardPaymentMethod.tsx
```diff
@@ -0,0 +1,276 @@
...
+        />
+    );
+};
+
+export default toResolvableComponent<PaymentMethodProps, PaymentMethodResolveId>(
+    HostedCreditCardPaymentMethod,
+    [
+        {
+            id: 'hosted-credit-card',
+            // type: 'PAYMENT_TYPE_API',
```
bc-peng, line 273: If setting it as below, Google Pay wallet button will use this strategy at the payment step. Therefore, I changed it.
```
   id: 'credit-card',
   type: 'PAYMENT_TYPE_API',
```

packages/core/src/app/payment/hostedCreditCard/withHostedCreditCardFieldset.spec.tsx
```diff
@@ -12,7 +12,6 @@ import { act } from 'react-dom/test-utils';
 
 import {
     CardInstrumentFieldsetValues,
-    HostedCreditCardFieldsetValues,
 } from '@bigcommerce/checkout/payment-integration-api';
```
bc-peng, line 15: `HostedCreditCardFieldsetValues` does not exist in the API. So import it from the core on line 26.
It's an error picked up by IDE, not exactly part of the work.

package.json
```diff
@@ -59,9 +59,9 @@
     "@types/shallowequal": "^1.1.0",
     "@types/yup": "^0.26.3",
     "@welldone-software/why-did-you-render": "^3.2.3",
-    "card-validator": "^6.1.0",
+    "card-validator": "^8.1.1",
```
davidchin, line 62: Do you need to do these upgrades as a part of this PR? It might be better to do it separately to reduce the risk?

bc-peng: I have reversed the upgrade, only installed `@types/enzyme-adapter-react-16`.

PR #1148, opened 2023-01-05
fix(checkout): CHECKOUT-7047 Unable to scroll Cart Summary on Mobile

bc-peng:
## What?
Fix the issue when multiple items are added to checkout, the cart summary modal cannot be exited or scrolled.

## Why?
Previously, the modal had a fixed position, so when its height exceeds the height of the screen, the close button is not visible.

## Testing / Proof
- Manual testing:

https://user-images.githubusercontent.com/88361607/210693686-15c2d7de-2c21-4760-9149-ad253d482435.mov

@bigcommerce/checkout

### DIFF
--- a/packages/core/src/scss/components/foundation/modal/_modal.scss
+++ b/packages/core/src/scss/components/foundation/modal/_modal.scss
@@ -40,6 +40,9 @@
         &.is-active,
         &.modal--afterOpen {
             bottom: 2%;
+            display: flex;
+            flex-direction: column;
+            justify-content: end;
             left: 2%;
             max-width: 96%;
             min-height: 96vh;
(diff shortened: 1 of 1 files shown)


packages/core/src/scss/components/foundation/modal/_modal.scss
```diff
@@ -170,3 +167,37 @@
     display: block;
     visibility: visible;
 }
+
+// Specific styles for displaying cart summary on mobile
+body.has-activeModal {
+    overflow: auto;
```
valentindellangela, line 173: 👍

valentindellangela (approved):
LGTM

packages/core/src/scss/components/foundation/modal/_modal.scss
bc-peng: Thanks @valentindellangela. I just found out this bugfix broke BC Wordpress plug-in when opening a cart summary modal. So have to keep digging.

animesh1987 (approved):

PR #1162, opened 2023-01-16
feat(checkout): CHECKOUT-6480 Offline payment method e2e tests

bc-peng:
## What?
Added two e2e tests to ensure all `offline` components of Checkout SDK/JS are working together.

Two tested payment methods are:
- Cash on delivery.
- Pay in store.

## Why?
Reduce the chance of introducing a regression.

## Testing / Proof
CI checks.

@bigcommerce/checkout

### DIFF
--- a/packages/offline-payment-integration/e2e/OfflinePaymentMethod.spec.ts
+++ b/packages/offline-payment-integration/e2e/OfflinePaymentMethod.spec.ts
@@ -1,0 +1,34 @@
+import {
+    PaymentStepAsGuestPreset,
+    test,
+} from '@bigcommerce/checkout/payment-integration-test-framework';
+
+test.describe('Offline payment method', () => {
+    test('`Pay in Store` payment method is working', async ({ assertions, checkout }) => {
+        // Testing environment setup
+        await checkout.use(new PaymentStepAsGuestPreset());
+        await checkout.start('Pay in Store');
+
+        // Playwright actions
+        await checkout.goto();
+        await checkout.selectPaymentMethod('instore');
+        await checkout.placeOrder();
+
+        // Assertions
+        await assertions.shouldSeeOrderConfirmation();
+    });
+
+    test('`Cash on Delivery` payment method is working', async ({ assertions, checkout }) => {
+        // Testing environment setup
+        await checkout.use(new PaymentStepAsGuestPreset());
+        await checkout.start('Cash on Delivery');
+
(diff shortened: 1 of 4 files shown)


packages/payment-integration-test-framework/src/fixture/pageObject/playwright/senstiveDataConfig.ts
```diff
@@ -1,16 +1,21 @@
...
 export const sensitiveHeaders = [
+    'accept',
+    'alt-svc',
     'authorization',
+    'cf-cache-status',
+    'cf-ray',
     'cookie',
+    'nel',
+    'referer',
+    'report-to',
```
bc-peng, line 11: Many new `cloudflare`-related headers appeared in the HAR file, which, in my opinion, revealed far too many tech details. 

Therefore, I updated the `sensitiveHeaders` config.

packages/payment-integration-test-framework/src/fixture/pageObject/playwright/senstiveDataConfig.ts
```diff
@@ -1,16 +1,21 @@
...
     'token',
-    'x-session-hash',
-    'x-xsrf-token',
-    'referer',
     'user-agent',
     'x-checkout-sdk-version',
-    'accept',
     'x-request-id',
+    'x-session-hash',
+    'x-xsrf-token',
```
bc-peng, line 18: Sort all attributes in ascending order as well.

packages/offline-payment-integration/e2e/OfflinePaymentMethod.spec.ts
```diff
@@ -0,0 +1,34 @@
+import {
+    PaymentStepAsGuestPreset,
+    test,
+} from '@bigcommerce/checkout/payment-integration-test-framework';
+
+test.describe('Offline payment method', () => {
+    test('`Pay in Store` payment method is working', async ({ assertions, checkout }) => {
+        // Testing environment setup
+        await checkout.use(new PaymentStepAsGuestPreset());
+        await checkout.start('Pay in Store');
```
animesh1987, line 10: Is this the text on the payment method field?

bc-peng: This is just a name for the HAR file.

animesh1987 (approved):

PR #1164, opened 2023-01-17
fix(payment): PAYPAL-1785 change button name to `Complete Order` on th...

koa75:
…e last step on Checkout page

## What?
- Changed  the button name from `Continue with PayPal` to `Complete order` and  from `Continue with PayLater` to `Complete order` for PayPalCommerce and Braintree providers in the last step on Checkout page for SPB flow
- Leave `Continue with PayPal` button name when this method is selected on the last step of the checkout page without using SPB 

## Why?
At the last step, we complete the payment, but do not proceed to the next step

## Depends on
https://github.com/bigcommerce/bigcommerce/pull/50888

## Testing / Proof
last step for PayPal SPB
Screenshot 2023-01-17 at 18 04 56
 last step for PayLater SPB
Screenshot 2023-01-17 at 18 06 00
last step for Braintree PayLater on checkout page
Screenshot 2023-01-24 at 16 44 06

@bigcommerce/checkout


BC-ACherednichenko (approved):

serhii-tkachenko (approved):
LGTM! 👍

packages/core/src/app/payment/PaymentForm.tsx
```diff
@@ -104,6 +104,18 @@ const PaymentForm: FunctionComponent<
...
+            return;
+        }
+
+        return (
+            selectedMethod.initializationData?.isCheckoutPaymentMethod !== null
+                ? selectedMethod.initializationData?.isCheckoutPaymentMethod
+                : true
+        );
+    }, [selectedMethod]);
+
```
serhii-tkachenko, line 118: This code can be removed

packages/core/src/app/payment/PaymentForm.tsx
```diff
@@ -152,6 +164,7 @@ const PaymentForm: FunctionComponent<
                 ) : (
                     <PaymentSubmitButton
                         brandName={brandName}
+                        isCheckoutPaymentMethod={isCheckoutPaymentMethod}
```
serhii-tkachenko, line 167: It might be better to update this code with:

```suggestion
                        isCheckoutPaymentMethod={!!selectedMethod?.initializationData?.isCheckoutPaymentMethod}
```

packages/core/src/app/payment/PaymentSubmitButton.tsx
```diff
@@ -68,17 +70,28 @@ const PaymentSubmitButtonText: FunctionComponent<PaymentSubmitButtonTextProps> =
...
-                        brandName
-                            ? 'payment.continue_with_brand'
-                            : 'payment.paypal_pay_later_continue_action'
+                        isCheckoutPaymentMethod
+                            ? brandName
+                                ? 'payment.continue_with_brand'
+                                : 'payment.paypal_pay_later_continue_action'
+                            : brandName
+                                ? 'payment.complete_with_brand'
+                                : 'payment.paypal_pay_later_complete_action'
```
serhii-tkachenko, line 94: We usually don't use several if conditions like this. Its better to create two separate variables. 

```suggestion
const continueTranslationId = brandName
    ? 'payment.continue_with_brand'
    : 'payment.paypal_pay_later_continue_action'
const completeTranslationId = brandName
    ? 'payment.complete_with_brand'
    : 'payment.paypal_pay_later_complete_action'

....
                        isCheckoutPaymentMethod
                            ? continueTranslationId
                            : completeTranslationId
```

BC-ACherednichenko (approved):

packages/core/src/app/payment/PaymentSubmitButton.tsx
```diff
@@ -14,6 +14,7 @@ interface PaymentSubmitButtonTextProps {
     methodName?: string;
     initialisationStrategyType?: string;
     brandName?: string;
+    isMethodCheckoutStarted?: boolean;
```
bc-peng, line 17: I understand that this flag indicates whether checkout will be continued or completed.
Can we give it a better name?

koa75: > I understand that this flag indicates whether checkout will be continued or completed. Can we give it a better name?

What name do you suggest?

koa75: We can start to place order on checkout page or via SPB. This flag show we start place order on checkout page after click on payment method

koa75: If we started place order on the checkout page then the button name should be `Continue with PayPal` otherwise `Complete order`

bc-peng: Perhaps `isComplete`? We used it a lot in `checkout-js`.

koa75: done

bc-peng (approved):
LGTM, thanks for your patience.

PR #1167, opened 2023-01-19
test(checkout): CHECKOUT-6472 External payment method e2e tests

bc-peng:
## What?
Added an e2e test to ensure all `external` components of Checkout SDK/JS are working together.

The main testing target of this PR is the `sezzle-integration` which extends `external-integration` in `checkout-sdk-js`.

## Why?
Reduce the chance of introducing a regression.

## Testing / Proof
CI checks.

@bigcommerce/checkout


packages/external-integration/e2e/ExternalPaymentMethod.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+    test('Sezzle is working', async ({ assertions, checkout, page }) => {
+        // Testing environment setup
+        const responseProps = { status: 200, contentType: 'application/json' };
+
+        await checkout.use(new PaymentStepAsGuestPreset());
+        await checkout.start('Sezzle');
+        await checkout.route(
+            'https://sandbox.checkout.sezzle.com?id=**',
+            `${__dirname}/support/sezzle.com.ejs`,
+        );
```
bc-peng, line 18: This test is similar to Google Pay and Apple Pay, mockups needed.
When everything goes as plan, the checkout page will turn to `https://sandbox.checkout.sezzle.com?id=**`, then come back to the order confirmation page.

packages/external-integration/e2e/ExternalPaymentMethod.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+
+test.describe('External payment method', () => {
+    test('Sezzle is working', async ({ assertions, checkout, page }) => {
+        // Testing environment setup
+        const responseProps = { status: 200, contentType: 'application/json' };
+
+        await checkout.use(new PaymentStepAsGuestPreset());
+        await checkout.start('Sezzle');
+        await checkout.route(
+            'https://sandbox.checkout.sezzle.com?id=**',
```
davidchin, line 16: Should this be a part of `sezzle-integration` if it is testing Sezzle specifically?

bc-peng: I agree. I can rename this package as `sezzle-integration` like what I did for the `offsite` payment method.

animesh1987 (approved):

PR #1181, opened 2023-02-01
chore(common): CHECKOUT-6271 Make changes to circle config

animesh1987:
## What?
- Move lint to a separate process
- Remove test artifact uploads
- Split test runs into two jobs one for core package and one for the rest of the packages
- Split test runs of core package into two parallel streams

## Why?
At the moment circlet CI runs take around 10 mins approximately, we can optimise this and have a quicker turn around in our CI steps.

## Testing / Proof
- Circle

Updated CI times for PRs

Screen Shot 2023-02-01 at 5 10 16 pm

@bigcommerce/checkout

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,26 +21,51 @@ orbs:
   node: bigcommerce/internal-node@volatile
 
 jobs:
-  test:
+  test-packages:
     <<: *node_executor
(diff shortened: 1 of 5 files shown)


bc-peng (approved):
Excellent👏👏

PR #1184, opened 2023-02-02
feat(checkout): ADYEN-715 updated credit card icons and added electron and bancontact icons

vitalii-koshovyi:
## What?
Update card icons due to the [Adyen request](https://docs.adyen.com/payment-methods/cards#supported-cards)

## Why?
Due to the Adyen request

## Testing / Proof

**Before:**
Screenshot 2023-02-02 at 15 24 12

**After:**
image

@bigcommerce/checkout


packages/core/src/app/ui/icon/IconCardJCB.tsx
```diff
@@ -3,45 +3,20 @@ import React, { FunctionComponent } from 'react';
...
-        strokeMiterlimit="1.4"
-        viewBox="0 0 480 301"
-        xmlns="http://www.w3.org/2000/svg"
-    >
+    <svg aria-labelledby="iconCardJCBTitle" viewBox="0 0 40 26" xmlns="http://www.w3.org/2000/svg">
         <title id="iconCardJCBTitle">JCB</title>
+        <path d="M0 0h40v26H0z" fill="#fff" />
         <path
-            d="M444 0H24C12 0 0 12 0 24v240.1c0 24 12 36 36 36h408c24 0 36-12 36-36v-228C480 12 468 0 444 0z"
-            fill="#f4f4f4"
```
animesh1987, line 19: Hey @vitalii-koshovyi just wondering do we need to update all the other icons? Have we tested this for other providers as well?

bc-peng: My concern is that we are updating card icons using Adyen's doc, is that a reliable source?

vitalii-koshovyi: @animesh1987  on the screenshot I show the difference between our current icons and updated icons from Adyen cdn. Credit card icons are updated for all providers and will render the same for every provider. Will check manually on other providers before merge.

@bc-peng I believe Adyen has up-to-date icons and I believe we can trust their icon base.
If we have a concern about Adyen as a source of icons, we need to find other sources and update card icons from there, because our credit card icons are a bit outdated.

animesh1987: Tagging @valterfatia for visibility.

valterfatia: Thanks @animesh1987! Although I do agree with the need to update the credit card marks, I'm concerned about us "brute-forcing" it by adopting Adyen's set. From a design perspective the icons should retain the border-radius and have a consistent height.

From a quick internal investigation I believe @bc-jennlindeman was the product designer who last updated our credit marks as seen on https://www.figma.com/file/f5sfaYcRzWTsK38rOC5Tud/Partner-logos-%26-marks?node-id=669%3A1655&t=aDaOF15PC42jykKk-11 and should be involved in this discussion since she has a wider view of where these marks are used.

vitalii-koshovyi: @valterfatia yeah, seems reasonable. Okay, we can wait for all credit card icons and when they'll be ready, update them. I'll leave this PR that designers can check what icons need to be updated/added.

vitalii-koshovyi: @valterfatia as you can see we already have icons that are out of our guideline. I believe we need to fix this soon, as our  partners recommend us to update these icons already.

bc-jennlindeman:
Heyo! I did last update the icons in that file I'm tagged on. I don't know how up-to-date they are, but I definitely recommend we adhere to updating them all to be that size/style. These are the same ones we used for mobile payment icons. If we need anyone to review them for update needs, that can be done but I don't know who would do that. Screenshot of the style we need to use:
Screenshot 2023-02-15 at 12 39 58 PM

These "after" screenshots you sent are not at all that style/size so we need to update our [library in figma](https://www.figma.com/file/f5sfaYcRzWTsK38rOC5Tud/Partner-logos-%26-marks?node-id=0%3A1&t=tgq5EQOyVDTmgDk3-0) where an icon needs refreshed image, and then use those, not the proposed "after" ones. Does that make sense?

bc-alextop:
Hi @bc-jennlindeman, thanks for your explanation - it's a bit clearer now. I believe in terms of Adyen's request we need to:

1. review and update the existing outdated icons 
2. add new icons which are available in Adyen library but missing in BC library

Can you suggest who is the best person in design team to help handle this request?

bc-jennlindeman:
@bc-alextop do you have a timeline of when they are needed if there is one? I can probably handle figuring out the details and sending you the updated ones when ready, but I will be out tomorrow - next week. I'll also need to align from the pd team/mgmt whether or not we need to confirm updates with either marketing or anyone on the payment team (PM or product marketing). Probably not but I'll figure that out.

Where did you get these proposed updated payment icons? I can use that route to short cut some of this.

bc-alextop:
@bc-jennlindeman We don't have the exact timeline for this task but it would be great to get the updated icons by early March if that works for you :) 
Adyen team ensures that their library contains the most up to date logos.  You can find the full list of Adyen payment methods in [this tracker](https://docs.google.com/spreadsheets/d/1l9uuT3IqZhQg19qZ-QavXbSdM2U8UW8oQBf3bIG-v8o/edit#gid=1800997837). In **column D** you will find the link that leads to the respective payment method icon (you just need to replace cell parameter in the link with the payment method name stated in column B). Those 50 payment methods will be available for Adyen V3 connector.

So we need:
1. review and update the existing outdated icons
2. add new icons which are available in Adyen library but missing in BC library

Thanks in advance for your help!

bc-0dp:
I don't have access to comment on the tracker.
But shouldn't `DN` for Dankort and MobilePay be `DK` for Denmark?

bc-alextop:
@bc-0dp  I have corrected column G for those two payment methods. Thanks for notifying!

bc-alextop:
Hi @bc-jennlindeman, I just wanted to check if I need to submit a ticket somewhere to get the icons update work started?

bc-jennlindeman:
@bc-alextop I'll make a ticket. I was out all last week but can complete them for ya this week.

bc-jennlindeman:
In column A there's 56 providers listed. Is this the number I need to provide? I also don't see how you pulled the updated imagery to propose the second/updated proposal in those images above. I'm going to need to get vectors or svg's that I can add to our badge layout styling.

bc-jennlindeman:
https://www.figma.com/file/f5sfaYcRzWTsK38rOC5Tud/Partner-logos-%26-marks?node-id=0%3A1&t=SzKXXtZ2dTJBR2WA-1

bc-jennlindeman:
Screenshot 2023-03-08 at 5 29 28 PM

bc-alextop:
Thanks @bc-jennlindeman !

animesh1987 (approved):

PR #1186, opened 2023-02-05
feat(checkout): CHECKOUT-7255 Remove checkout step numbers

bc-peng:
## What?
Removing all checkout step numbers.
- Remove the step counter of each unfinished checkout step.
- Remove the circle, check mark of each finished checkout step.
- Change the step header’s left padding to zero.
- Putting all changes behind an experiment so that we can roll out changes incrementally.

## Why?
Once we introduce wallet buttons on the top of the checkout page, the step numbers are no longer accurate progress indicators.

## Testing / Proof

customer

shipping

payment

https://user-images.githubusercontent.com/88361607/216851440-9d7db6c9-6f80-4427-8cd9-02e18ff33442.mov

@bigcommerce/checkout

### DIFF
--- a/packages/core/src/app/checkout/Checkout.tsx
+++ b/packages/core/src/app/checkout/Checkout.tsx
@@ -292,7 +292,7 @@ class Checkout extends Component<
         }
 
         return (
-            <div className={classNames({ 'is-embedded': isEmbedded(), 'remove_checkout_step_numbers': isHidingStepNumbers })}>
+            <div className={classNames({ 'is-embedded': isEmbedded(), 'remove-checkout-step-numbers': isHidingStepNumbers })}>
                 <div className="layout optimizedCheckout-contentPrimary">
                     {this.renderContent()}
                 </div>
--- a/packages/core/src/app/checkout/Checkout.tsx
+++ b/packages/core/src/app/checkout/Checkout.tsx
@@ -126,6 +126,7 @@ export interface CheckoutState {
     isRedirecting: boolean;
     hasSelectedShippingOptions: boolean;
     isBuyNowCartEnabled: boolean;
+    isHidingStepNumbers: boolean;
     isSubscribed: boolean;
 }
 
@@ -162,6 +163,7 @@ class Checkout extends Component<
         isMultiShippingMode: false,
         hasSelectedShippingOptions: false,
         isBuyNowCartEnabled: false,
+        isHidingStepNumbers: true,
         isSubscribed: false,
     };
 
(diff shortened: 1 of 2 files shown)


packages/core/src/app/checkout/Checkout.tsx
```diff
@@ -286,7 +292,7 @@ class Checkout extends Component<
         }
 
         return (
-            <div className={classNames({ 'is-embedded': isEmbedded() })}>
+            <div className={classNames({ 'is-embedded': isEmbedded(), 'remove_checkout_step_numbers': isHidingStepNumbers })}>
```
bc-peng, line 295: Initially, I changed the html structure as well, but then I realised I could control every UI element with pure CSS if there was a specific class name at the top level.

animesh1987, line 295:
```suggestion
            <div className={classNames({ 'is-embedded': isEmbedded(), 'remove-checkout-step-numbers': isHidingStepNumbers })}>
```

animesh1987: Also should we use `remove-checkout-step-numbers` ?

bc-peng: I believe we will remove this class name this or next month, so it is only there to support the experiment's roll-out.
Just want to confirm, are you suggesting changing a name or not to use this approach?

animesh1987: Just changing the name.

bc-peng: What name(s) do you have in mind?

bc-peng: Since it is most likely for us to have a toggle on this, maybe `is-hiding-step-numbers`?

bc-peng:
Hi @animesh1987, I have updated the code and tested it on my store. Can I have another review?

packages/core/src/scss/components/checkout/checkoutSteps/_checkoutSteps.scss
```diff
@@ -138,6 +138,21 @@
...
+        div.stepHeader-counter.icon {
+            display: none;
+        }
+
+        .checkout-form,
+        div.stepHeader-body,
+        div.checklist-skeleton,
+        h2.stepHeader-title {
+            margin-left: 0;
+        }
```
animesh1987, line 152: Looks good, but just wondering do we need to add `div, li, h2` as a part of css identifiers?

bc-peng: By doing so, I feel confident they're at the top of the checkout CSS hierarchy. 😂

animesh1987 (approved):

PR #1192, opened 2023-02-09
feat(checkout): CHECKOUT-7231 Display wallet buttons at top of checkout

bc-peng:
## What?
- Introduce a new component to display wallet buttons at the top of the checkout page.
- This component only shows up when both the experiment, `CHECKOUT-7222.checkout_settings_styling_section`, and the setting, `walletButtonsOnTop`, are on.
- The current button list UX stays the same when the experiment is off.

## Why?
Optimize the use of wallet buttons on the platform.

## Testing / Proof
This PR is only for structural changes. The UX/UI is not yet complete.

image

@bigcommerce/checkout


.gitignore
```diff
@@ -1,5 +1,6 @@
 .idea
 .vscode
+junit.xml
```
bc-peng, line 3: When running tests locally, this file pops up. Think we may ignore it.

packages/core/src/scss/components/checkout/customer/_customer.scss
```diff
@@ -91,3 +91,49 @@
     margin-right: spacing("single");
     width: 100%;
 }
+
+.wallet-buttons-container {
```
bc-peng, line 95: Everything here is not final.
All of them could be rewritten in the next ticket.

.gitignore
animesh1987: This is for CI only.

packages/core/src/app/customer/CheckoutButtonContainer.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const config = getConfig();
+
+  if (!config) {
+    return null;
+  }
+
+  return (
+    <div className='wallet-buttons-container'>
+      <CheckoutButtonListV1
+      checkEmbeddedSupport={checkEmbeddedSupport}
```
animesh1987, line 37: Indentation?

packages/core/src/app/customer/Customer.tsx
```diff
@@ -171,28 +172,30 @@ class Customer extends Component<CustomerProps & WithCheckoutCustomerProps & Ana
...
-                        <CheckoutButtonList
-                            checkEmbeddedSupport={checkEmbeddedSupport}
-                            deinitialize={deinitializeCustomer}
-                            initialize={initializeCustomer}
-                            isInitializing={isInitializing}
-                            methodIds={checkoutButtonIds}
-                            onError={onUnhandledError}
-                        />
-                    }
+                    checkoutButtons={checkoutButtons}
```
animesh1987, line 198: @AlanMunozN do we need to make some adjustments for wallet buttons on top of checkout approach. Since with the current approach of experiment and setting is on then wallet buttons will be visible even if shopper authenticates with stripe link.

packages/core/src/app/customer/CheckoutButtonContainer.tsx
```diff
@@ -34,13 +34,13 @@ const CheckoutButtonContainer: FunctionComponent<CheckoutButtonsListOnTopProps &
...
     <div className='wallet-buttons-container'>
       <CheckoutButtonListV1
-      checkEmbeddedSupport={checkEmbeddedSupport}
-      copywritingStringId='remote.start_with_text'
-      deinitialize={checkoutService.deinitializeCustomer}
-      initialize={checkoutService.initializeCustomer}
-      isInitializing={isInitializingCustomer()}
-      methodIds={config.checkoutSettings.remoteCheckoutProviders}
-      onError={onUnhandledError}
+        checkEmbeddedSupport={checkEmbeddedSupport}
```
animesh1987, line 37: We usually have an indentation of 4?

animesh1987 (approved):
One very minor comment, but nice work over all. LGTM 👍

PR #1194, opened 2023-02-10
fix(checkout): Re-initialize customer step payment buttons if cart balance has changed

vitalii-koshovyi:
## What?
Re-initialize customer step payment buttons if cart balance has changed.

## Why?
We've found that Google Pay does not include coupon discount at the 'Customer' step of checkout.
I decided to re-initialize customer step payment buttons if the cart balance has changed.

## Testing / Proof
**Before fix:**

https://user-images.githubusercontent.com/79574476/218151786-bacaa63a-82f6-4c4e-b267-d525a6a6be28.mov

**After fix:**

https://user-images.githubusercontent.com/79574476/218151769-9ca95fb7-be10-4c10-96a9-781739f1e677.mov

@bigcommerce/checkout

### DIFF
--- a/packages/core/src/app/customer/CheckoutButton.tsx
+++ b/packages/core/src/app/customer/CheckoutButton.tsx
@@ -1,15 +1,18 @@
 import { CustomerInitializeOptions, CustomerRequestOptions } from '@bigcommerce/checkout-sdk';
 import React, { PureComponent } from 'react';
 
+import { CheckoutContextProps, withCheckout } from '../checkout';
+
 export interface CheckoutButtonProps {
     containerId: string;
     methodId: string;
+    outstandingBalance?: number;
     deinitialize(options: CustomerRequestOptions): void;
     initialize(options: CustomerInitializeOptions): void;
     onError?(error: Error): void;
 }
 
-export default class CheckoutButton extends PureComponent<CheckoutButtonProps> {
+class CheckoutButton extends PureComponent<CheckoutButtonProps> {
     componentDidMount() {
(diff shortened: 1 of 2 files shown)


packages/core/src/app/customer/CheckoutButton.tsx
```diff
@@ -34,3 +52,12 @@ export default class CheckoutButton extends PureComponent<CheckoutButtonProps> {
...
     }
 }
+
+export default withCheckout(({
+    checkoutState,
+}: CheckoutContextProps) => {
+    const { data: { getCheckout } } = checkoutState;
+    const { outstandingBalance } = getCheckout() || {};
+
+    return { outstandingBalance };
```
bc-peng, line 62: Could we put them inside a function for readability?

animesh1987:
Interesting issue, usually we build the payment request for example for apple pay when the button clicked so we get the updated cart amount. 
Can we do the same for google pay? Since once you apply the coupon the cart is updated with the correct price, so you should be able to handle that logic in the strategy rather than refreshing the button components.

packages/core/src/app/customer/CheckoutButton.tsx
vitalii-koshovyi: Sorry, didn't understand what you mean

vitalii-koshovyi:
@animesh1987 I'll check it, but from what I saw earlier our implementation of google pay doesn't update price when we update outstandingBalance with coupon.

vitalii-koshovyi:
Closed this PR due to found better solution https://github.com/bigcommerce/checkout-sdk-js/pull/1839

PR #1196, opened 2023-02-13
refactor(checkout): CHECKOUT-7291 Add ApplePay button style

bc-peng:
## What?
After removing Apple Pay button inline styles from `checkout-sdk`, add those styles to `checkout-js`'s SCSS file.

## Depends on
Will bump SDK version in this PR once the below is merged.
- https://github.com/bigcommerce/checkout-sdk-js/pull/1836

## Why?
`checkout-js` can manage the appearance of an Apple Pay wallet button in the new wallet button container.

## Testing / Proof

Screenshot 2023-02-13 at 4 22 19 pm

@bigcommerce/checkout

### DIFF
--- a/packages/core/src/scss/components/checkout/checkoutRemote/_checkoutRemote.scss
+++ b/packages/core/src/scss/components/checkout/checkoutRemote/_checkoutRemote.scss
@@ -16,6 +16,12 @@
         width: remCalc(200px);
     }
 
+    > #applepayCheckoutButton button,
+    > #braintreepaypalcreditCheckoutButton,
+    > #braintreepaypalCheckoutButton {
+        width: 160px;
+    }
+
     > #applepayCheckoutButton button {
         background-color: #000;
         background-image: -webkit-named-image(apple-pay-logo-white);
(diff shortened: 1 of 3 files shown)


snaderiBC (approved):
Are we sure we want to do this as part of `CHECKOUT-0000`?

bc-peng:
> Are we sure we want to do this as part of `CHECKOUT-0000`?

Thanks for the reminder. I have created a sub ticket, [CHECKOUT-7291](https://bigcommercecloud.atlassian.net/browse/CHECKOUT-7291), to track this.

animesh1987:
The alignment of buttons looks a bit off, can we update that as well please.

packages/core/src/scss/components/checkout/checkoutRemote/_checkoutRemote.scss
```diff
@@ -15,6 +15,22 @@
     > .AmazonPayContainer {
         width: remCalc(200px);
     }
+
+    > #applepayCheckoutButton button,
+    > #braintreepaypalcreditCheckoutButton,
+    > #braintreepaypalCheckoutButton {
+        width: 160px;
+    }
```
bc-peng, line 23: @animesh1987 It turns out that both PayPal buttons are 10px short, caused by PayPal SDK.
This is the result of applying the width override.
Screenshot

animesh1987 (approved):

PR #1200, opened 2023-02-15
fix(shipping): STRIPE-202 Split name input into First and Last name

JavvmBC:
## What? [STRIPE-202](https://bigcommercecloud.atlassian.net/browse/STRIPE-202)
Split name input into First and Last name

## Why?
To have a better input field design

## Based documentation
[stripe_doc](https://stripe.com/docs/js/elements_object/create_address_element)

## Refactor
Noticed that there was a missing space in the stripe-upe-payment-strategy.ts so I added it to be more readable

## Testing / Proof
200685408-3923f183-8561-42a8-bd7a-49f864e01dc2

Screenshot 2023-02-15 at 14 33 24

## Depends on
[1675](https://github.com/bigcommerce/checkout-sdk-js/pull/1675)

### DIFF
--- a/packages/core/src/app/shipping/stripeUPE/StripeShippingAddress.tsx
+++ b/packages/core/src/app/shipping/stripeUPE/StripeShippingAddress.tsx
@@ -109,20 +109,21 @@ const StripeShippingAddress: FunctionComponent<StripeShippingAddressProps> = (pr
 
     const handleStripeShippingAddress = useCallback(async (shipping: StripeShippingEvent) => {
         const {complete, phoneFieldRequired, value: { address = { country: '', state: '', line1: '', line2: '', city: '', postal_code: '' }
-            , name = '', phone = '' } } = shipping;
+            , name = '', firstName = '', lastName = '', phone = '' } } = shipping;
 
         if (complete) {
             if (shouldShowContent(shipping?.isNewAddress, phoneFieldRequired, phone)) {
                 handleLoading();
             }
 
-            const names = name.split(' ');
+            const names = name?.split(' ');
+
             // @ts-ignore
             const country = availableShippingList?.find(country => country.code === address.country).name;
             const state = StripeStateMapper(address.country, address.state);
             const shippingValue = {
-                firstName: names[0],
-                lastName: names[1] || ' ',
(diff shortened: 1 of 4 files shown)


AlanMunozN (approved):

BC-ISilva:
💚 Tested!
QE Details:

PR implementation is working fine for Stripe Link. 

Following scenarios have been validated:

- Stripe’s address element will now display two separate fields for name: First Name and Last Name
- The first name value  will be mapped to First Name on the BigCommerce order.shipping_address object and order.billing_address object
- The last name value will be mapped to Last Name on the BigCommerce order.shipping_address object and order.billing_address object.
- Implementation only available for Stripe Link.

Please review attached word file document.
[INT-7335 Split Full Name to First Name And Last Name On Address Element.docx](https://github.com/bigcommerce/checkout-js/files/10768455/INT-7335.Split.Full.Name.to.First.Name.And.Last.Name.On.Address.Element.docx)

packages/core/src/app/shipping/stripeUPE/StripeShippingAddress.tsx
```diff
@@ -101,20 +101,21 @@ const StripeShippingAddress: FunctionComponent<StripeShippingAddressProps> = (pr
...
+            const names = name?.split(' ');
+
             // @ts-ignore
             const country = availableShippingList?.find(country => country.code === address.country).name;
             const state = StripeStateMapper(address.country, address.state);
             const shippingValue = {
-                firstName: names[0],
-                lastName: names[1] || ' ',
+                firstName: firstName || names[0],
+                lastName: lastName || names[1],
```
animesh1987, line 118: Nice work, will we be removing the names[0]/[1] approach in future?

animesh1987 (approved):

packages/core/src/app/shipping/stripeUPE/StripeShippingAddress.tsx
JavvmBC: Thank you, it would be better to have them as a fall back, if you have display split or full the variable name will always come with data so it is a useful just in case the split name fails.

PR #1202, opened 2023-02-16
feat(payment): INT-7503 adjusting stripe button style for stripe link

AlanMunozN:
## What? [INT-7503](https://bigcommercecloud.atlassian.net/browse/INT-7503)
adjusting stripe button style to match with the StripeLink iframe when the floating label experiment is on

**Why the change in display style:**

when we are creating a Stripe iframe with the new properties for stripe changes, the behavior is different depending on the browser.

We think it’s because the display style doesn’t reserve the space, which causes stripe to send us a very small input size in chrome, which doesn’t happen in other browsers (it works fine in safari).

To make it work in all browsers, we added visibility hidden to reserve the size but this can cause the shipping or customer step is growing up before time( the iframe adds the extra space before we have to show it, for example in the shipping step, there are several fields so the behavior is very visible, which causes a bad UI ).

To avoid this we simulate a display behavior but with a reserved space by adding position relative and absolute in such a way that when the component is loaded it does not visually affect the size of the step and when it is ready to be displayed it returns to its relative position.

## Why?
for StripeLink, we are looking to match the same styles used by BC

## Testing / Proof
**Customer**
**BC  / vs / LINK**
image
**Shipping**
**BC  / vs / LINK**
image

## Merge at the same time
https://github.com/bigcommerce/checkout-sdk-js/pull/1854

[INT-7503]: https://bigcommercecloud.atlassian.net/browse/INT-7503?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BC-ISilva:
💚 Tested!

User story implementation has been validated, following scenarios were tested:

1. Experiment “CHECKOUT-6879.enable_floating_labels“ OFF
2. Experiment “CHECKOUT-6879.enable_floating_labels“ ON 
3. All Stripe elements, labels are rendered inside the corresponding text field
4. All Stripe elements, when the shopper starts typing, then labels. shall remain and float to the top of the field allowing for the entered text to be prominent.
5. Stripe Elements will inherit all style rules defined in Page Builder (Page Builder > Theme Styles > Checkout Page > Form Input Fields) so that the Stripe Elements match the style of the native BigCommerce elements.
6. Works across desktop and mobile browsers (Safari and FireFox browsers were used for validation).

Please review attached word file with validation details.
[INT-7503  - [Link] Support Floating Labels for Stripe Address Element.docx](https://github.com/bigcommerce/checkout-js/files/10851458/INT-7503.-.Link.Support.Floating.Labels.for.Stripe.Address.Element.docx)

JavvmBC (approved):

packages/core/src/app/customer/StripeGuestForm.tsx
```diff
@@ -2,13 +2,14 @@ import { CustomerInitializeOptions, CustomerRequestOptions } from '@bigcommerce/
 import { FieldProps, FormikProps, withFormik } from 'formik';
 import React, { FunctionComponent, memo, ReactNode, useCallback, useEffect, useState } from 'react';
 
+import { CustomerSkeleton } from '@bigcommerce/checkout/ui';
```
bc-peng, line 5: Recently, we got rid of the loading process for the customer step with https://github.com/bigcommerce/checkout-js/pull/1204, could you please check if it is also viable for `Stripe`?

AlanMunozN: Hey @bc-peng, the problem is that it was a requirement made by Stripe since the loading times of the iframe are very long compared to the button, an apology

packages/core/src/scss/components/checkout/customer/_customer.scss
```diff
@@ -49,9 +49,15 @@
...
+    @extend .customerEmail-button;
+
     bottom: 3px;
-    padding: $customerEmail-button-padding;
-    width: 100%;
+
+    &--withFloatingLabels {
+        @extend .customerEmail-button;
+
+        bottom: 23px;
```
bc-peng, line 59: Sorry, forgot to mention, `23px` can be replaced with `spacing("single") + spacing("half")`.
We strongly oppose hardcoding values in SCSS files because it may not be consistent with the existing UI elements.

packages/core/src/app/customer/StripeGuestForm.tsx
bc-peng: OK, I was thinking removing `CustomerSkeleton` component.
In this case, we'll keep it.

packages/core/src/scss/components/checkout/customer/_customer.scss
AlanMunozN: @bc-peng totally understandable, I already tried the change and it works perfectly 🙇‍♂️

packages/core/src/scss/components/checkout/customer/_customer.scss
```diff
@@ -49,9 +49,15 @@
 }
 
 .stripeCustomerEmail-button {
```
animesh1987, line 51: Just thinking should we have a separate file for stripe styles? I mention this because in future we would like to move all stripe related code to it's own package which means we probably also look at moving the styles across as well.

AlanMunozN: Hey @animesh1987, Is this approach ok for you?

please let me know your thoughts

packages/core/src/scss/components/checkout/customer/_customer.scss
AlanMunozN: @bc-peng sorry friend, I think the day I uploaded those changes in my local did not take them correctly (with spacing("single") + spacing("half") the button looks very high, it gives like 2.25 rem), I was playing for a while with the spacing and I found another one, let me know if it's okay for you, I attach an image of how it looks (it gives 1.78125rem)
image

bc-peng: Thanks @AlanMunozN. I just spotted a difference on the screenshots.

219497079-5d0d2faf-d8ca-4021-973a-64693732e544
219496950-dfca4dd2-79a3-496e-8f92-169882031982

The space between fields in the checkout has been specified as `8px` by the designer. Can you please match Stripe forms with that? You simply need to use `$floating-form-field-spacing` as the value.

The rest looks good to me.

AlanMunozN: Hey @animesh1987 and @bc-peng, I made changes, and it's ready for another review ♻️

packages/core/src/app/customer/Customer.tsx
```diff
@@ -212,6 +212,7 @@ class Customer extends Component<CustomerProps & WithCheckoutCustomerProps & Ana
                     privacyPolicyUrl={privacyPolicyUrl}
                     requiresMarketingConsent={requiresMarketingConsent}
                     step={step}
+                    useFloatingLabel={useFloatingLabel}
```
bc-peng, line 215: Can you please rename all `useFloatingLabel` to `isFloatingLabelEnabled`?
We recently did this change to match the return flag from API in this PR, https://github.com/bigcommerce/checkout-js/pull/1230.

packages/core/src/app/shipping/stripeUPE/utils/stripe-upe-custom-styles.ts
bc-peng, line 1: This file contains a lot of hardcoded CSS style values.
Could you please manage those styles using the related SCSS file and existing size patterns, such as `$floating-form-field-spacing`, `$fontSize-root`? Because we may change those spacings and sizes in the future.

packages/core/src/scss/components/checkout/customer/_stripeUpeCustomer.scss
```diff
@@ -0,0 +1,11 @@
+.stripeCustomerEmail-button {
+    bottom: 3px;
```
bc-peng, line 2: Can you replace `3px` with a pre-defined value?

AlanMunozN:
@animesh1987 @bc-peng hey guys, I already made changes, please let me know if this approach is ok, also I had to modify packages/core/src/app/common/utility/isFloatingLabelEnabled.ts because always returns false but I'll delete that in the next or final review

packages/core/src/app/common/utility/isFloatingLabelEnabled.ts
```diff
@@ -1,5 +1,6 @@
 import { CheckoutSettings } from '@bigcommerce/checkout-sdk';
 
 export default function isFloatingLabelEnabled(checkoutSettings: CheckoutSettings): boolean {
-    return Boolean(checkoutSettings.checkoutUserExperienceSettings.floatingLabelEnabled);
+    console.log('ok:', checkoutSettings.features['CHECKOUT-6879.enable_floating_labels']);
+    return checkoutSettings.features['CHECKOUT-6879.enable_floating_labels'];
```
bc-peng, line 5: Just checking, so this file should be untouched when it is merged, isn't it?

packages/core/src/app/payment/paymentMethod/StripeUPEPaymentMethod.tsx
```diff
@@ -18,13 +20,14 @@ export type StripePaymentMethodProps = Omit<HostedWidgetPaymentMethodProps, 'con
...
     storeUrl: string;
+    isFloatingLabelEnabled: boolean;
 }
 
 const StripeUPEPaymentMethod: FunctionComponent<
     StripePaymentMethodProps &
         WithInjectedHostedCreditCardFieldsetProps &
         WithCheckoutStripePaymentMethodProps
-> = ({ initializePayment, method, storeUrl, onUnhandledError = noop, ...rest }) => {
+> = ({ initializePayment, method, storeUrl , onUnhandledError = noop, isFloatingLabelEnabled, ...rest }) => {
```
bc-peng, line 30:
```suggestion
> = ({ initializePayment, method, storeUrl, onUnhandledError = noop, isFloatingLabelEnabled, ...rest }) => {
```

packages/core/src/app/shipping/stripeUPE/utils/stripe-upe-custom-styles.ts
```diff
@@ -0,0 +1,86 @@
...
+        return spaceUnit;
+    }
+
+    const stripeDefault = {
+        'border-radius': '4px',
+        'font-size': '16px',
+        padding: '6px 12px 3.33333px 12px',
+        'padding-left': '4.3',
+        'padding-right': '2.8'
+    }
```
bc-peng, line 30: Can we also put this into the SCSS file?

AlanMunozN:
@bc-peng, I made changes, It's ready for another review ♻️

packages/core/src/scss/components/checkout/stripeLink/_stripeDefaultValues.scss
```diff
@@ -0,0 +1,13 @@
+@import "../../../../../src/app/ui/base";
+
+// We don't use these styles for the floating labels in the iframe.
+// These styles are in case of fallback for stripeIframe.scss
+// -----------------------------------------------------------------------------
+// stylelint-disable-next-line selector-pseudo-class-no-unknown
+:export {
+    border-radius: 4px;
```
bc-peng, line 8:
```suggestion
    border-radius: $global-radius;
```

packages/core/src/scss/components/checkout/stripeLink/_stripeDefaultValues.scss
```diff
@@ -0,0 +1,13 @@
...
+    font-size: 16px;
+    padding: 6px 12px 3.3px 12px;
+    padding-left: 4.3px;
+    padding-right: 2.8px;
```
bc-peng, line 12: Just wondering, how did you come up with those decimal padding values?
Why are they different from that in `_stripeIframe.scss`?

bc-peng: I saw this line `typeof stripeIframe === 'object' ? stripeIframe : defaultIframeValues`.
Should `typeof stripeIframe === 'object'` always be `true` since you have the full control here, if I am not mistaken?

packages/core/src/scss/components/checkout/stripeLink/_stripeDefaultValues.scss
AlanMunozN: @bc-peng the idea is to avoid dynamic data, but let me know if it's causing any trouble we could remove the idea

packages/core/src/scss/components/checkout/stripeLink/_stripeDefaultValues.scss
AlanMunozN: @bc-peng, the reason we use something by default that only stripeLink could manipulate is to avoid the use of dynamic data (without sass variables)

The decimal padding values are the result that I get when I inspected the elements, I use a default value only in case there is a failure or an accidental change in the sass variables, the idea is that these styles will never be used, I just add them as an extra precaution, since the component usually(depending of the browser) looks very bad without  styles, especially if the `floating labels` is activated, if you feel that it can make any trouble we can delete it :D

And regarding the last question you are correct, an apology, it was from a previous approach, indeed the sentence does not make sense, in this case, it should be `||` since in case of not sending styles the default value will be used, I will fix it

packages/core/src/app/shipping/stripeUPE/utils/stripe-upe-custom-styles.ts
```diff
@@ -0,0 +1,79 @@
+import defaultIframeValues from '../../../../scss/components/checkout/stripeLink/_stripeDefaultValues.scss';
+import stripeIframe from '../../../../scss/components/checkout/stripeLink/_stripeIframe.scss';
+
+export interface StripeStyles {
+    [key: string]: string;
+}
+
+export const getStripeCustomStyles = (
+    styles: StripeStyles | undefined | boolean,
```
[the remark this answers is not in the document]
AlanMunozN: @bc-peng as far as I know never since the values are extracted, and if it isn't found will return undefined, but someone some time ago added the boolean value for the interface of the customer component, I'm not sure if it was for some particular functionality

PR #1204, opened 2023-02-20
feat(common): CHECKOUT-7234 Remove loading for customer guest form

animesh1987:
## What?
As above

## Why?
As we are moving wallet buttons to top of checkout, we can render the form straightaway while the buttons are rendered in their own container. 

Therefore checkout page can have a perception that it loads faster.

## Testing / Proof
- Circle

## Experiment off 
https://user-images.githubusercontent.com/7134802/220211072-3d721c57-16e2-4c62-9ceb-44d4a7e8a118.mov

## Experiment on
https://user-images.githubusercontent.com/7134802/220211130-fdcbdc51-a6aa-4636-805a-03fbef391582.mov

@bigcommerce/checkout

### DIFF
--- a/packages/core/src/app/checkout/Checkout.tsx
+++ b/packages/core/src/app/checkout/Checkout.tsx
@@ -26,11 +26,11 @@ import { retry } from '../common/utility';
 import {
     CheckoutButtonContainer,
     CheckoutSuggestion,
+    Customer,
     CustomerInfo,
     CustomerSignOutEvent,
     CustomerViewType,
 } from '../customer';
-import Customer from '../customer/Customer';
 import { EmbeddedCheckoutStylesheet, isEmbedded } from '../embeddedCheckout';
 import { TranslatedString, withLanguage, WithLanguageProps } from '../locale';
 import { PromotionBannerList } from '../promotion';
--- a/packages/core/src/app/checkout/Checkout.tsx
+++ b/packages/core/src/app/checkout/Checkout.tsx
@@ -16,7 +16,7 @@ import { find, findIndex } from 'lodash';
 import React, { Component, lazy, ReactNode } from 'react';
 
 import { AnalyticsContextProps } from '@bigcommerce/checkout/analytics';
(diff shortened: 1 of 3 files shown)


packages/core/src/app/customer/Customer.tsx
```diff
@@ -117,7 +117,7 @@ class Customer extends Component<CustomerProps & WithCheckoutCustomerProps & Ana
         this.draftEmail = email;
 
         try {
-            if (providerWithCustomCheckout !== PaymentMethodId.StripeUPE) {
+            if (providerWithCustomCheckout && providerWithCustomCheckout !== PaymentMethodId.StripeUPE) {
```
bc-peng, line 120: What other providers do we have now? This looks like a bugfix then. Something should have been wrong without checking `providerWithCustomCheckout` before.

packages/core/src/app/checkout/Checkout.tsx
```diff
@@ -392,25 +383,23 @@ class Checkout extends Component<
...
+                    onContinueAsGuest={this.navigateToNextIncompleteStep}
+                    onContinueAsGuestError={this.handleError}
+                    onReady={this.handleReady}
+                    onSignIn={this.navigateToNextIncompleteStep}
+                    onSignInError={this.handleError}
+                    onSubscribeToNewsletter={this.handleNewsletterSubscription}
+                    onUnhandledError={this.handleUnhandledError}
+                    step={step}
+                    viewType={customerViewType}
+                />
```
bc-peng, line 402: How is the experience with the old wallet button list after the skeleton removal? Is there anything we need to adjust?

packages/core/src/app/customer/Customer.tsx
animesh1987: We have stripe and bolt as `providerWithCustomCheckout`, this bit of code should only execute in the case of Bolt as far as I understand. 

cc @AlanMunozN can you please confirm?

AlanMunozN: @animesh1987 yes, should only execute in the case of Bolt


image

bc-peng (approved):
Nice work👏 Let's ship it.

PR #1209, opened 2023-02-23
feat(checkout): CHECKOUT-7233 Introduce new wallet button container

bc-peng:
## What?
- Add a wallet button loading skeleton.
- Add a responsive layout to hold unlimited number of wallet buttons.
- Hide `LoadingNotification` component when using the new wallet button container.

## Why?
Optimize the use of wallet buttons on the platform.

## Testing / Proof

https://user-images.githubusercontent.com/88361607/223907650-b705da93-d9da-433a-b40f-25b29b82f218.mov

@bigcommerce/checkout


packages/core/src/app/customer/CheckoutButtonContainer.tsx
```diff
@@ -27,22 +37,63 @@ const CheckoutButtonContainer: FunctionComponent<CheckoutButtonsListOnTopProps &
...
+            <div className={classNames({
+                'checkout-buttons--1': buttonsCount === 1,
+                'checkout-buttons--2': buttonsCount === 2,
+                'checkout-buttons--3': buttonsCount === 3,
+                'checkout-buttons--4': buttonsCount === 4,
+                'checkout-buttons--5': buttonsCount === 5,
+                'checkout-buttons--n': buttonsCount > 5,
+            })}>
+                <WalletButtonsContainerSkeleton buttonsCount={buttonsCount} isLoading={isLoading}>
+                    <CheckoutButtonListV1
```
davidchin, line 86: When are we switching over to `CheckoutButtonListV2`? We want the buttons to load from payment integration packages. cc @animesh1987

packages/core/src/app/customer/CheckoutButtonList.tsx
```diff
@@ -36,28 +36,32 @@ export const SUPPORTED_METHODS: string[] = [
...
     isInitializing?: boolean;
-    copywritingStringId?: string;
+    hideText?: boolean;
     checkEmbeddedSupport?(methodIds: string[]): void;
     deinitialize(options: CustomerRequestOptions): void;
     initialize(options: CustomerInitializeOptions): void;
     onError?(error: Error): void;
 }
 
+export const filterUnsupportedMethodIds = (methodIds:string[]):string[] => {
```
davidchin, line 46: Is our ESLint not working? IIRC I thought we have a rule on [key-spacing](https://eslint.org/docs/latest/rules/key-spacing)?

packages/core/src/app/customer/CheckoutButtonContainer.tsx
```diff
@@ -27,22 +37,63 @@ const CheckoutButtonContainer: FunctionComponent<CheckoutButtonsListOnTopProps &
         },
     } = checkoutState;
     const config = getConfig();
+    const methodIds = config?.checkoutSettings.remoteCheckoutProviders ?? [];
+
+    useEffect(() => {
+        const changeState = setTimeout(()=>{
+            if(!isInitializingCustomer()) {
+                setIsLoading(false);
```
davidchin, line 45: What does this `useEffect` block do?

packages/core/src/scss/components/checkout/customer/_customer.scss
```diff
@@ -92,19 +92,156 @@
...
+        > div {
+            height: $wallet-button-height;
+            margin-right: 0;
+        }
+
+        .AmazonPayContainer,
+        > #applepayCheckoutButton button,
+        button.gpay-button,
+        > #braintreepaypalcreditCheckoutButton,
+        > #braintreepaypalCheckoutButton {
```
davidchin, line 137: Avoid nesting and overly specific selectors if possible, see https://github.com/bigcommerce/sass-style-guide#nesting and https://github.com/bigcommerce/sass-style-guide#specificity

packages/core/src/app/customer/CheckoutButtonContainer.tsx
bc-peng: To use `CheckoutButtonListV2`, we must complete the locale work which we discussed in a previous meeting. That's why we are still using the old ` CheckoutButtonList` now.

davidchin: Ah thanks Peng, I missed the previous meeting 😅

packages/core/src/app/customer/CheckoutButtonContainer.tsx
bc-peng: `isInitializingCustomer()` toggles between `true` and `false` during loading, forcing the loading skeleton to render whenever it is `false`, therefore I use this block to hide the skeleon when the loading is actully done.

I also tried other state selectors such as `isPending`, `isLoadingMethod`. They all act the same way.

packages/core/src/app/customer/CheckoutButtonContainer.tsx
animesh1987: Yes, we had a ticket to switch the buttons at top to V2. Then realised we need to complete the locale work in order to enable it. We have moved it as part of the react project for now and should be able to handle it then.

packages/core/src/app/customer/CheckoutButtonContainer.tsx
animesh1987: Do we know the reason why it toggles between the values? Is that because it processes requests for each button/method id?

packages/core/src/app/checkout/mapToCheckoutProps.ts
```diff
@@ -31,6 +37,8 @@ export default function mapToCheckoutProps({
         },
     );
 
+    const walletButtonsOnTopFlag = !!features['CHECKOUT-7222.checkout_settings_styling_section'] && walletButtonsOnTop;
```
animesh1987, line 40: Shall we change it to `Boolean(features['CHECKOUT-7222.checkout_settings_styling_section'])`

packages/core/src/app/customer/CheckoutButtonContainer.tsx
```diff
@@ -27,22 +37,63 @@ const CheckoutButtonContainer: FunctionComponent<CheckoutButtonsListOnTopProps &
...
+            }
+        },100);
+
+        return () => {
+            clearTimeout(changeState);
+            setIsLoading(true);
```
animesh1987, line 51: Why do we need to set the `isLoading` state to true when the component in unmounted?

packages/core/src/app/customer/CheckoutButtonContainer.tsx
```diff
@@ -27,22 +37,63 @@ const CheckoutButtonContainer: FunctionComponent<CheckoutButtonsListOnTopProps &
...
+    }, [isInitializingCustomer]);
+
+    if (!config || methodIds.length === 0) {
+        return null;
+    }
+
+    const supportedMethodIds = sortMethodIds(filterUnsupportedMethodIds(methodIds));
+    const buttonsCount = supportedMethodIds.length;
+
+    if (buttonsCount < 1) {
```
animesh1987, line 62:
```suggestion
    if (buttonsCount === 0) {
```

packages/core/src/app/customer/CheckoutButtonContainer.tsx
bc-peng: Restore the state to `loading` whenever `isInitializingCustomer` is changing.
But since I put a delay above, this seems unnecessary.

packages/core/src/app/customer/CheckoutButtonList.tsx
animesh1987: Strange, this should have been picked up.

packages/ui/src/loading/LoadingNotification.tsx
```diff
@@ -1,3 +1,4 @@
+// TODO: Remove this component after May, 2023
```
animesh1987, line 1: Why May?

bc-peng: I'm not sure whether any BCapp functional tests or out side apps use it, so leave it in place for an easy rollback. 
Thought two months should be enough. 😂
Will run functional tests as well when I have addressed all comments.

packages/core/src/app/customer/CheckoutButtonContainer.tsx
davidchin: Right, it sounds like we're missing something like `isInitialized` (https://github.com/bigcommerce/checkout-sdk-js/blob/e9d7830aae32d48fbaf825c232a7a99b0099767d/packages/core/src/checkout-buttons/checkout-button-selector.ts#L12). We have it for checkout buttons, but not for customer step. Could we consider adding one instead of using a hardcoded timeout? As far as I can tell, the timeout solution won't work if the initialisation takes longer than 100ms. We need to actually set a flag somewhere once the initialisation is complete.

packages/core/src/scss/components/checkout/checkoutRemote/_checkoutRemote.scss
```diff
@@ -15,22 +15,23 @@
...
-    }
+#applepayCheckoutButton button {
+    background-color: #000;
+    background-image: -webkit-named-image(apple-pay-logo-white);
+    background-position: 50% 50%;
+    background-repeat: no-repeat;
+    background-size: 100% 60%;
+    border-radius: $global-radius;
+    height: $wallet-button-height;
+    padding: spacing('single');
```
bc-peng, line 34: Removed unnecessary nesting and adjusted Apple Pay button size in the customer step.
image

packages/ui/src/loading/LoadingNotification.tsx
bc-peng: I removed `LoadingNotification.tsx` now.

packages/core/src/app/customer/CheckoutButtonContainer.tsx
animesh1987: I think that's a reasonable approach, and makes the overall solution more robust than relying on timeouts cc @bc-peng

packages/core/src/app/customer/CheckoutButtonContainer.tsx
```diff
@@ -1,48 +1,93 @@
...
+            getInitializeCustomerError,
+        }
     } = checkoutState;
     const config = getConfig();
+    const [methodIds, setMethodIds] = useState(filterUnsupportedMethodIds(config?.checkoutSettings.remoteCheckoutProviders ?? []));
+    const isLoading = methodIds.filter(
+        (methodId) => Boolean(getInitializeCustomerError(methodId)) || isInitializedCustomer(methodId)
+    ).length !== methodIds.length;
+
+    useEffect(() => {
```
davidchin, line 47: I think you can lift up most up of the filtering / mapping to the `withCheckout` mapper function. That way you don't need to use `useEffect` and `setState` inside `CheckoutButtonContainer` component.

```tsx
export default withCheckout(({
    data: {
        getConfig,
    },
    statuses: {
        isInitializedCustomer,
    },
    errors: {
        getInitializeCustomerError,
    }
},
{
    checkEmbeddedSupport,
    onUnhandledError,
}) => {
    const availableMethodIds = filterUnsupportedMethodIds(getConfig()?.checkoutSettings.remoteCheckoutProviders);
    const methodIds = availableMethodIds.filter(isInitializedCustomer);
    const isLoading = availableMethodIds.filter(
        (methodId) => getInitializeCustomerError(methodId) || isInitializedCustomer(methodId)
    ).length !== methodIds.length;

    return {
        checkEmbeddedSupport,
        checkoutService,
        isLoading,
        onUnhandledError,
        methodIds,
    };
})(CheckoutButtonContainer);
```

bc-peng: Thanks DC, I have updated the mapper function like you suggested.

Isn't this because we don't want more state management when we already have it in the SDK?

davidchin: Thanks @bc-peng. Yes, generally speaking, we want to [avoid deriving states from props](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#when-to-use-derived-state). In this case, the parent component (checkout HOC) controls the initialised / loading state of the buttons. So its child component (button container) can simply accept the data as props. Otherwise, if the button container also calls `setState`, there are be two places managing the initialised / loading state of the buttons.

davidchin (approved):

animesh1987 (approved):

PR #1210, opened 2023-02-23
Update translations

bc-svc-local:
ping @bigcommerce/localization


funivan:
see https://github.com/bigcommerce/checkout-js/pull/1215

PR #1211, opened 2023-02-23
CHECKOUT-000 Change commit validation rule

funivan:
## What?
Change commit scopes

## Why?
`other` is a standard scope in most BC Respos. 

## Testing / Proof
N/A

@bigcommerce/checkout


animesh1987 (approved):
Thanks for updating this. Just need to update the commit message.

PR #1223, opened 2023-03-14
feat(checkout): CHECKOUT-3430 Display inclusive tax below subtotal

animesh1987:
## What?
If tax is set as inclusive, display tax total under total cart price.

## Why?
Order Summary shows the Total price of the order as (Subtotal + Shipping). Although it might be clear to the merchant about inclusive tax, however for the customer it's not apparent. The customer can evaluate the price of the cart as (Subtotal + Shipping + tax) which is not true.

## Testing / Proof
- Circle

## Total Inclusive of tax
Screen Shot 2023-03-14 at 5 39 35 pm

Screen Shot 2023-03-14 at 5 39 49 pm

## Total exclusive of tax
Screen Shot 2023-03-14 at 5 40 46 pm

Screen Shot 2023-03-14 at 5 41 02 pm

@bigcommerce/checkout @theromulans

### DIFF
--- a/packages/core/src/app/locale/translations/en.json
+++ b/packages/core/src/app/locale/translations/en.json
@@ -461,6 +461,9 @@
             "label": "Yes, I agree with the <a href=\"{url}\" target=\"_blank\">privacy policy</a>.",
             "heading": "Privacy Policy"
         },
+        "tax": {
+            "inclusive_label": "Tax Included in Total:"
+        },
         "terms_and_conditions": {
             "agreement_required_error": "Please agree to the terms and conditions",
             "agreement_text": "Yes, I agree with the above terms and conditions.",
--- a/packages/core/src/app/order/OrderSummary.tsx
+++ b/packages/core/src/app/order/OrderSummary.tsx
@@ -42,7 +42,7 @@ const OrderSummary: FunctionComponent<OrderSummaryProps & OrderSummarySubtotalsP
             </OrderSummarySection>
 
             <OrderSummarySection>
-                <OrderSummarySubtotals isTaxIncluded taxes={taxes} {...orderSummarySubtotalsProps} />
+                <OrderSummarySubtotals isTaxIncluded={isTaxIncluded} taxes={taxes} {...orderSummarySubtotalsProps} />
                 {additionalLineItems}
             </OrderSummarySection>
 
@@ -54,16 +54,16 @@ const OrderSummary: FunctionComponent<OrderSummaryProps & OrderSummarySubtotalsP
                 />
             </OrderSummarySection>
 
-            <OrderSummarySection>
+            {isTaxIncluded && <OrderSummarySection>
                 <h5
-                    className="cart-section-heading optimizedCheckout-contentPrimary"
+                    className="cart-taxItem cart-taxItem--subtotal optimizedCheckout-contentPrimary"
                     data-test="tax-text"
                 >
                     <TranslatedString
                         id="tax.inclusive_label"
                     />
                 </h5>
-                {isTaxIncluded && (taxes || []).map((tax, index) => (
+                {(taxes || []).map((tax, index) => (
                     <OrderSummaryPrice
                         amount={tax.amount}
                         key={index}
@@ -71,7 +71,7 @@ const OrderSummary: FunctionComponent<OrderSummaryProps & OrderSummarySubtotalsP
                         testId="cart-taxes"
                     />
                 ))}
-            </OrderSummarySection>
+            </OrderSummarySection>}
         </article>
     );
 };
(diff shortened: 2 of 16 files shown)


theromulans:
The screenshots seem to only show one of the two display strategies - is that right? I was hoping to see an example of both. 

Ideally in the rate breakdown view there'd be multiple taxes too, just to cover our bases there 🙏 

Something along the lines of...

(Display setting: total only)
```
Tax included in total: $10
```

(Display setting: rate breakdown)
```
Tax included in total:
Tax Rate 1: $5
Tax Rate 2: $5
```

packages/core/src/app/locale/translations/en.json
```diff
@@ -461,6 +461,9 @@
             "label": "Yes, I agree with the <a href=\"{url}\" target=\"_blank\">privacy policy</a>.",
             "heading": "Privacy Policy"
         },
+        "tax": {
```
theromulans, line 464: 🍻 I note we have `cart.taxes_text` already - would it make sense to have this present there also?

e.g. `cart.taxes_inclusive_text`?

...or are you perhaps thinking that will move to `exclusive_label` here?

packages/core/src/app/locale/translations/en.json
```diff
@@ -461,6 +461,9 @@
...
+            "inclusive_label": "Tax Included in Total:"
```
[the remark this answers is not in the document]
animesh1987: Updated

packages/core/src/app/locale/translations/en.json
animesh1987: I was thinking we can have a separate section of the `inclusive_text`, maybe makes sense to move it as `cart.taxes_inclusive_text` if that's the only bit of text for that section 🤔

animesh1987:
> The screenshots seem to only show one of the two display strategies - is that right? I was hoping to see an example of both.
> 
> Ideally in the rate breakdown view there'd be multiple taxes too, just to cover our bases there 🙏
> 
> Something along the lines of...
> 
> (Display setting: total only)
> 
> ```
> Tax included in total: $10
> ```
> 
> (Display setting: rate breakdown)
> 
> ```
> Tax included in total:
> Tax Rate 1: $5
> Tax Rate 2: $5
> ```

Yes, the screenshots only display the breakdown strategy as discussed offline earlier. I didn't have my store configured with multiple tax line items but should be similar to 

```
Tax included in total:
Tax Rate 1: $5
Tax Rate 2: $5
```

theromulans:
Cool, if you're confident multiple rates will display as expected 👍 

In that case the only item remaining is that it'd be great to see a screenshot of the "Total only" view in addition to the current "Rate breakdown" view 🙏

animesh1987:
> Cool, if you're confident multiple rates will display as expected 👍
> 
> In that case the only item remaining is that it'd be great to see a screenshot of the "Total only" view in addition to the current "Rate breakdown" view 🙏

Screen Shot 2023-03-09 at 10 34 17 am

theromulans (approved):
Wonderful!

packages/core/src/app/order/OrderSummary.tsx
```diff
@@ -48,6 +53,25 @@ const OrderSummary: FunctionComponent<OrderSummaryProps & OrderSummarySubtotalsP
...
+            {isTaxIncluded && <OrderSummarySection>
+                <h5
+                    className="cart-taxItem cart-taxItem--subtotal optimizedCheckout-contentPrimary"
+                    data-test="tax-text"
+                >
+                    <TranslatedString
+                        id="tax.inclusive_label"
+                    />
+                </h5>
+                {(taxes || []).map((tax, index) => (
```
bc-peng, line 66: If `taxes` is empty, do we still need to display the whole  `<OrderSummarySection>`?

packages/core/src/app/order/OrderSummaryModal.tsx
```diff
@@ -64,6 +67,24 @@ const OrderSummaryModal: FunctionComponent<
...
+        {isTaxIncluded && <OrderSummarySection>
+                <h5
+                    className="cart-taxItem cart-taxItem--subtotal optimizedCheckout-contentPrimary"
+                    data-test="tax-text"
+                >
+                    <TranslatedString
+                        id="tax.inclusive_label"
+                    />
+                </h5>
+                {(taxes || []).map((tax, index) => (
```
bc-peng, line 79: Same concern here.

packages/core/src/scss/components/checkout/checkoutCart/_checkoutCart.scss
```diff
@@ -140,3 +140,20 @@
...
+// -----------------------------------------------------------------------------
+
+.cart-taxItem {
+    display: table;
+    margin-bottom: spacing("half");
+    width: 100%;
+}
+
+.cart-taxItem--subtotal {
+    font-weight: 600;
```
bc-peng, line 158:
```suggestion
    font-weight: fontWeight("semibold");
```

packages/core/src/app/order/OrderSummary.tsx
animesh1987: Good point, just didn't imagine a world without tax 😅

bc-peng (approved):
Nice work👏

PR #1236, opened 2023-03-30
feat(checkout): CHECKOUT-0000 Hide wallet buttons during payment step

bc-peng:
## What?
Hide wallet buttons when a shopper enters the payment step.

## Why?
`paypalcommerce` does not support simultaneously display wallet buttons and payment methods.

## Testing / Proof
- CI.

@bigcommerce/checkout

### DIFF
--- a/packages/core/src/app/customer/CheckoutButtonContainer.tsx
+++ b/packages/core/src/app/customer/CheckoutButtonContainer.tsx
@@ -18,6 +18,7 @@ interface CheckoutButtonContainerProps {
 interface WithCheckoutCheckoutButtonContainerProps{
     availableMethodIds: string[];
     isLoading: boolean;
+    isPaypalCommerce: boolean;
     initializedMethodIds: string[];
     deinitialize(options: CustomerRequestOptions): void;
     initialize(options: CustomerInitializeOptions): void;
@@ -42,6 +43,7 @@ const CheckoutButtonContainer: FunctionComponent<CheckoutButtonContainerProps & 
         checkEmbeddedSupport,
         deinitialize,
         isLoading,
+        isPaypalCommerce,
         isPaymentStepActive,
         initialize,
         initializedMethodIds,
@@ -55,9 +57,6 @@ const CheckoutButtonContainer: FunctionComponent<CheckoutButtonContainerProps & 
     } catch (error) {
         return null;
     }
-
-    const paypalCommerceIds = ['paypalcommerce', 'paypalcommercecredit', 'paypalcommercevenmo'];
-    const isPaypalCommerce = methodIds.some(id => paypalCommerceIds.includes(id));
 
     if (isPaypalCommerce && isPaymentStepActive) {
         return null;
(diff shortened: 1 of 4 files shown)


packages/core/src/app/customer/CheckoutButtonContainer.tsx
```diff
@@ -54,8 +56,16 @@ const CheckoutButtonContainer: FunctionComponent<CheckoutButtonContainerProps &
         return null;
     }
 
+    const isPaypalCommerce = methodIds.includes('paypalcommerce');
```
serhii-tkachenko, line 59: It might be better to add extra check for 'paypalcommercecredit' and 'paypalcommercevenmo' method ids too.

packages/core/src/app/customer/CheckoutButtonContainer.tsx
```diff
@@ -54,8 +56,16 @@ const CheckoutButtonContainer: FunctionComponent<CheckoutButtonContainerProps &
...
+
+    if (isPaypalCommerce && isPaymentStepActive) {
+        return null;
+    }
+
     return (
-        <div className='checkout-button-container'>
+        <div className='checkout-button-container'
+             style={ isPaymentStepActive ? { position: 'absolute', left: '100%', top: '-100%' } : undefined }
```
serhii-tkachenko, line 67: Is it possible to use { position: 'absolute', visibility: 'hidden' } in this case? Because we might have some issues in mobile Safari, when the user quickly scrolling up and down.

packages/core/src/app/customer/CheckoutButtonContainer.tsx
bc-peng: Thanks. Done.

packages/core/src/app/customer/CheckoutButtonContainer.tsx
bc-peng: My concern is changing `visibility` may cause some buttons to react. 
I did notice an issue on mobile devices, white space on the right side of the page when the buttons are hidden, hence I set the left to `0` so that the page size should stay the same.

packages/core/src/app/customer/CheckoutButtonContainer.tsx
```diff
@@ -54,8 +56,17 @@ const CheckoutButtonContainer: FunctionComponent<CheckoutButtonContainerProps &
         return null;
     }
 
+    const paypalCommerceIds = ['paypalcommerce', 'paypalcommercecredit', 'paypalcommercevenmo'];
+    const isPaypalCommerce = methodIds.some(id => paypalCommerceIds.includes(id));
```
animesh1987, line 60: Future optimisation, we should be able to move these methods out of the component body. So that they are not re-created on every re render.

animesh1987 (approved):

packages/core/src/app/customer/CheckoutButtonContainer.tsx
bc-peng: I have moved it to the wrapper/map function.

PR #1239, opened 2023-04-04
feat(checkout): CHECKOUT-0000 Measure checkout performance improvement

bc-peng:
_* This PR is intended solely for demonstration purposes and will not be merged._

Previously, shoppers would have to wait for the wallet buttons to load before being able to use the customer step. This delay was noticeable because the loading of a wallet button frequently involves multiple network requests. 

We separated the two processes as part of the project implementation of moving wallet buttons to the top of the checkout page.

This change effectively eliminated the delay and the updated checkout page now appears instantly ready for use to shoppers, streamlining the process even further. We also removed the loading skeleton of the customer step because it was no longer required.

In our test, we achieved a 52% reduction in the time it takes for `playwright` to input the customer email address since the window load event, corresponding to a 0.7 second decrement. That's 2 times faster than before.

The test setup has the following specifications:
- `checkout-js` version 1.264 is the first version that includes the customer step loading skeleton.
- `checkout-js` version 1.304 is the current lastest version.
- All `checkout-js` code is served through the store's WebDav storage.
- The test is performed 20 times for each version, utilizing the command `PORT=5003 MODE=RECORD STOREURL=https://me8tmnd9e9.myprivaterelay.com/ npx playwright test measureloadingtime --repeat-each=20`.
- Raw data generated by the code of this PR:

   > checkout-js version 1.264
   > Time to DOMContentLoaded: 783ms
   > Time to Load: 1298ms
   > Time to Done: 2574ms
   > 
   > checkout-js version 1.304
   > Time to DOMContentLoaded: 781ms
   > Time to Load: 1164ms
   > Time to Done: 1706ms

## Before:

https://user-images.githubusercontent.com/88361607/229677069-ce321716-6267-4414-b1d8-67a4996dea1d.mov

## After:

https://user-images.githubusercontent.com/88361607/229677098-96acd7c1-cc14-44fd-ac21-25b4e496a712.mov

### DIFF
--- a/packages/core/e2e/MeasureLoadingTime.spec.ts
+++ b/packages/core/e2e/MeasureLoadingTime.spec.ts
@@ -1,0 +1,64 @@
+import {
+    CustomerStepPreset,
+    test,
+} from '@bigcommerce/checkout/payment-integration-test-framework';
+
+interface timings {
+    goto:number;
+    doc:number;
+    load:number;
+    done:number
+}
+
+const data:timings[] = [];
+
+function calculateTimings(timings:timings) {
+    let timeToDoc = 0;
+    let timeToLoad = 0;
+    let timeToDone = 0;
+
+    data.push(timings);
+
+    for (const item of data) {
+        timeToDoc =+ item.doc - item.goto;
+        timeToLoad =+ item.load - item.goto;
+        timeToDone =+ item.done - item.goto;
+    }
+
+    console.log(`Time to DOMContentLoaded: ${timeToDoc/data.length}ms`);
+    console.log(`Time to Load: ${timeToLoad/data.length}ms`);
+    console.log(`Time to Done: ${timeToDone/data.length}ms`);
+}
+
+test.describe('Sample Test Group', () => {
+    test('Bigpay Test Payment Provider is working', async ({ checkout, page }) => {
+        const timings:timings = {
+            goto:0,
+            doc:0,
+            load:0,
+            done:0
+        };
+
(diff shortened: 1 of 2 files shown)


packages/core/e2e/MeasureLoadingTime.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+            load:0,
+            done:0
+        };
+
+        page.on('domcontentloaded', () => {
+            timings.doc = Date.now();
+        });
+        page.on('load', () => {
+            timings.load = Date.now();
+        });
```
bc-peng, line 47: https://playwright.dev/docs/api/class-page#page-event-load
Emitted when the JavaScript [load](https://developer.mozilla.org/en-US/docs/Web/Events/load) event is dispatched.

packages/core/e2e/MeasureLoadingTime.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+        const timings:timings = {
+            goto:0,
+            doc:0,
+            load:0,
+            done:0
+        };
+
+        page.on('domcontentloaded', () => {
+            timings.doc = Date.now();
+        });
```
bc-peng, line 44: https://playwright.dev/docs/api/class-page#page-event-dom-content-loaded
Emitted when the JavaScript [DOMContentLoaded](https://developer.mozilla.org/en-US/docs/Web/Events/DOMContentLoaded) event is dispatched.

PR #1250, opened 2023-04-20
feat(cart): CHECKOUT-7403 Update layout and UX for cart summary modal in mobile screens

animesh1987:
## What?
Update layout for cart summary modal in mobile screens
- Add quick summary in header
- Add a `Return to checkout` call to action button
- Swap position of Edit cart link and close icon

## Why?
There has been some feedback from users around poor UX for modal for cart summary in checkout page. Hence this is an effort to refresh the layout. Also add explicit call to action buttons for users to return to checkout.

## Testing / Proof
- Before / With experiment off
Screen Shot 2023-04-20 at 11 43 43 pm

- After when store currency is same as shopper currency
Screen Shot 2023-04-24 at 12 25 37 pm

- After when store currency is different from shopper currency
Screen Shot 2023-04-24 at 12 26 29 pm

- After with more than 3 items in cart
Screen Shot 2023-04-24 at 12 28 19 pm

- The changes are still gated by an experiment.

@bigcommerce/checkout


packages/core/src/app/order/OrderModalSummarySubheader.spec.tsx
```diff
@@ -0,0 +1,89 @@
...
+        const currencyAmount = orderModalSummarySubHeader.find(ShopperCurrency);
+
+        expect(orderModalSummarySubHeader.find(TranslatedString).props()).toMatchObject({
+            id: 'cart.item',
+        });
+
+        expect(currencyAmount.text()).toBe('$112.00');
+    });
+
+    it('when shopper has same currency as the store and 1 item, text id passed is cart.item', () => {
```
bc-peng, line 47:
```suggestion
    it('when shopper has same currency as the store and multiple items, text id passed is cart.items', () => {
```

packages/core/src/scss/components/checkout/cartModal/_cartModal.scss
```diff
@@ -37,3 +37,28 @@
...
+    .cart-modal-link {
+        text-align: left;
+    }
+
+    .cart-modal-close {
+        text-align: right;
+    }
+
+    .cart-heading-subheader {
+        font-weight: 500;
```
bc-peng, line 60:
```suggestion
        font-weight: fontWeight("medium");
```

bc-peng (approved):
Nice job.👏👏

PR #1272, opened 2023-05-11
fix(shipping): CHECKOUT-7338 Add available options include if custom fields change

animesh1987:
## What?
Add available options include to consignment call when custom fields are changed

## Why?
We do refresh shipping rates when custom form fields are changed in the backend, so it does make sense to refresh the rates in the client as well. 

The driver for this is some partner integrations who are using CFF's for adding things like insurance etc.

## Testing / Proof
https://github.com/bigcommerce/checkout-js/assets/7134802/1589c766-f0e8-4d33-b187-f5fe6372b9f1

@bigcommerce/checkout

### DIFF
--- a/packages/core/src/app/shipping/SingleShippingForm.tsx
+++ b/packages/core/src/app/shipping/SingleShippingForm.tsx
@@ -12,7 +12,7 @@ import {
     ShippingRequestOptions,
 } from '@bigcommerce/checkout-sdk';
 import { FormikProps, withFormik } from 'formik';
-import { debounce, noop } from 'lodash';
+import { debounce, isEqual, noop } from 'lodash';
 import React, { PureComponent, ReactNode } from 'react';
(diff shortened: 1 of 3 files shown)


packages/core/src/app/shipping/SingleShippingForm.tsx
```diff
@@ -238,6 +238,13 @@ class SingleShippingForm extends PureComponent<
 
         const updatedShippingAddress = addressForm && mapAddressFromFormValues(addressForm);
 
+        if (Array.isArray(shippingAddress?.customFields)) {
+            includeShippingOptions = !isEqual(
+                shippingAddress?.customFields,
+                updatedShippingAddress?.customFields
+            ) || includeShippingOptions;
+        }
```
bc-peng, line 246: LGTM.👍

PR #1279, opened 2023-05-17
fix(checkout): PI-48 Place Order button is disabled when customer with stored credit card deletes and re-adds card

bc-ania:
## What?
I removed the behavior to automatically disable the submit button when the form loads

## Why?
Place Order button on checkout is disabled when a customer with a stored credit card deletes, and re-adds card on checkout. The issue also occurs if deleting the stored card, and trying to use a different card with Adyen(non-stored).

image

[video how it worked before the changes](https://github.com/bigcommerce/checkout-js/assets/130754705/b7ac8249-2e1e-44d9-becf-434c81a5fcd4)

## Testing / Proof
After the changes:

[Here's the behavior now](https://github.com/bigcommerce/checkout-js/assets/130754705/1a4d05df-78cb-4dd2-a774-54723b9849d6)

@bigcommerce/checkout

### DIFF
--- a/packages/core/src/app/payment/paymentMethod/AdyenV2CardValidation.spec.tsx
+++ b/packages/core/src/app/payment/paymentMethod/AdyenV2CardValidation.spec.tsx
@@ -170,24 +170,5 @@ describe('AdyenV2CardValidation', () => {
             expect(field).toHaveLength(1);
         });
 
-        it('should disable submit payment button initially', () => {
-            paymentContext = {
-                disableSubmit: jest.fn(),
-                setSubmit: jest.fn(),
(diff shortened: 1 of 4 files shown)


vitalii-koshovyi (changes requested):
Could you please check it with AdyenV3 as well. I believe it has the same problem.

vitalii-koshovyi:
Please add some screenshots or video that will show changes .

bc-peng (approved):
LGTM 👍

animesh1987:
Hey @bc-ania can you please adjust the PR description before merging. Thanks.

PR #1288, opened 2023-05-22
chore(payment): CHECKOUT-7289 Use v2 button component to render buttons on top

animesh1987:
## What?
- Use v2 button component to render buttons on top
- Add a way to resolve V1 component if no button component is resolved

## Why?
In order to switch our approach to use package based wallet button component, we need to switch 
over to utilise package resolved button components.

## Testing / Proof
- Screenshots
- E2E

## With buttons on top
Screenshot 2023-05-23 at 9 54 07 am

## With buttons in customer step
Screenshot 2023-05-23 at 9 54 21 am

@bigcommerce/checkout


packages/core/auto-export.config.json
```diff
@@ -5,6 +5,12 @@
             "ignorePackages": ["test-utils"],
             "outputPath": "packages/core/src/app/generated/paymentIntegrations/index.ts",
             "memberPattern": ".+PaymentMethod$"
+        },
+        {
+            "inputPath": "packages/**/src/index.ts",
+            "ignorePackages": ["test-utils", "checkout-button-integration"],
```
bc-peng, line 11: Think we can also add them to both entries.
```
["analytics", "core", "instrument-utils", "locale", "payment-integration-api", "payment-integration-test-framework", "ui", "workspace-tools"]
```

packages/core/src/app/customer/WalletButtonV1Resolver.tsx
```diff
@@ -0,0 +1,67 @@
...
+                    containerId={`${methodId}CheckoutButton`}
+                    key={methodId}
+                    methodId={methodId}
+                    onError={onError}
+                    {...rest}
+                />
+            );
+    }
+
+    return <CheckoutButton
```
bc-peng, line 57: Would it be better to put this as the default action in the above `switch`?

animesh1987:
I did that before but then I had to move to a pattern to return `null` at the end.

packages/core/src/app/customer/CheckoutButtonContainer.tsx
```diff
@@ -113,7 +105,7 @@ function mapToCheckoutButtonContainerProps({
     checkoutService,
 }: CheckoutContextProps): WithCheckoutCheckoutButtonContainerProps | null {
     const config = getConfig();
-    const availableMethodIds = filterUnsupportedMethodIds(config?.checkoutSettings.remoteCheckoutProviders ?? []);
+    const availableMethodIds = getSupportedMethodIds(config?.checkoutSettings.remoteCheckoutProviders ?? []);
```
davidchin, line 108: Can we limit this call to only V1 buttons? V2 button shouldn't need it at all, as `remoteCheckoutProviders` that don't have a button won't resolve with anything. The reason we want to limit to V1 only is because we don't want to encourage teams to add more IDs to the list of "supported methods".

animesh1987: Yes that's correct in a future state when all buttons have been migrated to their own packages, but right we need to pass a list of methodIds to V2 button component to maintain parity and deprecate V1 list.

Once all buttons have been migrated to their respective packages we can get rid of the methods that need to be passed to the component.

davidchin: I meant move the call so that the result of the call wouldn't affect V2 buttons.

I notice that you need the count of `availableMethodIds` in order to set the correct CSS class for the container (e.g.: `checkout-buttons--1`). This means we will still need to modify `SUPPORTED_METHODS` in the `core` package every time we add a new V2 button.

To solve this problem, I think we need to resolve the V2/V1 buttons earlier. For example, in `CheckoutButtonContainer`, we could have:

```tsx
const resolvedCheckoutButtons = remoteCheckoutProviders.reduce((buttons, methodId) => {
    const resolvedCheckoutButtonV2 = resolveCheckoutButtonV2({ id: methodId });

    if (resolvedCheckoutButton) {
        return [...buttons, resolvedCheckoutButtonV2];
    }

    const resolvedCheckoutButtonV1 = resolveCheckoutButtonV1({ id: methodId });

    if (resolvedCheckoutButtonV1) {
        return [...buttons, resolvedCheckoutButtonV2];
    }

    return buttons;
}, []);
```

And in `resolveCheckoutButtonV1`, we could check to see if `methodId` is in the list of `SUPPORTED_METHODS`.

That way, `resolvedCheckoutButtons.length` can give you the correct count for the container while the buttons are loading. And `SUPPORTED_METHODS` only affects V1 button.

animesh1987: Right now we have the `SUPPORTED_METHODS` in other places as well. So I suppose we can move the buttons V2 component to accept a list of resolved buttons(combination of V1 and V2) as props. And they can be rendered accordingly.

animesh1987: Wanted to check do we want to remove V2 List component, if we are resolving buttons in Container itself. I say so since we might need to pass more information to V2 list in this case such as is the button a V1 or V2 component since they have different interfaces. So might need to pass an object with 
`[ { type: 'V1', methodId: 'applepay', component: V1Component }, { type: 'V2', methodId: 'googlepay', component: V2Component } ]`?

packages/core/src/app/customer/CheckoutButtonContainer.tsx
```diff
@@ -72,24 +100,17 @@ const CheckoutButtonContainer: FunctionComponent<CheckoutButtonContainerProps &
...
-                'checkout-buttons--3': methodIds.length === 3,
-                'checkout-buttons--4': methodIds.length === 4,
-                'checkout-buttons--5': methodIds.length === 5,
-                'checkout-buttons--n': methodIds.length > 5,
+                'checkout-buttons--1': availableMethodIds.length === 1,
+                'checkout-buttons--2': availableMethodIds.length === 2,
+                'checkout-buttons--3': availableMethodIds.length === 3,
+                'checkout-buttons--4': availableMethodIds.length === 4,
+                'checkout-buttons--5': availableMethodIds.length === 5,
+                'checkout-buttons--n': availableMethodIds.length > 5,
```
bc-peng, line 108:
```suggestion
                'checkout-buttons--1': methodIds.length === 1,
                'checkout-buttons--2': methodIds.length === 2,
                'checkout-buttons--3': methodIds.length === 3,
                'checkout-buttons--4': methodIds.length === 4,
                'checkout-buttons--5': methodIds.length === 5,
                'checkout-buttons--n': methodIds.length > 5,
```

packages/core/src/app/customer/CheckoutButtonContainer.tsx
```diff
@@ -72,24 +100,17 @@ const CheckoutButtonContainer: FunctionComponent<CheckoutButtonContainerProps &
...
-                        checkEmbeddedSupport={checkEmbeddedSupport}
-                        deinitialize={deinitialize}
-                        hideText={true}
-                        initialize={initialize}
-                        isInitializing={isLoading}
-                        isShowingWalletButtonsOnTop={true}
-                        methodIds={sortMethodIds(methodIds)}
-                        onError={onUnhandledError}
-                    />
+                <WalletButtonsContainerSkeleton buttonsCount={availableMethodIds.length} isLoading={isLoading}>
```
bc-peng, line 110:
```suggestion
                <WalletButtonsContainerSkeleton buttonsCount={methodIds.length} isLoading={isLoading}>
```

bc-peng (approved):
Great!!! Nicely done.

vmparra (approved):
LGTM 👍🏻

PR #1294, opened 2023-05-24
fix(checkout): PI-147 remove "Pay over time" text from the Clearpay

vitalii-koshovyi:
## What?
Text "Pay over time" removed from the Clearpay

## Why?
In the BigCommerce payment integration we currently use the copy “Pay over time” next to the Clearpay logo (see image below). This is no longer approved language for Clearpay UK merchants and needs to be removed before the FCA June 1st deadline. No need to replace the language with something else

## Testing / Proof
image

image

@bigcommerce/checkout


serhii-tkachenko (approved):
LGTM! 👍

bc-nick (approved):

animesh1987 (approved):

PR #1295, opened 2023-05-24
refactor(payment): PI-147 PaymentMethodTitle moved spec to RTL

vitalii-koshovyi:
## What?
Refactored PaymentMethodTitle from enzyme to the RTL

## Why?
Due to the recent work with this component

## Testing / Proof
units

@bigcommerce/checkout


packages/core/src/app/payment/paymentMethod/PaymentMethodTitle.test.tsx
```diff
@@ -8,14 +9,15 @@ import { createLocaleContext, LocaleContext, LocaleContextType } from '@bigcomme
...
 import { getStoreConfig } from '../../config/config.mock';
-import { CreditCardIconList } from '../creditCard';
 import { getPaymentMethod } from '../payment-methods.mock';
 
 import getPaymentMethodName from './getPaymentMethodName';
 import PaymentMethodId from './PaymentMethodId';
 import PaymentMethodTitle, { PaymentMethodTitleProps } from './PaymentMethodTitle';
 import PaymentMethodType from './PaymentMethodType';
 
+configure({testIdAttribute: 'data-test'})
```
vitalii-koshovyi, line 19: @bc-peng @animesh1987 I would recommend do this change on the global level, since we have data-test as test-id in our aplication.

bc-peng: Sure, please put it on the global level.
However, it is not a offical recommended approach to use test IDs, we should always aim to use `getByRole` frist.
https://testing-library.com/docs/queries/about/#priority

vitalii-koshovyi: @bc-peng I don't know where I should put it. Could you help me with the configuration?
Whole test is base on the data-test attribute so I decided to just make this refactoring with shortest part.

vitalii-koshovyi: @bc-peng updated with roles

packages/core/src/app/payment/paymentMethod/PaymentMethodTitle.test.tsx
```diff
@@ -8,14 +9,14 @@ import { createLocaleContext, LocaleContext, LocaleContextType } from '@bigcomme
...
 import { getStoreConfig } from '../../config/config.mock';
-import { CreditCardIconList } from '../creditCard';
 import { getPaymentMethod } from '../payment-methods.mock';
 
 import getPaymentMethodName from './getPaymentMethodName';
 import PaymentMethodId from './PaymentMethodId';
 import PaymentMethodTitle, { PaymentMethodTitleProps } from './PaymentMethodTitle';
 import PaymentMethodType from './PaymentMethodType';
 
+
```
bc-peng, line 19:
```suggestion
```

packages/core/src/app/payment/paymentMethod/PaymentMethodTitle.test.tsx
```diff
@@ -91,7 +93,9 @@ describe('PaymentMethodTitle', () => {
             />,
         );
 
-        expect(component.find('[data-test="payment-method-name"]').text()).toEqual(
+        const component = screen.getByRole('heading');
+
+        expect(component?.textContent).toEqual(
             defaultProps.method.config.displayName,
         );
```
bc-peng, line 100: Can I suggest using `toBeInTheDocument` instead of `toEqual` whenever possible for asserting text content in RTL.
```
expect(screen.getByRole('heading', { name: defaultProps.method.config.displayName })).toBeInTheDocument();
```

packages/core/src/app/payment/paymentMethod/PaymentMethodTitle.test.tsx
```diff
@@ -146,14 +153,17 @@ describe('PaymentMethodTitle', () => {
                 />,
             );
 
-            expect(component.find('[data-test="payment-method-logo"]').prop('src')).toBe(
+            const component = screen.getByRole('img');
+
+            expect(component?.getAttribute('src')).toEqual(
                 `${config.cdnPath}${LOGO_PATHS[method.id]}`,
             );
+            cleanup();
```
bc-peng, line 161: While using `cleanup()` in each test ensures a clean testing environment, it may be redundant to repeatedly test the same scenario with the same type of data multiple times.

Can we get rid of both `forEach` and `clearnup` here?

Also, please replace `toEqual` here.

vitalii-koshovyi: Discussed in slack

packages/core/src/app/payment/paymentMethod/PaymentMethodTitle.test.tsx
```diff
@@ -1,5 +1,6 @@
 import { CheckoutService, createCheckoutService, StoreConfig } from '@bigcommerce/checkout-sdk';
-import { mount } from 'enzyme';
+import { cleanup, render, screen } from '@testing-library/react'
+import '@testing-library/jest-dom'
```
bc-peng, line 3: Is there any special reason for `import @testing-library/jest-dom` again?
As you can see  on the first line of `jest-setup.ts`, it is already there.

packages/core/src/app/payment/paymentMethod/PaymentMethodTitle.test.tsx
animesh1987: Hey @vitalii-koshovyi does it make sense to rewrite the test without using a for loop, for example
```
it('renders only logo for AmazonPay', () => {
    const methodIds = [
        PaymentMethodId.AmazonPay,
    ];

    render(
        <PaymentMethodTitleTest
            {...defaultProps}
            method={{
                ...defaultProps.method,
                methodIds,
            }}
        />,
    );

    const logos = screen.getByRole('img');
    const names = screen.queryByRole('heading');

    expect(logos).toBeInTheDocument();

    expect(names).not.toBeInTheDocument();
});

it('renders only logo for certain ChasePay', () => {
    const methodIds = [
        PaymentMethodId.ChasePay,
    ];

    render(
        <PaymentMethodTitleTest
            {...defaultProps}
            method={{
                ...defaultProps.method,
                methodIds,
            }}
        />,
    );

    const logos = screen.getByRole('img');
    const names = screen.queryByRole('heading');

    expect(logos).toBeInTheDocument();

    expect(names).not.toBeInTheDocument();
});
```

I believe its a small refactor and that way we can avoid using `cleanup` manually.

packages/core/src/app/payment/paymentMethod/PaymentMethodTitle.test.tsx
vitalii-koshovyi: yes, we can delete it.

vitalii-koshovyi: without it I'm getting an error `Property 'toHaveAttribute' does not exist on type 'JestMatchers<HTMLElement>'.`

packages/core/src/app/payment/paymentMethod/PaymentMethodTitle.test.tsx
```diff
@@ -0,0 +1,688 @@
...
+                method={{
+                    ...defaultProps.method,
+                    method: PaymentMethodType.ApplePay,
+                }}
+            />,
+        );
+
+        const expectedPath = /^http/.test(LOGO_PATHS[PaymentMethodType.ApplePay])
+            ? LOGO_PATHS[PaymentMethodType.ApplePay]
+            : `${config.cdnPath}${LOGO_PATHS[PaymentMethodType.ApplePay]}`;
```
bc-peng, line 106: Is it possible that `expectedPath` here can be different on different test runs?
If not, we should be able to remove the `if` here.

packages/core/src/app/payment/paymentMethod/PaymentMethodTitle.test.tsx
```diff
@@ -0,0 +1,688 @@
...
+                    }}
+                />,
+            );
+    
+            const logos = screen.getByRole('img');
+            const names = screen.getByRole('heading');
+    
+            expect(logos).toBeInTheDocument();
+    
+            expect(names).toBeInTheDocument();
```
bc-peng, line 157: Can we make the selection more method specific?
Do we know the img url or heading text here?

packages/core/src/app/payment/paymentMethod/PaymentMethodTitle.test.tsx
```diff
@@ -0,0 +1,688 @@
...
+                    }}
+                />,
+            );
+
+            const logos = screen.getByRole('img');
+            const names = screen.queryByRole('heading');
+
+            expect(logos).toBeInTheDocument();
+
+            expect(names).not.toBeInTheDocument();
```
bc-peng, line 370: Same here. Can we add src or text content here?

packages/core/src/app/payment/paymentMethod/PaymentMethodTitle.test.tsx
vitalii-koshovyi: there is no need to do it, because our component is simple and it has just title and logo. So we don't have a need to test logo src or text content here. Just need to test is there both text and logo or just logo.

packages/core/src/app/payment/paymentMethod/PaymentMethodTitle.test.tsx
vitalii-koshovyi: same here

packages/core/src/app/payment/paymentMethod/PaymentMethodTitle.test.tsx
bc-peng: I believe we can all agree that the `checkout-js` is constantly changing.
As a result, we should make certain that the logo and name continuously satisfy our expectations in the long run.

PR #1299, opened 2023-05-25
fix(payment): CHECKOUT-7498 Fix issue with component resolver

animesh1987:
## What?
Fix issue with component resolver for a query.

## Why?
We have an issue if a component is registered with the following query `{ id: 'credit_card', gateway: 'bluesnap' }` and if we try to resolve a component with `{ id: 'credit_card', gateway: 'checkoutcom' }`. 

The first component is returned, which can cause potential issues. Hence add extra checks to resolver to return the correct component.

## Testing / Proof
CI 

@bigcommerce/checkout


vitalii-koshovyi (approved):
not bad!

serhii-tkachenko:
@animesh1987 thanks for this fix 🙏

Just want to understand if it covers an issue with gateways as well.

For example we have generic component for PPCP APMs, and one specific component for RatePay (for example). In this case APMs like:
```
{ id: 'giropay', gateway: 'paypalcommercealternativemethods' },
{ id: 'sepa', gateway: 'paypalcommercealternativemethods' },
{ id: 'sofort', gateway: 'paypalcommercealternativemethods' },
```
should be covered by component registered as `{ gateway: 'paypalcommercealternativemethods' }`

and 
`{ id: 'ratepay', gateway: 'paypalcommercealternativemethods' }`
should be covered by component registered as `{ id: 'ratepay', gateway: 'paypalcommercealternativemethods' }`.

Should it work correctly?

animesh1987:
Hey @serhii-tkachenko I can try and add tests for the above cases just to validate our cases, so we are on the same page.

packages/core/src/app/common/resolver/resolveComponent.spec.tsx
```diff
@@ -71,4 +71,56 @@ describe('resolveComponent', () => {
...
+            { id: 'credit_card', gateway: 'checkoutcom' },
+            { Component }
+        );
+
+        expect(CreditCard).toBeDefined();
+        expect(Bluesnap).toBeDefined();
+        expect(CheckoutCom).toBeUndefined();
+    });
+
+    it('returns correct component for an entry in paypal', () => {
```
serhii-tkachenko, line 101: We can update the naming with:
`returns default component for an entry which does not have specific component to resolve` or something similar

serhii-tkachenko (approved):
Nice work! 👍

bc-peng (approved):
Nice job. 👏
Converting this test suite into RTL seems not chanllenging.

animesh1987:
> Converting this test suite into RTL seems not challenging.

Should be straightforward as changing name of the file. Let me try later today.

packages/core/src/app/common/resolver/resolveComponent.test.tsx
```diff
@@ -1,4 +1,4 @@
-import { render } from 'enzyme';
+import {render} from '@testing-library/react'
```
bc-peng, line 1:
```suggestion
import { render } from '@testing-library/react'
```

bc-peng (approved):
😝 A message in Apple Style :
👏 Congratulations to my amazing colleague on successfully conquering his first React testing library test! Your dedication and commitment to mastering new skills are truly inspiring. Keep up the phenomenal work!

PR #1312, opened 2023-06-06
Return of the reverted "feat(payment): INT-7169 BlueSnapDirect: Add payment method component

vitalii-koshovyi:
## What?
Return of reverted https://github.com/bigcommerce/checkout-js/pull/1281

## Why?
Due to the resolver fix

## Testing / Proof
Tested manually

@bigcommerce/checkout


packages/bluesnap-direct-integration/src/BlueSnapDirectEcpFieldset.tsx
```diff
@@ -0,0 +1,65 @@
...
+            { label: 'Consumer savings', value: 'CONSUMER_SAVINGS' },
+            { label: 'Corporate checking', value: 'CORPORATE_CHECKING' },
+            { label: 'Corporate savings', value: 'CORPORATE_SAVINGS' },
+        ],
+    };
+
+    return (
+        <Fieldset
+            legend={<Legend hidden>Electronic Check (ACH/ECP)</Legend>}
+            style={{ paddingBottom: '1rem' }}
```
animesh1987, line 33: Is there a reason why we are using inline styles?

packages/hosted-credit-card-integration/src/HostedCreditCardPaymentMethod.tsx
```diff
@@ -276,5 +276,6 @@ export default toResolvableComponent<PaymentMethodProps, PaymentMethodResolveId>
             id: 'hosted-credit-card',
             // type: 'PAYMENT_TYPE_API',
```
animesh1987, line 277: Can we please remove this line

packages/bluesnap-direct-integration/src/BlueSnapDirectEcpFieldset.tsx
vitalii-koshovyi: Let me check it and fix it in next PRs

TomA-R (approved):

PR #1318, opened 2023-06-07
fix(payment): PAYPAL-2595 fixed the issue with place order button for free product

serhii-tkachenko:
## What?
Fixed the issue with place order button for cart with free products only

## Why?
Place order button was switched to PayPal smart button even for the the carts with free products only. This happened due to the paypal payment strategy initialisation on payment form render, so we should prevent initialisation process in cases when isPaymentDataRequired flag is false.

## Testing / Proof
Unit tests
Manual tests

Screenshot 2023-06-07 at 18 30 24
Screenshot 2023-06-07 at 18 32 28


packages/core/src/app/payment/PaymentSubmitButton.test.tsx
```diff
@@ -26,158 +26,155 @@ describe('PaymentSubmitButton', () => {
...
+    it('renders button with default label', () => {
+        render(<PaymentSubmitButtonTest />);
 
-        expect(component.find(Button).props()).toEqual(
-            expect.objectContaining({
-                disabled: true,
-            }),
+        expect(screen.getByTestId(submitButtonId)).toHaveTextContent(
+            languageService.translate('payment.place_order_action'),
         );
```
bc-peng, line 43: RTL encourages you to assert like this as stated here, https://testing-library.com/docs/queries/about/#priority.
Therefore, `toHaveTextContent` can be avoided.
```
expect(screen.getByText(languageService.translate('payment.place_order_action'))).toBeInTheDocument();
```

serhii-tkachenko: Oh, ok. Thanks! Give me a sec I will update the PR

bc-peng (approved):
LGTM, thanks for the fix.

andriiVitvitskyi1990 (approved):

vitalii-koshovyi (approved):

PR #1320, opened 2023-06-08
tech(checkout): PI-000 re-map test id for RTL

vitalii-koshovyi:
## What?
Re-mapping `data-testid` to `data-test` 

## Why?
I think that we using testid in wrong way, because we're duplicating usage of same attribute with different name
To avoid maintaining of two test IDs I've done re-mapping `data-testid` to `data-test` 
before:
image
after:
image

## Testing / Proof
Units

@bigcommerce/checkout


serhii-tkachenko (approved):
Nice work! 👍

animesh1987 (approved):

PR #1321, opened 2023-06-08
fix(payment): PAYPAL-2467 Add checkbox for mandateText

bc-nick:
## What?

Add checkbox for mandateText

## Why?

Add checkbox that the shopper should select in order consent

## Testing / Proof

https://github.com/bigcommerce/checkout-js/assets/99336044/71626891-68f1-4ff5-9c3b-fca25c90f6ff

@bigcommerce/checkout


serhii-tkachenko (approved):
Looks ok to me! But it seems that we can make some refactoring of `BraintreeACHForm` component to move validation and `disableSubmit` button logic to the form component. So we can clean some code and make it easier to understand.

However, we can move forward and make the refactoring later.

serhii-tkachenko:
[Just a little note for the future]
We should disable submit button in cases:
1. when braintree ach form is not valid or;
2. the form is valid and mandate text checkout is checked;

PR #1323, opened 2023-06-09
fix(payment): PAYPAL-2402 Removed billing address from Braintree ACH Form

bc-nick:
## What?

Removed billing address from Braintree ACH Form

## Why?

Remove billing address fields from the ACH payment component within Payments step of checkout. Shoppers should use our native billing address form on the checkout page. 

Depends:
https://github.com/bigcommerce/checkout-sdk-js/pull/2014 
https://github.com/bigcommerce/checkout-sdk-js/pull/2013

## Testing / Proof

Screenshot 2023-06-09 at 17 43 39

Screenshot 2023-06-09 at 17 44 42

### DIFF
--- a/packages/instrument-utils/src/storedInstrument/AccountInstrumentSelect/AccountInstrumentSelect.tsx
+++ b/packages/instrument-utils/src/storedInstrument/AccountInstrumentSelect/AccountInstrumentSelect.tsx
@@ -289,6 +289,9 @@ class AccountInstrumentSelect extends PureComponent<AccountInstrumentSelectProps
         const { selectedInstrumentId } = this.props;
 
         if (prevSelectedInstrumentId !== selectedInstrumentId) {
+            // FIXME: Used setTimeout here because setFieldValue call doesnot set value if called before formik is properly mounted.
+            //        This ensures that update Field value is called after formik has mounted.
+            // See GitHub issue: https://github.com/jaredpalmer/formik/issues/930
             setTimeout(() => this.updateFieldValue(selectedInstrumentId));
         }
     }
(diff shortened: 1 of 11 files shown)


andriiVitvitskyi1990 (approved):

vitalii-koshovyi (approved):

packages/instrument-utils/src/storedInstrument/AccountInstrumentSelect/AccountInstrumentSelect.tsx
```diff
@@ -289,7 +289,7 @@ class AccountInstrumentSelect extends PureComponent<AccountInstrumentSelectProps
         const { selectedInstrumentId } = this.props;
 
         if (prevSelectedInstrumentId !== selectedInstrumentId) {
-            this.updateFieldValue(selectedInstrumentId);
+            setTimeout(() => this.updateFieldValue(selectedInstrumentId));
```
bc-nick, line 292: Just for clarification

I have faced with the case when `updateFieldValue` function calls from `componentDidUpdate` and than from `componentDidMount` and as a result field value is updated with outdated value. It happens because setTimeout calls in the end when stack is empty. To fix this i added `setTimeout` to `componentDidUpdate`

animesh1987: Interesting seems like this is already what we did in `componentDidMount`, can you please add https://github.com/bigcommerce/checkout-js/pull/1323/files#diff-8eb58cc624cc62a268cc751632f75b94df5efcc4a7fc29c11be32f6a065d4825R281 comment here as well please.

animesh1987 (approved):

PR #1329, opened 2023-06-13
feat(payment): PI-34 [AmazonPay] Update Amazon Pay to use the latest …

bc-sebastianszafraniec:
## What?
AmazonPay button design is changed to C0001 in:
- checkout page - height is set to 36px
- cart, mini cart, pdp - height is default

## Testing / Proof
Zrzut ekranu 2023-06-13 o 16 43 38

@bigcommerce/checkout

### DIFF
--- a/packages/core/src/app/customer/customWalletButton/AmazonPayV2Button.tsx
+++ b/packages/core/src/app/customer/customWalletButton/AmazonPayV2Button.tsx
@@ -16,28 +16,18 @@ const AmazonPayV2Button: FunctionComponent<CheckoutButtonProps> = (props) => {
 
 export default AmazonPayV2Button;
 
-const beautifyAmazonButton = ():void => {
+const beautifyAmazonButton = (): void => {
     if (!document.querySelector('.checkout-button-container')) {
         return;
     }
 
     const container = document.querySelector('#amazonpayCheckoutButton > div') as unknown as HTMLElement;
-    const tagline = container?.shadowRoot?.querySelector('.amazonpay-button-view3');
+    const amazonButton = container?.shadowRoot?.querySelector('.amazonpay-button-view1') as unknown as HTMLElement;
 
-    if (container) {
-        container.style.visibility = 'hidden';
-        container.style.width = '100%';
-    }
-
-    if (tagline) {
-        tagline.remove();
-        (container?.shadowRoot?.querySelector('.amazonpay-button-view1') as unknown as HTMLElement).style.height = '36px'
-        container?.shadowRoot?.querySelector('.amazonpay-button-view2')?.remove();
-        container?.shadowRoot?.querySelector('.amazonpay-button-chevrons')?.remove();
-        container.style.visibility = 'visible';
-
+    if (container && amazonButton) {
+        amazonButton.style.height = '36px';
         return;
     }
 
-    setTimeout(beautifyAmazonButton, 100);
+    setTimeout(beautifyAmazonButton, 10);
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
-        "@bigcommerce/checkout-sdk": "^1.393.2",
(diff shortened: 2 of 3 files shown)


packages/core/src/app/customer/customWalletButton/AmazonPayV2Button.tsx
```diff
@@ -16,28 +16,18 @@ const AmazonPayV2Button: FunctionComponent<CheckoutButtonProps> = (props) => {
...
     const container = document.querySelector('#amazonpayCheckoutButton > div') as unknown as HTMLElement;
-    const tagline = container?.shadowRoot?.querySelector('.amazonpay-button-view3');
+    const amazonButton = container?.shadowRoot?.querySelector('.amazonpay-button-view1') as unknown as HTMLElement;
 
-    if (container) {
-        container.style.visibility = 'hidden';
-        container.style.width = '100%';
+    if (container && amazonButton) {
+        amazonButton.style.height = '36px'
+        return
```
vitalii-koshovyi, line 29: ;

packages/checkout-button-integration/src/CheckoutButton.tsx
```diff
@@ -12,19 +12,49 @@ const CheckoutButton: FunctionComponent<CheckoutButtonProps> = ({
     methodId,
     onUnhandledError,
 }) => {
+    const beautifyCheckoutButtons = useCallback((): void => {
+        if (methodId === 'amazonpay') {
+            if (!document.querySelector('.checkout-button-container')) {
+                return;
+            }
+
+            const container: HTMLElement | null = document.querySelector(
```
vitalii-koshovyi, line 21: HTMLElement | undefined - double check it, please

packages/core/src/app/customer/customWalletButton/AmazonPayV2Button.tsx
bc-sebastianszafraniec: corrected

packages/checkout-button-integration/src/CheckoutButton.tsx
bc-sebastianszafraniec: I checked it and it must stay HTMLElement | null as querySelector method is returning HTMLElement or null if it doesn't find any matches.

bc-sebastianszafraniec: I also modified a little bit code below so amazonPayButton variable can match HTMLElement | null types too (it was previously HTMLElement | null | undefined)

packages/checkout-button-integration/src/CheckoutButton.tsx
```diff
@@ -12,19 +12,49 @@ const CheckoutButton: FunctionComponent<CheckoutButtonProps> = ({
...
+            );
+
+            if (!container || !container.shadowRoot) return;
+
+            const amazonPayButton: HTMLElement | null =
+                container.shadowRoot.querySelector('.amazonpay-button-view1');
+
+            if (amazonPayButton) {
+                amazonPayButton.style.height = '36px';
+            }
```
bc-peng, line 32: It looks like we are changing `.amazonpay-button-view1` twice here and the file below. Is this intentional?
Also, the brutal method of beautifiying a button is really not ideal. Can we just keep it inside the button file, not this shared component?

vitalii-koshovyi:
@animesh1987 Could we do it in another ticket? @bc-sebastianszafraniec is a new guy here, so I believe it would a bit stressful for him

packages/checkout-button-integration/src/CheckoutButton.tsx
vitalii-koshovyi: Agree with Peng here

bc-peng: Building an Amazon package, as Animesh recommended, should be the best practice.
The v1 button appears in the customer step rather than at the top of the checkout page. As a result, we ought to focus on V2 button design right now because these are the ones we want merchants to use.

animesh1987 (approved):

vitalii-koshovyi (approved):

PR #1366, opened 2023-07-06
feat(checkout): PI-287 hide cvv validation field for specific providers

vitalii-koshovyi:
## What?
Do not show cvv validation field for specific providers

## Why?
Some providers don't require cvv validation for stored cards(like Bluesnap), so that is why I've made these changes

## Testing / Proof
Before:
Previously checkout ask to validate CVV code for digital items
image
After:
After the changes we can skip CVV for some providers
image

@bigcommerce/checkout

### DIFF
--- a/packages/core/src/app/payment/storedInstrument/isInstrumentCardCodeRequired.ts
+++ b/packages/core/src/app/payment/storedInstrument/isInstrumentCardCodeRequired.ts
@@ -2,6 +2,8 @@ import { LineItemMap, PaymentInstrument, PaymentMethod } from '@bigcommerce/chec
 
 import { UntrustedShippingCardVerificationType } from './CardInstrumentFieldset';
 
+export const PROVIDERS_WITHOUT_CARD_CODE = ['bluesnapdirect'];
+
 export interface IsInstrumentCardCodeRequiredState {
     instrument: PaymentInstrument;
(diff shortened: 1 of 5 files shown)


packages/core/src/app/payment/storedInstrument/isInstrumentCardCodeRequired.ts
```diff
@@ -13,6 +15,11 @@ export default function isInstrumentCardCodeRequired({
     lineItems,
     paymentMethod,
 }: IsInstrumentCardCodeRequiredState): boolean {
+    // If payment provider doesn't require CVV field
```
animesh1987, line 18: Do we need this comment?

packages/instrument-utils/src/guards/isInstrumentCardCodeRequired/isInstrumentCardCodeRequired.ts
```diff
@@ -11,6 +13,11 @@ export default function isInstrumentCardCodeRequired({
     lineItems,
     paymentMethod,
 }: IsInstrumentCardCodeRequiredState): boolean {
+    // If payment provider doesn't require CVV field
```
animesh1987, line 16: Same for this one.

animesh1987 (approved):

PR #1373, opened 2023-07-10
refactor(checkout): PI-245 moved FormContext fully to the UI package

vitalii-koshovyi:
## What?
Moved FormContext fully to the UI package

## Why?
I found that validation errors didn't appear on the `hosted-credit-card-integration` payment form. The issue was that `hosted-credit-card-integration` and credit-card integration used two different instances of the form context. On the payment step we used context from the  `packages/core/src/app/ui` and hosted-credit-card-integration fields used context from the `packages/ui`. So I decided to use one form context across all packages.

## Testing / Proof
Before:
image
After:
image

@bigcommerce/checkout


bc-peng:
Could you please fix the failed unit tests?

animesh1987 (approved):

PR #1383, opened 2023-07-19
feat(payment): PAYPAL-2752 render wallet buttons depends on coupon

bc-dronov:
## What?
As a shopper who just applied 100% discount coupon during the checkout, we don’t want to see PayPal wallet buttons anymore without having to reload the page manually, so that I can complete the purchase without accidentally clicking on the button that doesn’t work anymore.

## Testing / Proof

https://github.com/bigcommerce/checkout-js/assets/130665807/86c8c1f9-0746-4a4f-a099-2cfd362e6474

@bigcommerce/checkout

### DIFF
--- a/packages/core/src/app/customer/CheckoutButtonContainer.spec.tsx
+++ b/packages/core/src/app/customer/CheckoutButtonContainer.spec.tsx
@@ -22,6 +22,7 @@ describe('CheckoutButtonContainer', () => {
         localeContext = createLocaleContext(getStoreConfig());
 
         jest.spyOn(checkoutState.data, 'getCustomer').mockReturnValue(getGuestCustomer());
+        jest.spyOn(checkoutState.data, 'isPaymentDataRequired').mockReturnValue(true);
         jest.spyOn(checkoutState.data, 'getConfig').mockReturnValue(
             merge(getStoreConfig(), {
                 checkoutSettings: {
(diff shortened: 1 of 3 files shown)


packages/core/src/app/customer/CheckoutButtonContainer.tsx
```diff
@@ -138,6 +139,10 @@ function mapToCheckoutButtonContainerProps({
     const availableMethodIds = getSupportedMethodIds(config?.checkoutSettings.remoteCheckoutProviders ?? []);
     const customer = getCustomer();
 
+    if (!isPaymentDataRequired()) {
```
bc-peng, line 142: Could you please add a unit test for this case?

packages/core/src/app/customer/CheckoutButtonContainer.spec.tsx
```diff
@@ -22,6 +22,7 @@ describe('CheckoutButtonContainer', () => {
         localeContext = createLocaleContext(getStoreConfig());
 
         jest.spyOn(checkoutState.data, 'getCustomer').mockReturnValue(getGuestCustomer());
+        jest.spyOn(checkoutState.data, 'isPaymentDataRequired').mockReturnValue(true);
```
vitalii-koshovyi, line 25: should you cover the case in test when `isPaymentDataRequired` returns false?

bc-dronov: Done

vitalii-koshovyi (approved):

bc-nick (approved):

animesh1987 (approved):

bc-peng (approved):

PR #1386, opened 2023-07-20
feat(payment): PAYPAL-2726 added BraintreeAcceleratedCheckout payment method

serhii-tkachenko:
## What?
Added BraintreeAcceleratedCheckout payment method

## Why?
To render Braintree PayPal Connect component in paywall

## Testing / Proof
Unit tests
Manual tests

### DIFF
--- a/packages/braintree-integration/src/BraintreeAcceleratedCheckout/BraintreeAcceleratedCheckoutPaymentMethod.test.tsx
+++ b/packages/braintree-integration/src/BraintreeAcceleratedCheckout/BraintreeAcceleratedCheckoutPaymentMethod.test.tsx
@@ -1,0 +1,68 @@
+import { createCheckoutService, LanguageService } from '@bigcommerce/checkout-sdk';
+import { render } from '@testing-library/react';
+import React from 'react';
+
+import { PaymentFormService } from '@bigcommerce/checkout/payment-integration-api';
+
+import BraintreeAcceleratedCheckoutPaymentMethod from './BraintreeAcceleratedCheckoutPaymentMethod';
+
+describe('BraintreeAcceleratedCheckoutPaymentMethod', () => {
+    const checkoutService = createCheckoutService();
+    const checkoutState = checkoutService.getState();
+
+    const method = {
+        clientToken: 'token',
+        config: {
+            displayName: 'Credit Card',
+            testMode: true,
+        },
+        id: 'braintreeacceleratedcheckout',
+        initializationData: {
+            isAcceleratedCheckoutEnabled: true,
+        },
+        logoUrl: 'http://logo_url_path',
+        method: 'credit-card',
+        supportedCards: ['VISA', 'MC'],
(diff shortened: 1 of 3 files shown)


andriiVitvitskyi1990 (approved):

packages/braintree-integration/src/BraintreeAcceleratedCheckout/BraintreeAcceleratedCheckoutPaymentMethod.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+import { PaymentFormService } from '@bigcommerce/checkout/payment-integration-api';
+
+import BraintreeAcceleratedCheckoutPaymentMethod from './BraintreeAcceleratedCheckoutPaymentMethod';
+
+describe('BraintreeAcceleratedCheckoutPaymentMethod', () => {
+    const checkoutService = createCheckoutService();
+    const checkoutState = checkoutService.getState();
+
+    const method = {
```
bc-nick, line 13: I think would be better to create `getBraintreeAcceletatedCheckoutPaymentMethod` here `packages/test-utils/src/payment-methods.mock.ts`

What do you think?

serhii-tkachenko: I don't think that this data will be useful for other places. So it might be better to keep it here.

serhii-tkachenko: `packages/test-utils/src/payment-methods.mock.ts` is a global package for all integrations, therefore we can keep common test logic for all integration instead of adding provider specific data that used only in that package..

What do you think?

PavlenkoM (approved):

packages/braintree-integration/src/BraintreeAcceleratedCheckout/BraintreeAcceleratedCheckoutPaymentMethod.test.tsx
bc-nick: I got it. Let's leave it as it is

bc-nick (approved):

bc-dronov (approved):

packages/braintree-integration/src/BraintreeAcceleratedCheckout/BraintreeAcceleratedCheckoutPaymentMethod.test.tsx
animesh1987: I think @bc-nick's idea is better to keep the payment method specific mocks in related packages. That way you don't need to have dependency on the test-utils package. Also ideally braintree mocks should only be used by braintree package in theory.

PR #1392, opened 2023-07-21
feat(checkout): CHECKOUT-7549 Execute Commands within Extension Context

bc-peng:
## What?
`checkout-js` executes three extension commands.

## Why?
Enable an extension to update checkout UI.

## Testing / Proof
- CI checks.

@bigcommerce/checkout

### DIFF
--- a/packages/checkout-extension/src/ExtensionContext.tsx
+++ b/packages/checkout-extension/src/ExtensionContext.tsx
@@ -1,13 +1,11 @@
 import { createContext } from 'react';
 
+import { ExtensionState } from './ExtensionProvider';
 import { ExtensionService } from './ExtensionService';
 
 export interface ExtensionContextProps {
     extensionService: ExtensionService;
-    isExtensionEnabled: boolean;
-    isShowingLoadingIndicator: boolean;
-    setIsExtensionEnabled: (isExtensionEnabled: boolean) => void;
-    setIsShowingLoadingIndicator: (isShowingLoadingIndicator: boolean) => void;
+    extensionState: ExtensionState;
 }
 
 export const ExtensionContext = createContext<ExtensionContextProps | undefined>(undefined);
--- a/packages/checkout-extension/src/ExtensionContext.tsx
+++ b/packages/checkout-extension/src/ExtensionContext.tsx
@@ -1,0 +1,13 @@
+import { createContext } from 'react';
+
+import { ExtensionService } from './ExtensionService';
+
+export interface ExtensionContextProps {
+    extensionService: ExtensionService;
+    isExtensionEnabled: boolean;
+    isShowingLoadingIndicator: boolean;
+    setIsExtensionEnabled: (isExtensionEnabled: boolean) => void;
+    setIsShowingLoadingIndicator: (isShowingLoadingIndicator: boolean) => void;
+}
+
+export const ExtensionContext = createContext<ExtensionContextProps | undefined>(undefined);
--- a/packages/checkout-extension/src/ExtensionContext.tsx
+++ b/packages/checkout-extension/src/ExtensionContext.tsx
@@ -4,6 +4,7 @@ import { ExtensionState } from './ExtensionProvider';
 import { ExtensionService } from './ExtensionService';
 
 export interface ExtensionContextProps {
+    isExtensionEnabled: () => boolean;
     extensionService: ExtensionService;
     extensionState: ExtensionState;
 }
--- a/packages/checkout-extension/src/ExtensionProvider.tsx
+++ b/packages/checkout-extension/src/ExtensionProvider.tsx
@@ -15,7 +15,7 @@ export interface ExtensionAction {
 }
 
 export enum ExtensionActionType {
-    SET_IS_LOADING_INDICATOR,
+    SHOW_LOADING_INDICATOR,
 }
 
 export interface ExtensionProviderProps {
@@ -27,19 +27,18 @@ export const ExtensionProvider = ({ checkoutService, children }: ExtensionProvid
     const isExtensionEnabled = () =>
         isCheckoutExtensionEnabled(checkoutService.getState().data.getConfig()?.checkoutSettings);
 
-    const initialState = {
-        isShowingLoadingIndicator: false,
-    };
     const extensionReducer = (state: ExtensionState, action: ExtensionAction): ExtensionState => {
         switch (action.type) {
-            case ExtensionActionType.SET_IS_LOADING_INDICATOR:
+            case ExtensionActionType.SHOW_LOADING_INDICATOR:
                 return { ...state, isShowingLoadingIndicator: action.payload };
 
             default:
                 return state;
         }
     };
-    const [extensionState, dispatch] = useReducer(extensionReducer, initialState);
+    const [extensionState, dispatch] = useReducer(extensionReducer, {
+        isShowingLoadingIndicator: false,
+    });
 
     const extensionService = new ExtensionService(checkoutService, dispatch);
 
--- a/packages/checkout-extension/src/ExtensionProvider.tsx
+++ b/packages/checkout-extension/src/ExtensionProvider.tsx
@@ -1,0 +1,28 @@
+import { CheckoutService } from '@bigcommerce/checkout-sdk';
+import React, { ReactNode, useState } from 'react';
+
+import { ExtensionService } from './ExtensionService';
+import { ExtensionContext } from './ExtensionContext';
+
+export interface ExtensionProviderProps {
+    checkoutService: CheckoutService;
+    children: ReactNode;
+}
+
+export const ExtensionProvider = ({ checkoutService, children }: ExtensionProviderProps) => {
+    const [isExtensionEnabled, setIsExtensionEnabled] = useState(false);
+    const [isShowingLoadingIndicator, setIsShowingLoadingIndicator] = useState(false);
+
+    const extensionService = new ExtensionService(checkoutService, setIsShowingLoadingIndicator);
+    const extensionValues = {
+        extensionService,
+        isExtensionEnabled,
+        isShowingLoadingIndicator,
+        setIsExtensionEnabled,
+        setIsShowingLoadingIndicator,
+    };
+
+    return (
+        <ExtensionContext.Provider value={extensionValues}>{children}</ExtensionContext.Provider>
+    );
+};
--- a/packages/checkout-extension/src/ExtensionProvider.tsx
+++ b/packages/checkout-extension/src/ExtensionProvider.tsx
@@ -1,8 +1,23 @@
 import { CheckoutService } from '@bigcommerce/checkout-sdk';
-import React, { ReactNode, useState } from 'react';
+import React, { ReactNode, useReducer } from 'react';
 
-import { ExtensionService } from './ExtensionService';
 import { ExtensionContext } from './ExtensionContext';
+import { ExtensionService } from './ExtensionService';
+
+export interface ExtensionState {
+    isExtensionEnabled: boolean;
+    isShowingLoadingIndicator: boolean;
+}
+
+export interface ExtensionAction {
+    type: ExtensionActionType;
+    payload: boolean;
+}
+
+export enum ExtensionActionType {
+    SET_IS_EXTENSION_ENABLED,
+    SET_IS_LOADING_INDICATOR,
+}
 
 export interface ExtensionProviderProps {
     checkoutService: CheckoutService;
@@ -10,16 +25,31 @@ export interface ExtensionProviderProps {
 }
 
 export const ExtensionProvider = ({ checkoutService, children }: ExtensionProviderProps) => {
-    const [isExtensionEnabled, setIsExtensionEnabled] = useState(false);
-    const [isShowingLoadingIndicator, setIsShowingLoadingIndicator] = useState(false);
+    const initialState = {
+        isExtensionEnabled: false,
+        isShowingLoadingIndicator: false,
+    };
 
-    const extensionService = new ExtensionService(checkoutService, setIsShowingLoadingIndicator);
+    const extensionReducer = (state: ExtensionState, action: ExtensionAction): ExtensionState => {
+        switch (action.type) {
+            case ExtensionActionType.SET_IS_EXTENSION_ENABLED:
+                return { ...state, isExtensionEnabled: action.payload };
+
+            case ExtensionActionType.SET_IS_LOADING_INDICATOR:
+                return { ...state, isShowingLoadingIndicator: action.payload };
+
+            default:
+                return state;
+        }
+    };
+
+    const [extensionState, dispatch] = useReducer(extensionReducer, initialState);
+
+    const extensionService = new ExtensionService(checkoutService, dispatch);
+
     const extensionValues = {
         extensionService,
-        isExtensionEnabled,
-        isShowingLoadingIndicator,
-        setIsExtensionEnabled,
-        setIsShowingLoadingIndicator,
+        extensionState,
     };
 
     return (
--- a/packages/checkout-extension/src/ExtensionProvider.tsx
+++ b/packages/checkout-extension/src/ExtensionProvider.tsx
@@ -3,9 +3,9 @@ import React, { ReactNode, useReducer } from 'react';
 
 import { ExtensionContext } from './ExtensionContext';
 import { ExtensionService } from './ExtensionService';
+import { isCheckoutExtensionEnabled } from './isCheckoutExtensionEnabled';
 
 export interface ExtensionState {
-    isExtensionEnabled: boolean;
     isShowingLoadingIndicator: boolean;
 }
 
@@ -15,7 +15,6 @@ export interface ExtensionAction {
 }
 
 export enum ExtensionActionType {
-    SET_IS_EXTENSION_ENABLED,
     SET_IS_LOADING_INDICATOR,
 }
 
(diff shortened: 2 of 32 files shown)


bc-peng:
WIP: Working on adding unit tests.

packages/checkout-extension/src/withExtension.ts
```diff
@@ -0,0 +1,7 @@
+import { createInjectHoc } from '@bigcommerce/checkout/locale';
```
animesh1987, line 1: Do we need to go the HOC pattern with this? Since we are using a context we should avoid this.

bc-peng: I used the analytics context's method to set up the context for extensions.
When I searched for `extends Component`, I found around 40 class component objects within `checkout-js`, including important components like `Checkout`, `Shipping`, and `Payment`.

packages/checkout-extension/src/ExtensionContext.tsx
```diff
@@ -0,0 +1,13 @@
+import { createContext } from 'react';
+
+import { ExtensionService } from './ExtensionService';
+
+export interface ExtensionContextProps {
+    extensionService: ExtensionService;
```
animesh1987, line 6: Is this something that will keep growing as we add more features?

packages/checkout-extension/src/ExtensionProvider.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { ExtensionService } from './ExtensionService';
+import { ExtensionContext } from './ExtensionContext';
+
+export interface ExtensionProviderProps {
+    checkoutService: CheckoutService;
+    children: ReactNode;
+}
+
+export const ExtensionProvider = ({ checkoutService, children }: ExtensionProviderProps) => {
+    const [isExtensionEnabled, setIsExtensionEnabled] = useState(false);
```
animesh1987, line 13: Can we try using a `useReducer` for this?

packages/checkout-extension/src/withExtension.ts
animesh1987, line 1: Another comment not related to this PR importantly is that we need to get this `createInjectHoc` out of locale package. Maybe as a followup after this.

packages/core/src/app/checkout/CheckoutApp.tsx
```diff
@@ -4,10 +4,12 @@ import React, { Component } from 'react';
 import ReactModal from 'react-modal';
 
 import { AnalyticsProvider } from '@bigcommerce/checkout/analytics';
-import '../../scss/App.scss';
+import { ExtensionProvider } from '@bigcommerce/checkout/checkout-extension';
 import { getLanguageService, LocaleProvider } from '@bigcommerce/checkout/locale';
 import { CheckoutProvider } from '@bigcommerce/checkout/payment-integration-api';
 
+import '../../scss/App.scss';
```
animesh1987, line 11: Is this intentional?

packages/core/src/app/shipping/ShippingHeader.tsx
```diff
@@ -12,37 +18,64 @@ interface ShippingHeaderProps {
...
+                            id={isMultiShippingMode ? 'shipping.ship_to_single' : 'shipping.ship_to_multi'}
+                        />
+                    </a>
+                )}
+            </div>
+        </>
+    );
+}
+
+export default withExtension(memo(ShippingHeader));
```
animesh1987, line 81: Another follow up improvement, maybe we can avoid using HOC for function components and use the useContext hook or create a custom useExtensions hook similar to useLocale.

packages/checkout-extension/src/ExtensionService.ts
```diff
@@ -0,0 +1,111 @@
...
+            return;
+        }
+
+        await this._checkoutService.renderExtension(container, region);
+
+        if (!this._handlerRemovers[extension.id]) {
+            this._handlerRemovers[extension.id] = [];
+        }
+
+        this._handlerRemovers[extension.id].push(
```
animesh1987, line 33: Do we need to call these `_handlerRemovers` since that name seems a bit confusing. Can we just call them `_handlers`?

packages/checkout-extension/src/ExtensionService.ts
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    isRegionInUse(region: ExtensionRegion): boolean {
+        const extension = this._getExtensionByRegion(region);
+
+        return Boolean(extension);
+    }
+
+    private _getExtensionByRegion(region: ExtensionRegion): Extension | undefined {
+        const extensions = this._checkoutService.getState().data.getExtensions();
```
animesh1987, line 107: Any reason we are not using `getExtensionByRegion` selector?

packages/checkout-extension/src/ExtensionContext.tsx
bc-peng: Once it is converted using `useReducer`, we'll only have an extension state.

packages/checkout-extension/src/ExtensionProvider.tsx
bc-peng: Will do.

packages/checkout-extension/src/withExtension.ts
bc-peng: Noted. Shall I create a ticket?

packages/core/src/app/checkout/CheckoutApp.tsx
bc-peng: Yes, I was suprised linter didn't pick it up before.

packages/core/src/app/shipping/ShippingHeader.tsx
bc-peng: I built 'useExtensions' before, but we have to use HOC now, so I deleted it to keep usage constant.

packages/checkout-extension/src/ExtensionService.ts
bc-peng: `this._handlerRemovers` holds functions to remove an extension's handlers.
It is essential to be invoked after removing each extension from UI, otherwise, we'll handle commands multiple times if an extension is rendered serveral times.

packages/checkout-extension/src/ExtensionService.ts
bc-peng: `getExtensionByRegion` is a pirvate selector in SDK 😅, only `getExtensions` is public available.

packages/checkout-extension/src/ExtensionService.ts
```diff
@@ -0,0 +1,111 @@
...
+        if (!this._handlerRemovers[extension.id]) {
+            this._handlerRemovers[extension.id] = [];
+        }
+
+        this._handlerRemovers[extension.id].push(
+            this._checkoutService.handleExtensionCommand(
+                extension.id,
+                ExtensionCommandType.ReloadCheckout,
+                () => {
+                    void this._checkoutService.loadCheckout(
```
davidchin, line 38: Does this `void` do anything, considering you're using `() => {}`?

packages/checkout-extension/src/ExtensionService.ts
```diff
@@ -0,0 +1,111 @@
...
+                },
+            ),
+        );
+
+        this._handlerRemovers[extension.id].push(
+            this._checkoutService.handleExtensionCommand(
+                extension.id,
+                ExtensionCommandType.SetIframeStyle,
+                (data) => {
+                    if (data.type === ExtensionCommandType.SetIframeStyle) {
```
davidchin, line 50: I think we can make it so that `handleExtensionCommand` can map `ExtensionCommandType.SetIframeStyle` to a different handler signature. For example, something like this: https://github.com/bigcommerce/checkout-sdk-js/blob/38555c99fe63c6dfccf767c1f9c3ff8962b4cfee/packages/core/src/common/iframe/iframe-event-listener.ts#L41-L44. That way you don't need to do this check on the client side.

packages/core/src/app/shipping/ShippingFormFooter.tsx
```diff
@@ -18,19 +24,50 @@ export interface ShippingFormFooterProps {
...
+
+        if (this._isRegionInUse) {
+            extensionService.removeListeners(ExtensionRegion.ShippingShippingAddressFormAfter);
+        }
+    }
+
+    async componentDidMount(): Promise<void> {
+        const { extensionService } = this.props;
+
+        if (this._isRegionInUse) {
```
davidchin, line 41: Do you need this check or the need to keep track of `_isRegionInUse` in here? I think `ExtensionService#renderExtension` method already does a check so it won't render it again if the region is already occupied?

packages/checkout-extension/src/ExtensionService.ts
```diff
@@ -0,0 +1,111 @@
...
+                ExtensionCommandType.ReloadCheckout,
+                () => {
+                    void this._checkoutService.loadCheckout(
+                        this._checkoutService.getState().data.getCheckout()?.id,
+                    );
+                },
+            ),
+        );
+
+        this._handlerRemovers[extension.id].push(
```
davidchin, line 45: It's worth thinking about to how to extract the implementation of the handlers out, because I can imagine that this method can get quite long when we support more commands in the future.

Maybe each handler could be an exportable function on its own. And then we have a way of importing and registering them against each command type?

packages/core/src/app/checkout/Checkout.tsx
```diff
@@ -253,6 +260,12 @@ class Checkout extends Component<
             }
 
             window.addEventListener('beforeunload', this.handleBeforeExit);
+
+            if (isCheckoutExtensionEnabled(data.getConfig()?.checkoutSettings)) {
+                setIsExtensionEnabled(true);
```
davidchin, line 265: I think you can directly read from the return value of `isCheckoutExtensionEnabled(data.getConfig()?.checkoutSettings)` and pass it as a prop via the `ExtensionContext` rather than having to set it to the state first and then read from it.

packages/checkout-extension/src/withExtension.ts
animesh1987: Yeah I think good idea, maybe with the onboarding label 😄

packages/checkout-extension/src/ExtensionService.ts
animesh1987: Hmm I think might be a good idea to make it available outside SDK as the clients as we can see here need to duplicate that logic.

packages/checkout-extension/src/ExtensionService.ts
animesh1987: I think these are still handlers in concept, the function as part of clean up can be called removehandlers 🤔 I say so because handlerRemovers feels as an odd name and in theory it is a list of handlers/functions to be executed on receiving a command.

packages/checkout-extension/src/withExtension.ts
bc-peng: For the record, the ticket is https://bigcommercecloud.atlassian.net/browse/CHECKOUT-7633.

packages/checkout-extension/src/ExtensionService.ts
bc-peng: I want to avoid SDK depency in this PR.
Will do it right after this PR, here is a [ticket](https://bigcommercecloud.atlassian.net/browse/CHECKOUT-7634.) to track it.

packages/core/src/app/shipping/ShippingFormFooter.tsx
bc-peng: `isRegionInUse` also represents the extension experiment flag.
So here is to skip the function if the experiment flag is `off`.

packages/checkout-extension/src/ExtensionService.ts
animesh1987: Sounds perfect, we can do it as a fast follow 😄

packages/checkout-extension/src/ExtensionService.ts
bc-peng: Yes, eslint suggests marking the returned promise from `this._checkoutService.loadCheckout` as ignored with a `void` operator. Otherwise, I need to change it to `async () => void`.

packages/checkout-extension/src/ExtensionService.ts
animesh1987: I think this is a good starting point in my opinion, for extracting handlers and making things more generic we can probably do it as a fast follow once we deliver this desired functionality 🤔 We can even come back to this and re evaluate in more detail when we do come back to extensions as a more robust feature.

packages/core/src/app/shipping/ShippingFormFooter.tsx
davidchin: I suggest moving the experiment check in `ExtensionService#renderExtension` as well

bc-peng: Sure, can do that in a follow-up SDK PR.
Regarding `isRegionInUse`, do we want to avoid putting an empty div?
```
            {isRegionInUse && (
                <div id={ExtensionRegionContainer.ShippingShippingAddressFormBefore} />
            )}
```

packages/checkout-extension/src/ExtensionService.ts
davidchin: Unless you think the proposed refactoring would take a lot of time, my recommendation is to do it now when things are still fresh in our minds so we have a scalable pattern to follow moving forward.

animesh1987: I think we can do the refactoring as a fast follow in this case, so we can unblock the feature delivery and meet the agreements and make the improvements as a follow up.

bc-peng: I believe this task is time consuming, but we can handle it in a separate ticket (https://bigcommercecloud.atlassian.net/browse/CHECKOUT-7635) following this PR. 
This way, we can share sample apps even before completing the entire project.

packages/checkout-extension/src/ExtensionService.ts
```diff
@@ -0,0 +1,111 @@
...
+        delete this._handlerRemovers[extension.id];
+    }
+
+    isRegionInUse(region: ExtensionRegion): boolean {
+        const extension = this._getExtensionByRegion(region);
+
+        return Boolean(extension);
+    }
+
+    private _getExtensionByRegion(region: ExtensionRegion): Extension | undefined {
```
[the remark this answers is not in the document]
bc-peng: Updated for all occurrences.

packages/checkout-extension/src/ExtensionService.ts
bc-peng: Renamed.

packages/checkout-extension/src/ExtensionService.ts
bc-peng: Thanks for pointing out the mapping.
I have added this as a new task in the follow-up SDK ticket.

packages/core/src/app/shipping/ShippingFormFooter.tsx
davidchin: > do we want to avoid putting an empty div?

Yes we want to, but not by setting a property inside `render` method and reading from it.

packages/checkout-extension/src/ExtensionService.ts
davidchin: > I believe this task is time consuming

Ok sure, I'm happy for you to do to tackle it separately if there's a contractual deadline we need to meet. 🙂 

Please reference the ticket number as a code comment here.

packages/core/src/app/checkout/Checkout.tsx
bc-peng: As discussed offline, I have changed the experiment value reading method.
Therefore, the condition here is updated as:
https://github.com/bigcommerce/checkout-js/blob/994a53343bd3dc4af0f2407e7cf33076088c88f2/packages/core/src/app/checkout/Checkout.tsx#L264

packages/checkout-extension/src/ExtensionService.ts
bc-peng: I have put a TODO comment.
https://github.com/bigcommerce/checkout-js/blob/994a53343bd3dc4af0f2407e7cf33076088c88f2/packages/checkout-extension/src/ExtensionService.ts#L40

packages/core/src/app/shipping/ShippingFormFooter.tsx
bc-peng: I've converted `ShippingFormFooter` into a functional component, so `isRegionInUse` is assigned differently now.
https://github.com/bigcommerce/checkout-js/blob/994a53343bd3dc4af0f2407e7cf33076088c88f2/packages/core/src/app/shipping/ShippingFormFooter.tsx#L32

davidchin (approved):
Fantastic work @bc-peng LGTM! :shipit:

packages/checkout-extension/src/ExtensionProvider.tsx
```diff
@@ -0,0 +1,54 @@
...
+export interface ExtensionProviderProps {
+    checkoutService: CheckoutService;
+    children: ReactNode;
+}
+
+export const ExtensionProvider = ({ checkoutService, children }: ExtensionProviderProps) => {
+    const isExtensionEnabled = () =>
+        isCheckoutExtensionEnabled(checkoutService.getState().data.getConfig()?.checkoutSettings);
+
+    const extensionReducer = (state: ExtensionState, action: ExtensionAction): ExtensionState => {
```
animesh1987, line 30: In the next clean up PR maybe a good idea to move this out of provider as well :D

animesh1987 (approved):
Fantastic work. LGTM 👍

packages/checkout-extension/src/ExtensionProvider.tsx
bc-peng: Sure, I can include it when bumping https://github.com/bigcommerce/checkout-sdk-js/pull/2085.

PR #1411, opened 2023-08-02
fix(checkout): CHECKOUT-7150 aria-expanded property does not toggle based on the associated drop-down

bc-NikitaPuzanenko:
## What?
Adding an "onBlur" property to the address select button to set the "aria-expanded" property to false when the element loses focus.
## Why?
Currently when a user changes focus to another element, the aria-expanded property does not change back to false. This causes problems for users who navigate checkout using screen reader software.
## Testing / Proof
Before:
Screenshot 2023-08-02 at 1 41 49 PM

After:
Screenshot 2023-08-02 at 1 40 44 PM

Additional note:
I am not sure if it would be better to track when the user clicks off the address select element instead of when it loses focus.

@bigcommerce/checkout


bc-peng (approved):
LGTM. If I remember correctly, the dropdown will be collpased when a shopper clicks anywhere on the page. So setting the value to false on blur offers additional insurance for the correct aria value.

Issue #1390, opened 2023-07-20
Google AddressSelector sometimes autofills the wrong city


egarrett:
We utilize [Google Autocomplete for Addresses](https://support.bigcommerce.com/s/article/Address-Autocomplete?language=en_US) and we've recently discovered an issue that is causing some of our orders to come through  with an incorrect city in the shipping address. 

What seams to be happening is that for some municipalities, the  https://maps.googleapis.com/maps/api/place/js/PlaceService.GetPlaceDetails API returns the city in the `sublocality_level_1` address component, not the `locality` address component.  

To fix this, [AddressSelector.getCity()](https://github.com/bigcommerce/checkout-js/blob/09304e79c27c0d5d1312a697c915803c85347dae/packages/core/src/app/address/googleAutocomplete/AddressSelector.ts#L30) may need to be checking for the `sublocality_level_1` property and using that value if its set. 

You can reproduce this as follows:

1. While checking out, start typing in the following Canadian address: `1934 Montréal Rd, Gloucester, ON K1J 6N2`
2. When you see the auto suggestion for that address, select it to autofill.
3. After the autofill, you will see `Ontario` in the City field but it should be `Gloucester`.

animesh1987:
Thanks for reporting this @egarrett, we will have a look at this in the coming weeks.

bc-peng:
We might need to implement auto address filling rules for CA, similar to [what we did for the UK](https://github.com/bigcommerce/checkout-js/blob/09304e79c27c0d5d1312a697c915803c85347dae/packages/core/src/app/address/googleAutocomplete/AddressSelectorUk.ts#L4).

animesh1987:
Fix released in https://github.com/bigcommerce/checkout-js/commit/88cced4fe3d5b9c2eeb9376fbca3b455d551e533.

PR #1419, opened 2023-08-04
fix(shipping): CHECKOUT-7627 Google AddressSelector sometimes autofills the wrong city for Canada addresses

animesh1987:
## What?
Fix issue where Google maps API sometimes fills the incorrect value in the city field.

## Why?
This was raised in an issue here https://github.com/bigcommerce/checkout-js/issues/1390.

## Testing / Proof
- circle
- screenshot
Screenshot 2023-08-04 at 10 29 20 am
Screenshot 2023-08-04 at 10 29 52 am

@bigcommerce/checkout

### DIFF
--- a/packages/core/src/app/address/googleAutocomplete/AddressSelectorCA.ts
+++ b/packages/core/src/app/address/googleAutocomplete/AddressSelectorCA.ts
@@ -1,0 +1,8 @@
+import AddressSelector from './AddressSelector';
+
+export default class AddressSelectorUK extends AddressSelector {
+    getCity(): string {
+        return this._get('political', 'long_name');
+        ;
+    }
+}
--- a/packages/core/src/app/address/googleAutocomplete/AddressSelectorCA.ts
+++ b/packages/core/src/app/address/googleAutocomplete/AddressSelectorCA.ts
@@ -2,7 +2,6 @@ import AddressSelector from './AddressSelector';
 
 export default class AddressSelectorUK extends AddressSelector {
     getCity(): string {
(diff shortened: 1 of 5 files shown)


packages/core/src/app/address/googleAutocomplete/AddressSelectorCA.ts
```diff
@@ -0,0 +1,7 @@
+import AddressSelector from './AddressSelector';
+
+export default class AddressSelectorUK extends AddressSelector {
```
bc-peng, line 3:
```suggestion
export default class AddressSelectorCA extends AddressSelector {
```

bc-peng: Now you see why I don't like using `default` 🥴

packages/core/src/app/address/googleAutocomplete/AddressSelectorCA.ts
```diff
@@ -0,0 +1,7 @@
...
+    getCity(): string {
+        return this._get('sublocality_level_1', 'long_name') || this._get('locality', 'long_name');
```
bc-peng, line 5: Do we need someone to verify this rule is valid for all CA addresses?

packages/core/src/app/address/googleAutocomplete/AddressSelectorCA.ts
animesh1987: Wow that's next level copy paste error 🤦

bc-peng (approved):

PR #1422, opened 2023-08-07
feat(payment): PAYPAL-2732 new renderer method for address select component

PavlenkoM:
## What?
Adding new renderer method for AddressSelect component

## Why?
Need this new method to get ability to add custom AddressSelect component for PayPal AXO project
Example in this PR: [https://github.com/bigcommerce/checkout-js/pull/1423](https://github.com/bigcommerce/checkout-js/pull/1423)

## Testing / Proof
Manually tested and unit test

@bigcommerce/checkout

### DIFF
--- a/packages/core/src/app/shipping/ItemAddressSelect.spec.tsx
+++ b/packages/core/src/app/shipping/ItemAddressSelect.spec.tsx
@@ -18,6 +18,7 @@ describe('ItemAddressSelect Component', () => {
         addresses: getCustomer().addresses,
         onSelectAddress: jest.fn(),
         onUseNewAddress: jest.fn(),
+        renderAddressSelect: (props) => <AddressSelect {...props} />,
     };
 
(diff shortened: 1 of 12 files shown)


serhii-tkachenko (approved):

packages/core/src/app/address/isValidCustomerAddress.ts
```diff
@@ -1,12 +1,12 @@
...
+import { Address, FormField } from '@bigcommerce/checkout-sdk';
 import { some } from 'lodash';
 
 import isEqualAddress from './isEqualAddress';
 import isValidAddress from './isValidAddress';
 
 export default function isValidCustomerAddress(
     address: Address | undefined,
-    addresses: CustomerAddress[],
+    addresses: Address[],
```
animesh1987, line 9: This feels odd as we are changing the address interface that is passed in, and hence essentially changing the method.

PR #1423, opened 2023-08-07
feat(payment): PAYPAL-2732 PayPalAxo Shipping step

PavlenkoM:
## What?
Add labels for PayPal Connect addresses

## Why?
We need to show for shoppers addresses which we get from PayPal Connect.

## Testing / Proof
BC - guest
PP Connect - guest

https://github.com/bigcommerce/checkout-js/assets/9430298/1027b77f-e1df-4c18-ac96-5bc6bf3c91da

BC - registered
PP Connect - guest

https://github.com/bigcommerce/checkout-js/assets/9430298/6894e174-03ed-4c99-bac2-7a09b3ba5bd5

BC - guest
PP Connect - registered

https://github.com/bigcommerce/checkout-js/assets/9430298/8291b50a-99bb-4f65-9c27-fe50553cdc0d

BC - registered
PP Connect - registered (with address merge)

https://github.com/bigcommerce/checkout-js/assets/9430298/aa14d130-f221-4fbb-805e-c65ee0d44c36

@bigcommerce/checkout

### DIFF
--- a/packages/core/src/app/address/AddressSelect.tsx
+++ b/packages/core/src/app/address/AddressSelect.tsx
@@ -8,6 +8,7 @@ import { DropdownTrigger } from '../ui/dropdown';
 
 import AddressSelectButton from './AddressSelectButton';
 import isEqualAddress from './isEqualAddress';
+import { PoweredByPaypalConnectLabel, usePayPalConnectAddress } from './PayPalAxo';
 import StaticAddress from './StaticAddress';
 
 import './AddressSelect.scss';
@@ -38,7 +39,7 @@ const AddressSelectMenu: FunctionComponent<AddressSelectProps> = ({
         {addresses.map((address) => (
             <li className="dropdown-menu-item dropdown-menu-item--select" key={address.id}>
                 <a href="#" onClick={preventDefault(() => onSelectAddress(address))}>
-                    <StaticAddress address={address} />
+                    <StaticAddress address={address} showProviderIcon />
                 </a>
             </li>
         ))}
@@ -51,6 +52,8 @@ const AddressSelect = ({
     onSelectAddress,
     onUseNewAddress,
 }: AddressSelectProps) => {
+    const { shouldShowPayPalConnectLabel } = usePayPalConnectAddress();
+
     const handleSelectAddress = (newAddress: Address) => {
         if (!isEqualAddress(selectedAddress, newAddress)) {
             onSelectAddress(newAddress);
@@ -80,6 +83,8 @@ const AddressSelect = ({
                     />
                 </DropdownTrigger>
             </div>
+
+            {shouldShowPayPalConnectLabel() && <PoweredByPaypalConnectLabel />}
         </div>
     );
 }
--- a/packages/core/src/app/address/AddressSelect.tsx
+++ b/packages/core/src/app/address/AddressSelect.tsx
@@ -39,7 +39,7 @@ const AddressSelectMenu: FunctionComponent<AddressSelectProps> = ({
         {addresses.map((address) => (
             <li className="dropdown-menu-item dropdown-menu-item--select" key={address.id}>
                 <a href="#" onClick={preventDefault(() => onSelectAddress(address))}>
-                    <StaticAddress address={address} showProviderIcon />
+                    <StaticAddress address={address} />
                 </a>
             </li>
         ))}
@@ -84,7 +84,7 @@ const AddressSelect = ({
                 </DropdownTrigger>
             </div>
 
-            {shouldShowPayPalConnectLabel() && <PoweredByPaypalConnectLabel />}
+            {shouldShowPayPalConnectLabel && <PoweredByPaypalConnectLabel />}
         </div>
     );
 }
--- a/packages/core/src/app/address/StaticAddress.scss
+++ b/packages/core/src/app/address/StaticAddress.scss
@@ -1,3 +1,5 @@
+@import "../ui/Base";
+
 .checkout-address--static {
     .address-entry {
         margin: 0;
@@ -5,17 +7,15 @@
 }
 
 .checkout-address--with-provider-icon {
-    $icon-size: 1.85rem;
-
     position: relative;
 
     .address-entry {
-        padding-right: $icon-size;
+        padding-right: $paypal-connect-icon-size;
     }
 
     .icon {
-        width: $icon-size;
-        height: $icon-size;
+        width: $paypal-connect-icon-size;
+        height: $paypal-connect-icon-size;
         position: absolute;
         top: 0;
         right: 0;
--- a/packages/core/src/app/address/StaticAddress.scss
+++ b/packages/core/src/app/address/StaticAddress.scss
@@ -1,5 +1,18 @@
 .checkout-address--static {
+    $icon-size: 1.85rem;
+
+    position: relative;
+
     .address-entry {
         margin: 0;
+        padding-right: $icon-size;
+    }
+
+    .icon {
+        width: $icon-size;
+        height: $icon-size;
+        position: absolute;
+        top: 0;
+        right: 0;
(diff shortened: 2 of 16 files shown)


serhii-tkachenko (approved):

serhii-tkachenko (approved):
Great work!

animesh1987:
Hey @PavlenkoM sorry to cause issues again, can you please breakdown this PR into smaller chunks. For example you can isolate AddressSelect changes to another PR.

packages/core/src/app/address/PayPalAxo/PoweredByPaypalConnectLabel.scss
```diff
@@ -0,0 +1,10 @@
+.powered-by-paypal-connect-label {
+    font-size: 0.77rem;
+    margin: 0.77rem 0 0;
+
+    & > .icon {
+        margin: 0 0.5em;
+        width: 4.24rem;
+        height: 0.77rem;
+    }
+}
```
animesh1987, line 10: EOL

packages/core/src/app/address/AddressSelect.tsx
```diff
@@ -1,65 +1,25 @@
...
 import { TranslatedString } from '@bigcommerce/checkout/locale';
 
 import { preventDefault } from '../common/dom';
 import { DropdownTrigger } from '../ui/dropdown';
 
-import isEqualAddress from './isEqualAddress';
-import './AddressSelect.scss';
 import AddressSelectButton from './AddressSelectButton';
+import isEqualAddress from './isEqualAddress';
+import { PoweredByPaypalConnectLabel, usePayPalConnectAddress } from './PayPalAxo';
```
bc-peng, line 11: This PR looks good to me.
Does it make sense to create a PalPalAxo package?
cc @animesh1987

PavlenkoM: This changes will be moved in Braintree package in future.

PavlenkoM:
> Hey @PavlenkoM sorry to cause issues again, can you please breakdown this PR into smaller chunks. For example you can isolate AddressSelect changes to another PR.
> 
> And then different shopper flows.

Most of changes in this PR is fixing of "unit" test and snapshot.
But ok, I will split it

packages/core/src/app/address/AddressSelect.tsx
bc-peng: Maybe we can create `Braintree` package now so that we don't forget to come back here.

serhii-tkachenko: Hey @bc-peng. We created a task for that and I am planing to move this part to PayPalConnect/PayPalAxo package in i19 sprint.

Is it ok for you?

serhii-tkachenko: We will clean up core package from paypal specific code as soon as possible

packages/core/src/app/address/StaticAddress.scss
```diff
@@ -1,5 +1,18 @@
 .checkout-address--static {
+    $icon-size: 1.85rem;
```
bc-peng, line 2: Sorry I missed this before.
Could you please move this setting to `packages/core/src/scss/settings/foundation/forms/_settings.scss` and provide more information in the varible's name?
Plus, I didn't see we use `1.85rem` before, can you please justify why this value is appropriate?

packages/core/src/app/address/StaticAddress.scss
```diff
@@ -1,5 +1,18 @@
...
+
+    position: relative;
+
     .address-entry {
         margin: 0;
+        padding-right: $icon-size;
```
bc-peng, line 8: I undertand you want to make space for the icon.
However, this class also applies to the condition when there is no icon.
Could you please create a new CSS class for PP icon so that it won't impact existing UI?

packages/core/src/app/address/StaticAddress.scss
PavlenkoM: This size is specific only to one icon PP Connect, and size 1.85 is regarding to design.
In case when it is a size only for one icon in specific case I don't think that we need to move it global settings.

packages/core/src/app/address/StaticAddress.scss
PavlenkoM: yes, you are right, it will be better, will do

packages/core/src/app/address/StaticAddress.scss
bc-peng: I'm aware that it's a setting just for address.
However, except in the setting file, we strongly oppose putting numbers into CSS code. By doing so, we can control all of these different sizes from a single location.
Perhaps we could name it `$paypal-connect-icon-size` and put it there.

PavlenkoM: done

packages/core/src/app/address/StaticAddress.scss
PavlenkoM: done

packages/core/src/app/address/AddressSelect.tsx
```diff
@@ -38,7 +39,7 @@ const AddressSelectMenu: FunctionComponent<AddressSelectProps> = ({
         {addresses.map((address) => (
             <li className="dropdown-menu-item dropdown-menu-item--select" key={address.id}>
                 <a href="#" onClick={preventDefault(() => onSelectAddress(address))}>
-                    <StaticAddress address={address} />
+                    <StaticAddress address={address} showProviderIcon />
```
animesh1987, line 42: Does this mean we are always passing `showProviderIcon` as `true` for `<StaticAddress />` now?

PavlenkoM: in current place - yes
need this flag because of we should show this icon not for all places when we reuse this code
and checking for show/hide PP icon inside [the component](https://github.com/bigcommerce/checkout-js/pull/1423/files#diff-ded4d841a24b30f3c3ba48ec964a53a61045c8156846829a1aab4bb7ee986ed9R49)

animesh1987: Can we not just make the decision to show the label if the address is from paypal instead of adding an optional prop? Since we check if the address is from Paypal or not, I think that check should be good enough 🤔

PavlenkoM: the same components `StaticAddress` with the same addresses but one with icon and another without
Screenshot 2023-08-16 at 15 58 21
Screenshot 2023-08-16 at 15 58 41

PavlenkoM: PayPal approved to add icon to each places with this component, so will remove this flag

PavlenkoM:
Hi @animesh1987 
As we discussed I have updated this PR
• remove additional property from `StaticAddress` component
• remove logic with LocalStorage
• updated hook, for now it returns only value object and method was moved to utils

packages/core/src/scss/settings/foundation/forms/_settings.scss
```diff
@@ -165,3 +165,5 @@ $select-outline:                     0;
 $select-paddingRight:                spacing("single");
 $select-focus-borderColor:           $input-focus-border-color;
 $select-focus-boxShadow:             $input-focus-boxShadow;
+
+$paypal-connect-icon-size:           remCalc(30px);
```
bc-peng, line 169:
```suggestion

// We use it to display a PayPal Connect icon next to the static shipping address
$paypal-connect-icon-size:           remCalc(30px);
```

animesh1987 (approved):

animesh1987:
Nice work @PavlenkoM, thanks for the patience with this one.

bc-peng:
Kudos to you!!!!

PR #1432, opened 2023-08-14
refactor(extension): CHECKOUT-7635 Improve Command Handler Scalability

bc-peng:
## What?
Refactor extension command handlers by:

- Extracting handlers into separate functions.

- Implementing registration of handlers.

## Why?
Currently, command handlers are tightly integrated, making future scalability challenging. 

## Testing / Proof
- CI checks.


bc-peng:
Will add tests later.

packages/checkout-extension/src/handler/ReloadCheckout.ts
```diff
@@ -0,0 +1,13 @@
+import { ExtensionCommandType } from '@bigcommerce/checkout-sdk';
+
+import { CommandHandler } from './CommandHandler';
+
+export const reloadCheckout: CommandHandler = ({ checkoutService, extension }) => {
```
davidchin, line 5: Perhaps we could simplify this by doing the following?

```ts
export function createReloadCheckoutHandler(props: CommandHandlerProps): CommandHandler {
    return (command: ReloadCheckoutCommand) => {
        // Reload checkout
    };
}

createReloadCheckoutHandler.commandType = ExtensionCommandType.ReloadCheckout;
```

And then we could do similar to what you've done in `ExtensionService` where we loop through each exported handler and register them:

```ts
Object.values(commandHandlerFactories).forEach((createHandler) => {
    this.handlers[extension.id].push(this.checkoutService.handleExtensionCommand(
        extension.id,
        createHandler.commandType,
        createHandler(props),
    ));
});
```

That way we don't have to call `handleExtensionCommand` in each of the function?

packages/checkout-extension/src/ExtensionService.ts
```diff
@@ -104,4 +59,37 @@ export class ExtensionService {
...
+            checkoutService: this.checkoutService,
+            dispatch: this.dispatch,
+            extension,
+        };
+
+        if (!this.handlers[extension.id]) {
+            this.handlers[extension.id] = [];
+        }
+
+        const isCommandHandler = <T extends keyof ExtensionCommandMap>(
```
animesh1987, line 74: Minor comment, does it make sense to move this as a private method outside this function?

packages/checkout-extension/src/ExtensionService.ts
```diff
@@ -104,4 +59,37 @@ export class ExtensionService {
...
+            type: T,
+            handler: CommandHandler<any>,
```
animesh1987, line 76:
```suggestion
            handler: CommandHandler<unknown>,
```

bc-peng: TS does not allow it to be unknown and throws an error:
```
TS2344: Type 'unknown' does not satisfy the constraint 'keyof ExtensionCommandMap'
```

animesh1987 (approved):
Nice work 👏

PR #1451, opened 2023-08-21
fix(checkout): PI-122 fixed credit card display name localization

vitalii-koshovyi:
## What?
Fixed credit card display name localization

## Why?
Previously we used non-localized version due to the limitation of getPaymentMethodName method

## Testing / Proof
Manully
@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/payment/paymentMethod/PaymentMethodTitle.tsx
+++ b/packages/core/src/app/payment/paymentMethod/PaymentMethodTitle.tsx
@@ -215,6 +215,21 @@ function getPaymentMethodTitle(
             },
         };
 
+
+        if (method.gateway === PaymentMethodId.BlueSnapDirect) {
+            if (method.id === 'credit_card') {
+                return { logoUrl: '', titleText: language.translate('payment.credit_card_text') };
+            }
+
+            if (method.id === 'ecp') {
+                return { logoUrl: '', titleText: language.translate('payment.bluesnap_direct_electronic_check_label') };
+            }
+
+            if (method.id === 'banktransfer') {
+                return { logoUrl: '', titleText: language.translate('payment.bluesnap_direct_local_bank_transfer_label') };
+            }
+        }
+
         if (method.id === PaymentMethodId.PaypalCommerceVenmo) {
             return customTitles[PaymentMethodId.PaypalCommerceAlternativeMethod];
         }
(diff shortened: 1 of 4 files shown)


packages/core/src/app/payment/paymentMethod/PaymentMethodTitle.tsx
```diff
@@ -44,7 +44,7 @@ function getPaymentMethodTitle(
         const customTitles: { [key: string]: { logoUrl: string; titleText: string } } = {
             [PaymentMethodType.CreditCard]: {
                 logoUrl: '',
-                titleText: methodName,
```
bc-peng, line 47: I see they are not identical.
```
const methodName = getPaymentMethodName(language)(method);
const methodDisplayName = getPaymentMethodDisplayName(language)(method);
```
Could I know why did you replace it?

vitalii-koshovyi: methodName just grab method name from provider config.
methodDisplayName locally wraps credit card name in translation method.

vitalii-koshovyi: [methodName](https://github.com/bigcommerce/checkout-js/blob/master/packages/payment-integration-api/src/getPaymentMethodName.ts#L100)

        
[ methodDisplayName ](https://github.com/bigcommerce/checkout-js/blob/master/packages/core/src/app/payment/paymentMethod/getPaymentMethodDisplayName.tsx#L33)

animesh1987: Hey @vitalii-koshovyi they are still essentially different things. Can we not add text for bluesnap here https://github.com/bigcommerce/checkout-js/blob/master/packages/payment-integration-api/src/getPaymentMethodName.ts#L119. 

As a separate task we can think of how to consolidate the display name/method name into one. But that will require more testing across the app.

packages/core/src/app/payment/paymentMethod/PaymentMethodTitle.tsx
```diff
@@ -215,6 +215,21 @@ function getPaymentMethodTitle(
             },
         };
 
+
+        if (method.gateway === PaymentMethodId.BlueSnapDirect) {
```
animesh1987, line 219: Can we update test for these conditions?

vitalii-koshovyi: done

animesh1987 (approved):
Nice work 👏

PR #1459, opened 2023-08-25
fix(checkout): CHECKOUT-7656 Rectify vanishing shipping step

bc-peng:
## What?
Improve the conditions for throwing `ShippingOptionExpiredError`.

## Why?
After a shopper refreshes the checkout page, trying to remove a free shipping coupon doesn't work properly because the checkout fails to trigger the expected `ShippingOptionExpiredError`.

This PR fixes the problem above.

### BEFORE
 
https://github.com/bigcommerce/checkout-js/assets/88361607/45323035-479f-4880-babb-b4f8c642a333

### AFTER

https://github.com/bigcommerce/checkout-js/assets/88361607/374bf8f4-5c73-4945-bb92-511bf558a1f0

### DIFF
--- a/packages/core/src/app/checkout/Checkout.tsx
+++ b/packages/core/src/app/checkout/Checkout.tsx
@@ -608,7 +608,7 @@ class Checkout extends Component<
     };
 
     private handleConsignmentsUpdated: (state: CheckoutSelectors) => void = ({ data }) => {
-        const { hasSelectedShippingOptions: prevHasSelectedShippingOptions, activeStepType } =
+        const { hasSelectedShippingOptions: prevHasSelectedShippingOptions, activeStepType, defaultStepType } =
             this.state;
 
         const { steps } = this.props;
@@ -620,8 +620,17 @@ class Checkout extends Component<
         if (
             prevHasSelectedShippingOptions &&
             !newHasSelectedShippingOptions &&
-            findIndex(steps, { type: CheckoutStepType.Shipping }) <
-                findIndex(steps, { type: activeStepType })
+            (
+                findIndex(steps, { type: CheckoutStepType.Shipping }) <
+                findIndex(steps, { type: activeStepType }) ||
+                (
+                    !activeStepType &&
+                    (
+                        defaultStepType === CheckoutStepType.Payment ||
+                        defaultStepType === CheckoutStepType.Billing
+                    )
+                )
+            )
         ) {
             this.navigateToStep(CheckoutStepType.Shipping);
             this.setState({ error: new ShippingOptionExpiredError() });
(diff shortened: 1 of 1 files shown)


packages/core/src/app/checkout/Checkout.tsx
bc-peng, line 1: When the checkout page reloads and the default step is billing or shipping, `activeStepType` will always be `null`, despite the fact that `defaultStepType` can be any of them.

Previous condition `findIndex(steps, { type: CheckoutStepType.Shipping }) < findIndex(steps, { type: activeStepType })` only works when activeStepType is not `null`.

Therefore, add new conditions here to accommodate the page reload case.

bc-peng: I found after a page reload, `activeStepType` remains `null` until a step on the checkout page is completed manually .

bc-peng: I found after a page reload, activeStepType remains null until a step on the checkout page is completed manually .

Previous condition findIndex(steps, { type: CheckoutStepType.Shipping }) < findIndex(steps, { type: activeStepType }) only works when activeStepType is not null.

Therefore, add new conditions here to accommodate the page reload case.

bc-peng: I found after a page reload, `activeStepType` remains `null` until a step on the checkout page is completed manually .

Previous condition `findIndex(steps, { type: CheckoutStepType.Shipping }) < findIndex(steps, { type: activeStepType })` only works when activeStepType is not null.

Therefore, add new conditions here to accommodate the page reload case.

snaderiBC: Was this due to our recent changes? regression?

bc-peng: This logic was built 4 years ago in [this pr](https://github.com/bigcommerce/checkout-js/pull/149), which means the problem has been there for four years.
The root cause is after a page refresh, we don't set the value of `activeStepType` until a step is completed. I tested changing the way of setting `activeStepType`, and it turned out to be a very dangerous move.

snaderiBC (commented):
The change makes sense, thanks for explaining.
Can we add any test to cover this scenario?

animesh1987 (approved):

animesh1987:
Hey @snaderiBC this behaviour requires a e2e or a functional test to cover. We will need to investigate if we can solve this by moving our test suite to RTL. But since this is a csmed will say we can move ahead with this change.

bc-peng:
RTL focues on testing UI changes. I believe `playwright` would be a superior choice for testing UI changes below. Alternatively, testing it in Node seems less effective.
cc @animesh1987 @snaderiBC 
image

snaderiBC:
>I believe playwright would be a superior choice for testing UI changes below

@bc-peng I agree with you. I wasn't suggesting one over the other (RTL vs. PW)
I was merely suggesting we should add a test to cover this particular scenario that we're fixing.
Please raise a ticket for adding that test since this PR was merged.

PR #1468, opened 2023-08-28
feat(payment): PI-744 fix payment method rendering for mollie klarna

bc-wpietrzak:
## What?
Added additional condition for rendering Klarna payment method.

## Why?
...

## Testing / Proof
...

@bigcommerce/team-checkout


serhii-tkachenko (approved):
Looks good to me. Don't forget to cover this changes with tests

animesh1987:
Strange change, can I ask why are we returning klarna component if Id is for Mollie?

bc-wpietrzak:
As I explained in the comment, klarna is used as a `methodId` both for native Klarna and for One Klarna provided by Mollie. Since the only condition for rendering `KlarnaPaymentMethod` was klarna `methodId` (and the if statement for `KlarnaPaymentMethod` is above the statement for `MolliePaymentMethod`) I added a new condition to stop `KlarnaPaymentMethod` from rendering based solely on the methodId.

bc-wpietrzak:
Added tests

animesh1987 (approved):

serhii-tkachenko (approved):
Nice work!

PR #1469, opened 2023-08-29
fix(checkout): CHECKOUT-7606 Fix GoogleAutocomplete max length

richapatel1510:
## What?

Fix: Google Autocomplete does not respect Maximum Field Length set on Form Fields. Considered the max length from configuration for google autocomplete input field as well.

**Note:** If more characters are returned from Google autocomplete address itself then the entire address will be considered and max-length will be ignored in that case.

## Why?
To respect Maximum Field Length setting configured in the Control Panel when typing manual address in Google Autocomplete field.

## Testing / Proof

1. Unit test
2. Manual test

**BEFORE:**

https://github.com/bigcommerce/checkout-js/assets/141614330/d40e59d6-2d4c-4385-a2dd-82d575518681

**AFTER:**

https://github.com/bigcommerce/checkout-js/assets/141614330/42c5c478-5535-4f9d-b310-f4ff9b4dc191


packages/core/src/app/address/googleAutocomplete/GoogleAutocomplete.spec.tsx
```diff
@@ -19,4 +19,16 @@ describe('GoogleAutocomplete Component', () => {
 
         expect(toJson(tree)).toMatchSnapshot();
     });
+
+    it('renders input with max length', () => {
```
animesh1987, line 23: This seems like a small enough test file, can we convert the suite to RTL?

richapatel1510: @animesh1987 Converted it.

packages/core/src/app/address/googleAutocomplete/GoogleAutocomplete.test.tsx
```diff
@@ -0,0 +1,33 @@
...
+    it('renders input with max length', () => {
+        expect(
+            render(
+                <GoogleAutocomplete
+                    apiKey="bar"
+                    initialValue="fo"
+                    inputProps={{ maxLength: "50" }}
+                />,
+            ),
+        ).toMatchSnapshot();
```
bc-peng, line 31: Can we test `GoogleAutocomplete` in the way the user would use it?

animesh1987: Yes I will say probably need to avoid snapshot testing.

richapatel1510: @bc-peng @animesh1987 I tried mocking google maps service following some suggestions [here](https://github.com/hibiken/react-places-autocomplete/issues/189) and [here](https://stackoverflow.com/questions/58439971/how-can-i-test-google-maps-geocoder) but those only give a way to check if the service has been called `n` number of times and not give the actual results like we get from the maps service. Hence, I cannot add more tests to check if the services work fine and give correct results. However, I was able to add tests to check the scenario for max-length and placeholder texts using RTL and jest and avoid snapshot testing.

packages/core/src/app/address/googleAutocomplete/GoogleAutocomplete.test.tsx
```diff
@@ -0,0 +1,33 @@
...
+        render(<ComponentTest {...defaultProps} inputProps={{ placeholder: 'NO PO BOX' }} />);
+        expect(screen.getByPlaceholderText('NO PO BOX')).toBeInTheDocument();
+    });
+
+    it('renders input and limits value to max length', async () => {
+        render(<ComponentTest {...defaultProps} initialValue="" inputProps={{ placeholder: 'Max length 50', maxLength: "50" }} />);
+        expect(screen.getByPlaceholderText('Max length 50')).toBeInTheDocument();
+        await userEvent.type(screen.getByPlaceholderText('Max length 50'), '120 South Jean Baptiste Point du Sable Lake Shore Drive');
+        const inputElement: HTMLInputElement = screen.getByPlaceholderText('Max length 50');
+        expect(inputElement.value).toBe('120 South Jean Baptiste Point du Sable Lake Shore ');
```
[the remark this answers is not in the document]
richapatel1510: @bc-peng Fixed those.

packages/core/src/app/address/googleAutocomplete/GoogleAutocomplete.test.tsx
```diff
@@ -0,0 +1,33 @@
+import '@testing-library/jest-dom';
```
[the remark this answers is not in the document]
bc-peng: The purpose of `import '@testing-library/jest-dom';` in `jest-setup.ts` is to avoid importing it in every single test file. Could you please check what is missing there?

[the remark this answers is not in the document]
bc-peng: I didn't see any error after removing this line in both IDE and terminal.
image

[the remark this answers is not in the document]
richapatel1510: @bc-peng  I get this kind of warnings when I remove that line.
Screenshot 2023-09-15 at 11 23 17 am

animesh1987 (approved):

bc-peng (approved):

PR #1472, opened 2023-08-29
feat(payment): PAYPAL-2728 create shipping strategy for Braintree AXO

PavlenkoM:
## What?
Create shipping strategy for Braintree Accelerated checkout

## Why?
To check logged in user and show OTP modal for PayPal Connect users

## Testing / Proof
Unit test and manually tested

https://github.com/bigcommerce/checkout-sdk-js/assets/9430298/dfb0a7d0-8186-4744-bccc-f8a41b765652

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/shipping/ShippingForm.spec.tsx
+++ b/packages/core/src/app/shipping/ShippingForm.spec.tsx
@@ -1,25 +1,28 @@
-import { CheckoutService, createCheckoutService } from '@bigcommerce/checkout-sdk';
+import { CheckoutService, createCheckoutService, CustomerAddress } from '@bigcommerce/checkout-sdk';
 import { mount, ReactWrapper } from 'enzyme';
 import React from 'react';
 
 import { ExtensionProvider } from '@bigcommerce/checkout/checkout-extension';
 import { createLocaleContext, LocaleContext, LocaleContextType } from '@bigcommerce/checkout/locale';
 import { CheckoutProvider } from '@bigcommerce/checkout/payment-integration-api';
+import { getShippingAddress } from '@bigcommerce/checkout/test-utils';
 
+import * as usePayPalConnectAddress from '../address/PayPalAxo/usePayPalConnectAddress';
 import { getCart } from '../cart/carts.mock';
 import { getPhysicalItem } from '../cart/lineItem.mock';
 import { getStoreConfig } from '../config/config.mock';
 import { getCustomer } from '../customer/customers.mock';
 import { getCountries } from '../geography/countries.mock';
 import { OrderComments } from '../orderComments';
+import { PaymentMethodId } from '../payment/paymentMethod';
 
 import BillingSameAsShippingField from './BillingSameAsShippingField';
 import { getConsignment } from './consignment.mock';
 import MultiShippingForm from './MultiShippingForm';
-import { getShippingAddress } from './shipping-addresses.mock';
 import ShippingAddress from './ShippingAddress';
 import ShippingForm, { ShippingFormProps } from './ShippingForm';
 import { ShippingOptions } from './shippingOption';
+import SingleShippingForm from './SingleShippingForm';
 
 describe('ShippingForm Component', () => {
     let component: ReactWrapper;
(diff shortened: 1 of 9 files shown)


packages/core/src/app/shipping/ShippingForm.spec.tsx
```diff
@@ -358,4 +368,117 @@ describe('ShippingForm Component', () => {
             });
         });
     });
+
+    describe('Braintree AXO', () => {
+        afterEach(() => {
+            jest.clearAllMocks();
+        });
```
serhii-tkachenko, line 375: I think this one can be moved to the first describe level (below beforeEach function)

packages/core/src/app/shipping/ShippingForm.spec.tsx
```diff
@@ -358,4 +368,117 @@ describe('ShippingForm Component', () => {
...
+                expect.objectContaining({
+                    addresses: payPalConnectAddresses,
+                }),
+            );
+            expect(initializeMock).toHaveBeenCalledWith({
+                methodId: PaymentMethodId.BraintreeAcceleratedCheckout,
+            });
+        });
+
+        it('renders when PayPal AXO is enabled but shipping method Id is different', () => {
```
serhii-tkachenko, line 444:
```suggestion
        it('renders shipping for with paypal connect addresses without strategy initialisation if there is preselected shipping method id', () => {
```

serhii-tkachenko (approved):
LGTM

packages/core/src/app/shipping/ShippingForm.spec.tsx
PavlenkoM: good point, thanks

packages/core/src/app/shipping/getShippingMethodId.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+    it('returns shipping method id from preselected payment method', () => {
+        const [checkout, config] = mockOptions(
+            PaymentMethodId.BraintreeAcceleratedCheckout,
+            PaymentMethodId.AmazonPay,
+        );
+
+        expect(getShippingMethodId(checkout, config)).toBe(PaymentMethodId.AmazonPay);
+    });
+});
```
animesh1987, line 62: EOL ?

packages/core/src/app/shipping/ShippingForm.tsx
```diff
@@ -14,11 +14,12 @@ import {
     ShippingInitializeOptions,
     ShippingRequestOptions,
 } from '@bigcommerce/checkout-sdk';
-import React from 'react';
+import React, { useEffect } from 'react';
 
 import { withLanguage, WithLanguageProps } from '@bigcommerce/checkout/locale';
 
 import { usePayPalConnectAddress } from '../address/PayPalAxo';
+import { isBraintreeConnectPaymentMethod } from '../payment';
```
animesh1987, line 22: I think we need to move this method to some common place. As now we are starting to create a coupling between payment and shipping folders.

PavlenkoM: maybe we can do this in separate PR? Because it will effect other files which are not related to current changes and will be better to split this from  PRs?

cc: @serhii-tkachenko

packages/core/src/app/shipping/getShippingMethodId.test.tsx
PavlenkoM: done

packages/core/src/app/shipping/ShippingForm.tsx
serhii-tkachenko: I will move it in PayPal Connect integration package

packages/core/src/app/shipping/getShippingMethodId.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+        );
+
+        return [checkoutDefault, config];
+    }
+
+    afterEach(() => {
+        jest.clearAllMocks();
+    });
+    it('returns no shipping method ID if no preselected payment method and no provider with custom checkout', () => {
```
bc-peng, line 42:
```suggestion
    });
    
    it('returns no shipping method ID if no preselected payment method and no provider with custom checkout', () => {
```

bc-peng (approved):
LGTM. 👏

animesh1987 (approved):

PR #1474, opened 2023-08-30
fix(checkout): CHECKOUT-5966 Enable multishipping for more than 50 items in cart

richapatel1510:
## What?

Remove top limit of 50 cart items to show multi-shipping.

## Why?

Ship to multiple address should be visible if cart items is greater than 1 and no upper limit.

## Testing / Proof

1. Unit test
2. Manual test

**BEFORE**

https://github.com/bigcommerce/checkout-js/assets/141614330/79b5c819-334f-4f75-879c-b8a3d028f861

**AFTER**

https://github.com/bigcommerce/checkout-js/assets/141614330/9da79de4-28a1-48e0-9752-e168aa7fc626

### DIFF
--- a/packages/core/src/app/shipping/Shipping.tsx
+++ b/packages/core/src/app/shipping/Shipping.tsx
@@ -383,7 +383,7 @@ export function mapToShippingProps({
         isUpdatingCheckout() ||
         isCreatingCustomerAddress();
     const shouldShowMultiShipping =
-        hasMultiShippingEnabled && !methodId && shippableItemsCount > 1 && shippableItemsCount < 50;
+        hasMultiShippingEnabled && !methodId && shippableItemsCount > 1;
     const countriesWithAutocomplete = ['US', 'CA', 'AU', 'NZ'];
 
     if (features['CHECKOUT-4183.checkout_google_address_autocomplete_uk']) {
--- a/packages/core/src/app/config/config.mock.ts
+++ b/packages/core/src/app/config/config.mock.ts
@@ -20,7 +20,7 @@ export function getStoreConfig(): StoreConfig {
             enableTermsAndConditions: false,
             googleRecaptchaSitekey: 'sitekey',
             googleMapsApiKey: '',
-            hasMultiShippingEnabled: true,
+            hasMultiShippingEnabled: false,
             guestCheckoutEnabled: true,
             isAnalyticsEnabled: true,
             isAccountCreationEnabled: true,
--- a/packages/core/src/app/shipping/Shipping.spec.tsx
+++ b/packages/core/src/app/shipping/Shipping.spec.tsx
@@ -349,6 +349,27 @@ describe('Shipping Component', () => {
         expect(checkoutService.deleteConsignment).not.toHaveBeenCalled();
     });
 
+    it('shows multi address shipping link for more than 50 cart items', async () => {
+        jest.spyOn(checkoutState.data, 'getCart').mockReturnValue({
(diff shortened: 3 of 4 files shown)


snaderiBC (commented):
Thanks @richapatel1510 
Can you add a unit test please?

richapatel1510:
> Thanks @richapatel1510 Can you add a unit test please?

@snaderiBC Added unit test. Thanks!

animesh1987 (approved):

snaderiBC (approved):

snaderiBC (commented):
Thanks for adding the unit test 👏 
As a general rule any "fix" PR should have a unit test

bc-peng (approved):

bc-peng (changes requested):
Could you please move this test to `#when multishipping mode is on` > `when shopper is signed` group?

richapatel1510:
> Could you please move this test to `#when multishipping mode is on` > `when shopper is signed` group?

@bc-peng That `#when multishipping mode is on` is the case when `Ship to multiple addresses` toggle is already clicked and UI is in multishipping mode. The case that we covered in this PR is specifically to check if that toggle is present or not, and that's why I think it wont be a part of `#when multishipping mode is on` group.

packages/core/src/app/shipping/Shipping.tsx
```diff
@@ -383,7 +383,7 @@ export function mapToShippingProps({
         isUpdatingCheckout() ||
         isCreatingCustomerAddress();
     const shouldShowMultiShipping =
-        hasMultiShippingEnabled && !methodId && shippableItemsCount > 1 && shippableItemsCount < 50;
+        hasMultiShippingEnabled && !methodId && shippableItemsCount > 1;
```
bc-peng, line 386: @richapatel1510 As you can see here, multishipping mode has to be on to display this button while the shopper's sign-in state does not matter, hence the test should be inside the `when multishipping mode is on` group.
Maybe we can put the test at top of this group.

bc-peng: In addtition, `hasMultiShippingEnabled` comes from the checkoutSettings, not a user action.

richapatel1510: @bc-peng In the group `when multishipping mode is on` the component is `component = mount(<ComponentTest {...defaultProps} isMultiShippingMode={true} />);` that is the case when the UI is in multishipping mode. That is not same as `hasMultiShippingEnabled` flag.

bc-peng: I see your point, which would be valid if we always display 'Ship to multiple addresses' toggle as part of the shipping component.
However, the toggle does require `hasMultiShippingEnabled` to show up on the page.
Also, in order to set `isMultiShippingMode` to be true, `hasMultiShippingEnabled` needs to be true.

richapatel1510: `hasMultiShippingEnabled` is true by default in `packages/core/src/app/config/config.mock.ts` mock json.

bc-peng: Then we should change it to `false` in order to keep it in line with our production environment. 😂

richapatel1510: @bc-peng Made the changes to other tests as well that were written based on default value of that flag.

bc-peng (approved):

PR #1489, opened 2023-09-06
fix(payment): PAYPAL-2859 fixed the issue with wallet buttons on top rendering process

serhii-tkachenko:
## What?
In short: Fixed the issue with wallet buttons on top rendering process.

We had a condition to remove (instead of hiding) all buttons containers to avoid some issues with PayPal Commerce  conflicts between paywall and buttons. This is related for cases when PayPalCommerce enabled on the store. 

So we updated the logic and we remove only PayPal Commerce buttons instead of removing whole container.

## Why?
Some customers get an error like: '**Unable to create sign-in button without valid container ID**'. 

The error occurs in one scenario:
When the store has PayPalCommerce enabled on the store and the customer goes to payment step and refresh the page.

## Testing / Proof
Unit tests
Manual tests

Before:

https://github.com/bigcommerce/checkout-js/assets/25133454/8d257565-e346-4142-96db-f0f540a3486c

After:

https://github.com/bigcommerce/checkout-js/assets/25133454/690fc3aa-486a-445a-8e2a-4abd078a70b8

### DIFF
--- a/packages/core/src/app/customer/CheckoutButtonContainer.tsx
+++ b/packages/core/src/app/customer/CheckoutButtonContainer.tsx
@@ -46,9 +46,7 @@ const sortMethodIds = (methodIds:string[]): string[] => {
     return methodIds.sort((a, b) => order.indexOf(b) - order.indexOf(a));
 }
 
-const isPayPalCommerce = (methodId: string): boolean => {
-    return paypalCommerceIds.includes(methodId);
-}
+const isPayPalCommerce = (methodId: string): boolean => paypalCommerceIds.includes(methodId);
 
 const CheckoutButtonContainer: FunctionComponent<CheckoutButtonContainerProps & WithCheckoutCheckoutButtonContainerProps> = (
     {
--- a/packages/core/src/app/customer/CheckoutButtonContainer.tsx
+++ b/packages/core/src/app/customer/CheckoutButtonContainer.tsx
@@ -46,6 +46,10 @@ const sortMethodIds = (methodIds:string[]): string[] => {
     return methodIds.sort((a, b) => order.indexOf(b) - order.indexOf(a));
 }
 
+const isPayPalCommerce = (methodId: string): boolean => {
+    return paypalCommerceIds.includes(methodId);
+}
+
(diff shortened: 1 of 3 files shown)


packages/core/src/app/customer/CheckoutButtonContainer.tsx
```diff
@@ -63,11 +67,11 @@ const CheckoutButtonContainer: FunctionComponent<CheckoutButtonContainerProps &
         return null;
     }
 
-    if (isPaypalCommerce && isPaymentStepActive) {
-        return null;
-    }
-
     const renderButtons = () => availableMethodIds.map((methodId) => {
+        if (isPaymentStepActive && paypalCommerceIds.includes(methodId)) {
```
bc-peng, line 71: Could you please convert `paypalCommerceIds.includes(methodId)` into  `isPaypalCommerce(methodId)` for readability and place it next to `sortMethodIds()`?

animesh1987 (approved):

packages/core/src/app/customer/CheckoutButtonContainer.tsx
serhii-tkachenko: Sure, I will update the PR with second commit message

packages/core/src/app/customer/CheckoutButtonContainer.tsx
```diff
@@ -46,6 +46,10 @@ const sortMethodIds = (methodIds:string[]): string[] => {
     return methodIds.sort((a, b) => order.indexOf(b) - order.indexOf(a));
 }
 
+const isPayPalCommerce = (methodId: string): boolean => {
+    return paypalCommerceIds.includes(methodId);
+}
+
```
animesh1987, line 52: const isPayPalCommerce = (methodId: string): boolean => paypalCommerceIds.includes(methodId);

animesh1987 (approved):

packages/core/src/app/customer/CheckoutButtonContainer.tsx
serhii-tkachenko: Oopss, sorry for that

PR #1493, opened 2023-09-08
test(checkout): CHECKOUT-7685 e2e tests for adding/removing coupon

richapatel1510:
## What?

Added e2e test for adding and removing coupon code on cart after shipping step is submitted.

## Why?

Add e2e tests to cover issue of Shipping step getting vanished after removing coupon; fixed in https://github.com/bigcommerce/checkout-js/pull/1459

## Testing / Proof
[Promotion-e2e.webm](https://github.com/bigcommerce/checkout-js/assets/141614330/2ec33b64-0eaa-4914-91d3-b004447631f4)


packages/core/e2e/Promotion.spec.ts
```diff
@@ -0,0 +1,30 @@
+import { AddCouponPreset, PaymentStepAsGuestPreset, test } from '@bigcommerce/checkout/test-framework';
+
+test.describe('Shipping with coupon', () => {
+    test('`Adding coupon` after shipping step is submitted shows shipping error modal', async ({ assertions, checkout, page }) => {
```
bc-peng, line 4: Adding a coupon after shipping step does not always result in an error modal showing up.
Perhaps we should mention it is a free-shipping coupon.

packages/test-framework/src/fixture/pageObject/Assertions.ts
```diff
@@ -27,4 +27,14 @@ export class Assertions {
...
     }
+
+    async shouldSeeErrorModal(): Promise<void> {
+        const page = this.page;
+
+        await page.locator('data-test=modal-body').waitFor({ state: 'visible' });
+        await expect(page.locator('data-test=modal-body')).toBeVisible();
+        await expect(page.locator('#errorModalMessage')).toHaveText(
+            'The shipping price you were quoted is no longer valid. Click OK to see the most up-to-date shipping prices.',
+        );
```
bc-peng, line 38: Please make the text as a parameter so that this method can be applied to different error cases.

packages/test-framework/src/fixture/pageObject/Checkout.ts
```diff
@@ -86,4 +86,14 @@ export class Checkout {
...
+
+    async applyCoupon(couponCode: string): Promise<void> {
+        await this.page.locator('[data-test="cart"] [data-test="redeemable-label"]').click();
+        await this.page.locator('[data-test="redeemableEntry-input"]').fill(couponCode);
+        await this.page.locator('[data-test="redeemableEntry-submit"]').click();
+    }
+
+    async removeCoupon(): Promise<void> {
+        await this.page.locator('[data-test="cart-price-callback"]').click();
+    }
```
bc-peng, line 98: Is this still working when there are multiple coupons?

packages/test-framework/src/fixture/pagePreset/AddCouponPreset.ts
```diff
@@ -0,0 +1,16 @@
+import { Page } from '@playwright/test';
+
+import { ApiRequestsSender } from './ApiRequestsSender';
+import { CheckoutPagePreset } from './CheckoutPagePreset';
+
+export class AddCouponPreset implements CheckoutPagePreset {
```
bc-peng, line 6: I noticed that this preset not only adds a coupon but also performs all the actions of the `PaymentStepAsGuestPreset`. This goes against our goal of encouraging developers to use multiple presets in a single test, like this:

```javascript
await checkout.use(new UseAUDPreset('AUD'));
```
In this situation, I think it's better to use `await checkout.applyCoupon('FREESHIP')` because depending on a hard-coded coupon code seems to restrict the use of a preset.

packages/test-framework/src/fixture/pageObject/Checkout.ts
richapatel1510: Yes, its working because at a time it allows to only add one coupon as per coupon UI.

packages/test-framework/src/fixture/pagePreset/AddCouponPreset.ts
richapatel1510: @bc-peng Removed the ApplyCouponPreset and added condition to apply coupon in `PaymentStepAsGuestPreset` itself. Cannot use `await checkout.applyCoupon('FREESHIP')` because then we will get the modal popup again we will need to close that and continue the shipping step.

packages/test-framework/src/fixture/pageObject/Checkout.ts
bc-peng: Becuase this method won't work if a store allows applying multi coupons at a time, I recommend that we leave it out of the test framework and instead use `await this.page.locator('[data-test="cart-price-callback"]').click();`.In the test directly.

packages/test-framework/src/fixture/pagePreset/AddCouponPreset.ts
bc-peng: Ideally, I think we want a coupon preset like this:
```
await checkout.use(new ApplyCouponPreset('COUPON CODE'));
```
So `PaymentStepAsGuestPreset`'s purpose is clear. Applying a coupon can be combined into a test case.
Your test case is special, the coupon will trigger a modal to impact the shipping selection.
This is not a general case, does it make sense to use playwright actions like `await checkout.applyCoupon('FREESHIP')`(and more actions) to setup the coupon?

richapatel1510: `await checkout.use(new PaymentStepAsGuestPreset());
await checkout.use(new ApplyCouponPreset('COUPON CODE'));`

Using above preset will open modal popup at the very beginning because shipping details are changed after Payment step is reached, hence not able to test the remove coupon.

And other way is If we write actions to setup the coupon then we won't get the warning while removing coupon becuase the shipping step needs to be updated after applying coupon and hence no warning after removing coupon. 

Moreover, similar to how completing customer, shipping and billing section is part of PaymentStepAsGuest, adding a coupon can also be part of that process as well because there are scenarios where user will apply coupon before reaching Payments step to check for accurate shipping quote.

bc-peng: >Using above preset will open modal popup at the very beginning because shipping details are changed after Payment step is reached, hence not able to test the remove coupon.

This coupon triggers an error modal. Building a preset for this coupon doesn’t feel like a reusable solution.

That's why I think we'd prefer playwright actions for this case.

Perhaps we can combine the current two e2e tests into one, which starts at `PaymentStepAsGuestPreset`, then using playwright actions to:
1. apply the coupon;
2. select the free shipping method which is generated by the coupon;
3. come back to the payment step;
4. refresh page;
5. remove coupon and assert the error modal.

>And other way is If we write actions to setup the coupon then we won't get the warning while removing coupon

This can be avoided by configuring the store shipping to ensure `PaymentStepAsGuestPreset` ends up with a paid shipping method.

richapatel1510: Updated the code to cover both scenarios in single test and to accomodate above suggestion.

packages/core/e2e/Promotion.spec.ts
bc-peng, line 1: Your code is quite self-explanatory.
Next time, simply use blank lines to group your code.
🏆  Kudos to you on your first `playwright` test.

bc-peng (approved):

animesh1987 (approved):
Nice work 👏

PR #1506, opened 2023-09-18
refactor(checkout): CHECKOUT-7450 remove language HOC from locale package

richapatel1510:
## What?

- Removed all occurrences of the `withLanguage` HOC inside the `locale` package. 
- Updated all the required snapshots and unit tests for modified components.

## Why?

`checkout-js` can move away from HOC-pattern

## Testing / Proof

- CI checks


packages/locale/src/TranslatedHtml.spec.tsx
```diff
@@ -4,11 +4,11 @@ import React from 'react';
 
 import { getStoreConfig } from '@bigcommerce/checkout/test-utils';
 
-import LocaleContext, { LocaleContextType } from './LocaleContext';
-import TranslatedHtml from './TranslatedHtml';
+import * as LocaleContext from './LocaleContext';
```
bc-peng, line 7:
```suggestion
import LocaleContext, { LocaleContextType } from './LocaleContext';
```
Can we avoid `import *` as we prefer importing things explicitly?

packages/locale/src/TranslatedString.tsx
```diff
@@ -1,16 +1,29 @@
...
+    return <>{language.translate(id, data)}</>;
+};
+
+const TranslatedString: FunctionComponent<TranslatedStringProps> = (props) => {
+    const localeContextValue: LocaleContextType = {
+        language: getLanguageService(),
+    };
+
+    return (
+        <LocaleContext.Provider value={localeContextValue}>
```
bc-peng, line 23: Is there any special reason for putting `<LocaleContext.Provider value={localeContextValue}>` here?
We have `useLocale();` to check if this component is inside a `localeContext`, hence here we are applying another layer of the same context.

richapatel1510: @bc-peng To test the changes for this requirement we only ran the unit tests. Now if we don't wrap it with the provider then several tests were failing. 
Screenshot 2023-09-25 at 7 15 56 pm

I added a provider to this file because the provider just needed language service which is common throughout and does not change.

What do you suggest? All those tests should be wrapped with the provider? And how do we make sure all those integration packages will work fine without wrapping `TranslatedString` with provider?

bc-peng: We have applied the locale context at the very top.
https://github.com/bigcommerce/checkout-js/blob/a255f2ab02ef0178b3bc0bb2b0977829625f8b73/packages/core/src/app/checkout/CheckoutApp.tsx#L59

e2e tests inside each integration package ensure the package will work fine.

I think you are adding the context again for the testing purpose, could we find out how many tests will be impacted here if we don't do so?

richapatel1510: Screenshot 2023-09-26 at 1 57 05 pm

@bc-peng I have ran unit tests here `npm run test:core` and we have 52 failures there.

vitalii-koshovyi (approved):
LGTM!

packages/core/src/app/order/OrderSummaryModal.spec.tsx
```diff
@@ -45,11 +48,12 @@ describe('OrderSummaryModal', () => {
                 customItems: [],
             };
 
-            lineItems.physicalItems.push(getPhysicalItem(), {...getPhysicalItem(), id: '888', parentId: 'test'})
+            lineItems.physicalItems.push(getPhysicalItem(), { ...getPhysicalItem(), id: '888', parentId: 'test' })
 
             const order = { ...getOrder(), lineItems };
 
             orderSummary = shallow(
+                // <LocaleProvider checkoutService={checkoutService}>
```
bc-peng, line 56:
```suggestion
```

packages/core/src/app/order/OrderSummaryModal.spec.tsx
```diff
@@ -59,10 +63,11 @@ describe('OrderSummaryModal', () => {
                     storeCurrency={getStoreConfig().currency}
                     total={order.orderAmount}
                 />,
+                // </LocaleProvider>,
```
bc-peng, line 66:
```suggestion
```

bc-peng (approved):
Kudos to you. Such a large PR.🎉

animesh1987 (approved):

serhii-tkachenko (approved):
Nice work!

PR #1520, opened 2023-09-26
test(checkout): CHECKOUT-7488 Add e2e test to cover shipping area of checkout

richapatel1510:
## What?

Added e2e test for guest checkout, customer checkout and multi address shipping checkout.

## Why?

To cover shipping area of checkout with e2e tests, so that our tests can give us more confidence.

## Testing / Proof
[5d09438de7273d08779592824ea6cc74.webm](https://github.com/bigcommerce/checkout-js/assets/141614330/4cc25537-dd84-482f-ac5c-8558999a0f97)
[8c7fc28524d72055c611f927db1f991e.webm](https://github.com/bigcommerce/checkout-js/assets/141614330/e792a9a9-07a2-41fe-8744-ccf3a75a9dea)
[dd2ace437f170f8f195af5e9a9de18cb.webm](https://github.com/bigcommerce/checkout-js/assets/141614330/415c32c0-db8d-4c75-9b36-6fe84d98f516)
[c1a6377d2595dc982017212a70f93a72.webm](https://github.com/bigcommerce/checkout-js/assets/141614330/1e035fb9-1354-425b-92be-7baea29498b1)


packages/test-framework/src/fixture/pageObject/Checkout.ts
```diff
@@ -92,4 +92,152 @@ export class Checkout {
         await this.page.locator('[data-test="redeemableEntry-input"]').fill(couponCode);
         await this.page.locator('[data-test="redeemableEntry-submit"]').click();
     }
+
+    async completeCustomerStepAsGuest(): Promise<void> {
+        const email = `test@example.com`;
```
bc-peng, line 97: Can we make it as a property with the default value as `test@example.com`?

packages/test-framework/src/fixture/pageObject/Checkout.ts
```diff
@@ -92,4 +92,152 @@ export class Checkout {
...
+    }
+
+    async fillAddressForm({
+        formId,
+        addressAppendText = '',
+    }: {
+        formId: string;
+        addressAppendText: string;
+    }): Promise<void> {
+        const address = {
```
bc-peng, line 129: Can we make the address data as a prop so that devs can use their own address for testing when they try to `completeShippingAddressForm`?

packages/test-framework/src/fixture/pageObject/Checkout.ts
```diff
@@ -92,4 +92,152 @@ export class Checkout {
...
+            .locator('.ReactModalPortal .modal.optimizedCheckout-contentPrimary')
+            .waitFor({ state: 'visible' });
+        await this.fillAddressForm({
+            formId: '[data-test="modal-body"] .checkout-address',
+            addressAppendText,
+        });
+        await this.page.locator('#checkout-save-address').click();
+    }
+
+    async completeMultiShippingAddressForm(): Promise<void> {
```
bc-peng, line 220: A shopper can also finish the multishipping form by selecting an existing address?
Can we cover that use case? 
Meanwhile, perhaps we can rename this method as `completeMultiShippingWithTwoNewAddresses`.

packages/test-framework/src/fixture/pagePreset/AddItemToCart.ts
```diff
@@ -0,0 +1,19 @@
+import { Page } from '@playwright/test';
+
+import { ApiRequestsSender } from './ApiRequestsSender';
+import { CheckoutPagePreset } from './CheckoutPagePreset';
+
+export class AddItemToCart implements CheckoutPagePreset {
```
bc-peng, line 6: This preset name may confuse devs because it still ends at `PaymentStepAsGuest`.
The additional call in `PaymentStepAsGuestPreset` is to ensure the wallet buttons will be presented.
Therefore, can we combine the two and use a toggle to control the addtional call?

packages/test-framework/src/fixture/pageObject/Checkout.ts
richapatel1510: I changed existing test case's second item to have select address from saved address.

packages/test-framework/src/fixture/pagePreset/CustomerStepAsGuest.ts
```diff
@@ -4,11 +4,23 @@ import { ApiRequestsSender } from './ApiRequestsSender';
...
-        await api.addPhysicalItemToCart();
-        await api.setShippingQuote();
-        await api.dispose('You should be able to see checkout at the customer step now');
+        await api.addPhysicalItemToCart(this.quantity);
+        await api.dispose(`Item(s) is added to cart now`);
+
+        if (this.setShippingQuote) {
+            await api.setShippingQuote();
+            await api.dispose('You should be able to see checkout at the customer step now');
+        }
```
bc-peng, line 24: `dispose` method should be only called once in an API context's lifecycle.

packages/test-framework/src/fixture/pageObject/Checkout.ts
```diff
@@ -92,4 +93,143 @@ export class Checkout {
...
+            .waitFor({ state: 'visible' });
+        await this.fillAddressForm({
+            formId: '[data-test="modal-body"] .checkout-address',
+            address,
+            addressAppendText,
+        });
+        await this.page.locator('#checkout-save-address').click();
+    }
+
+    async completeMultiShippingAddressForm(address: Address): Promise<void> {
```
bc-peng, line 211: I think a developer can hardly figure out what the cart and account should look like to use this method.
Also, there are too many ways to complete a multi shipping form.
Perhaps we should put playwright actions directly in the test.

packages/core/e2e/Shipping.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+        await page.locator(`#billingAddresses .dropdown--select`).click();
+        await page.locator(`#billingAddresses .dropdownMenu`).waitFor({ state: 'visible' });
+        await page.locator(`#billingAddresses .dropdownMenu li`).nth(1).click();
+
+        await page.locator('#checkout-billing-continue').click();
+
+        // Assertions
+        await assertions.shouldSeePaymentStep();
+    });
+});
```
bc-peng, line 101: Missing EOF here. The rest looks good to me.

bc-peng (approved):
Nicely done. 👏👏

animesh1987 (approved):

PR #1521, opened 2023-09-26
feat(payment): PAYPAL-2946 added loading spinner to ratepay init options

andriiVitvitskyi1990:
## What?
Added loading spinner to ratepay init options

## Why?
To show loading process when polling mechanism is occurs

## Testing / Proof

272492002-0a24607c-c4a1-425e-8a90-ad77ef0baf43

@bigcommerce/team-checkout @bc-peng @animesh1987

### DIFF
--- a/packages/core/src/scss/components/bigcommerce/loading/_loading.scss
+++ b/packages/core/src/scss/components/bigcommerce/loading/_loading.scss
@@ -64,3 +64,16 @@
         width: 0;
     }
 }
+
+.embedded-checkout-loading-spinner-overlay {
+    position: fixed;
+    top: 0;
+    left: 0;
+    background-color: $loading-spinner-overlay;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    width: 100%;
+    height: 100%;
+    z-index: 999;
+}
--- a/packages/paypal-commerce-integration/src/PaypalCommerceRatePayPaymentMethod.tsx
+++ b/packages/paypal-commerce-integration/src/PaypalCommerceRatePayPaymentMethod.tsx
@@ -1,4 +1,4 @@
-import React, { FunctionComponent, useCallback, useEffect, useMemo, useRef } from 'react';
+import React, { FunctionComponent, useCallback, useEffect, useMemo, useRef, useState } from 'react';
 
 import {
     PaymentMethodProps,
@@ -8,6 +8,7 @@ import {
 import { DynamicFormField, DynamicFormFieldType, FormContext } from '@bigcommerce/checkout/ui';
 import { FormField } from '@bigcommerce/checkout-sdk';
 import getPaypalCommerceRatePayValidationSchema from './validation-schemas/getPaypalCommerceRatePayValidationSchema';
+import { LoadingSpinner } from '@bigcommerce/checkout/ui';
 
 interface RatePayFieldValues {
     ratepayBirthDate: {
@@ -66,6 +67,7 @@ const PaypalCommerceRatePayPaymentMethod: FunctionComponent<any> = ({
 }) => {
     const fieldsValues = useRef<Partial<RatePayFieldValues>>({});
     const isPaymentDataRequired = checkoutState.data.isPaymentDataRequired();
+    const [isPaymentSubmitting, setIsPaymentSubmitting] = useState(false);
 
     if (!isPaymentDataRequired) {
         return null;
@@ -80,6 +82,7 @@ const PaypalCommerceRatePayPaymentMethod: FunctionComponent<any> = ({
                     container: '#checkout-payment-continue',
                     legalTextContainer: 'legal-text-container',
                     getFieldsValues: () => fieldsValues.current,
+                    onPaymentSubmission: (isSubmitting: boolean) => setIsPaymentSubmitting(isSubmitting),
                     onError: (error: Error) => {
                         paymentForm.disableSubmit(method, true);
                         onUnhandledError(error);
@@ -140,6 +143,11 @@ const PaypalCommerceRatePayPaymentMethod: FunctionComponent<any> = ({
 
     return (
         <div style={{marginBottom:'20px'}}>
+            { isPaymentSubmitting &&
+                <div className='embedded-checkout-loading-spinner-overlay'>
+                    <LoadingSpinner isLoading={true}/>
+                </div>
+            }
             <FormContext.Provider value={{ isSubmitted, setSubmitted }}>
                 { formFieldData.map((field) => {
                         return  <DynamicFormField
--- a/packages/core/src/scss/settings/global/overlay/_overlay.scss
+++ b/packages/core/src/scss/settings/global/overlay/_overlay.scss
@@ -11,3 +11,4 @@
 // -----------------------------------------------------------------------------
 
 $overlay-fill:                  rgba($color-white, 0.9);
+$loading-spinner-overlay:       rgba($color-black, .3);
--- a/packages/paypal-commerce-integration/src/PaypalCommerceRatePayPaymentMethod.test.tsx
+++ b/packages/paypal-commerce-integration/src/PaypalCommerceRatePayPaymentMethod.test.tsx
@@ -71,6 +71,7 @@ describe('PaypalCommerceRatePayPaymentMethod', () => {
             gatewayId: props.method.gateway,
             methodId: props.method.id,
             paypalcommerceratepay: {
+                onPaymentSubmission: expect.any(Function),
                 container: '#checkout-payment-continue',
                 legalTextContainer: 'legal-text-container',
                 getFieldsValues: expect.any(Function),
--- a/packages/ui/src/index.ts
+++ b/packages/ui/src/index.ts
@@ -63,6 +63,6 @@ export {
     IconUsdCoin,
     IconAch,
 } from './icon';
-export { LoadingOverlay } from './loading';
+export { LoadingOverlay, LoadingSpinner } from './loading';
 export { Modal, ModalHeader, ModalTrigger, ModalTriggerModalProps } from './modal';
 export { TooltipTrigger } from './tooltip';


packages/paypal-commerce-integration/src/PaypalCommerceRatePayPaymentMethod.tsx
```diff
@@ -79,6 +80,7 @@ const PaypalCommerceRatePayPaymentMethod: FunctionComponent<any> = ({
                 paypalcommerceratepay: {
                     container: '#checkout-payment-continue',
                     legalTextContainer: 'legal-text-container',
+                    loadingSpinner: ()  => LoadingSpinnerAnimation({}),
```
serhii-tkachenko, line 83: I don't really sure that we should pass the whole component to the checkout sdk strategy and combine UI component logic between 2 parts (overlay on `checkout-sdk` side, loading component on `checkout-js` side)... 

`checkout-js` should be responsible for UI part of the implementation (to decide what should be done on UI side when the process is in pending state),  while `checkout-sdk` should tell `checkout-js` when to trigger that implementation.

So in this case my proposal is:
- to pass onPending callback to checkout-sdk and trigger it there whenever needed;
- implement the loader logic with overlay and spinner on `checkout-js` side.

It will help to keep our code simple to understand. Also we can describe in checkout-sdk what `onPending(isPending: true|false)` callback used for.

Whad do you think? @andriiVitvitskyi1990 @animesh1987 @bc-peng

serhii-tkachenko (commented):
I don't really sure that it is a right way to implement such feature

packages/paypal-commerce-integration/src/PaypalCommerceRatePayPaymentMethod.tsx
serhii-tkachenko: Yes. So when you call onPending callback in checkout sdk strategy, then checkout-js should show the loader to the customer. It will give other developers some flexibility for cases if they want to log something or do something else, when transaction is in pending status.

We should move loading UI responsibilities to checkout-js

packages/paypal-commerce-integration/src/PaypalCommerceRatePayPaymentMethod.tsx
```diff
@@ -79,6 +82,7 @@ const PaypalCommerceRatePayPaymentMethod: FunctionComponent<any> = ({
                 paypalcommerceratepay: {
                     container: '#checkout-payment-continue',
                     legalTextContainer: 'legal-text-container',
+                    onPending: (isLoading: boolean) => onPending && onPending(isLoading),
```
serhii-tkachenko, line 85:
```suggestion
                    onPending: (isLoading: boolean) => toggleLoadingOverlay(isLoading),
```

serhii-tkachenko: "Call pending when pending"

It might be great to understand what this method do by reading the function name. From SDK perspective it has an option onPending that will be called (ideally) when the flow will get to this pending state. From UI perspective, you passing the function like: 'You should do that wen the callback will be called'.

So you can say -> 'Show me a loading overlay when the strategy will call onPending callback' -> or 'Hide the loading overlay if pending state finished'.

packages/paypal-commerce-integration/src/PaypalCommerceRatePayPaymentMethod.tsx
```diff
@@ -67,6 +68,8 @@ const PaypalCommerceRatePayPaymentMethod: FunctionComponent<any> = ({
     const fieldsValues = useRef<Partial<RatePayFieldValues>>({});
     const isPaymentDataRequired = checkoutState.data.isPaymentDataRequired();
 
+    const { onPending } = useContext(PaymentContext) as PaymentContextProps;
```
serhii-tkachenko, line 71:
```suggestion
    const { toggleLoadingOverlay } = useContext(PaymentContext) as PaymentContextProps;
```

packages/core/src/scss/components/bigcommerce/loading/_loading.scss
```diff
@@ -64,3 +64,29 @@
         width: 0;
     }
 }
+
+//LOADING SPINNER OVERLAY
+
+#embedded-checkout-loading-spinner-overlay-rotation {
```
serhii-tkachenko, line 70: Why do you use ID for that?

packages/core/src/app/payment/Payment.tsx
```diff
@@ -419,6 +427,11 @@ class Payment extends Component<
         }
     };
 
+    private showLoadingOverlay = (isLoading: boolean): void => {
```
serhii-tkachenko, line 430:
```suggestion
    private toggleLoadingOverlay = (isLoading: boolean): void => {
```

packages/paypal-commerce-integration/src/PaypalCommerceRatePayPaymentMethod.tsx
andriiVitvitskyi1990: Fixed

packages/paypal-commerce-integration/src/PaypalCommerceRatePayPaymentMethod.tsx
andriiVitvitskyi1990: fixed

packages/core/src/scss/components/bigcommerce/loading/_loading.scss
andriiVitvitskyi1990: Change to class

packages/core/src/app/payment/Payment.tsx
andriiVitvitskyi1990: fixed

packages/paypal-commerce-integration/src/PaypalCommerceRatePayPaymentMethod.tsx
```diff
@@ -8,6 +8,7 @@ import {
 import { DynamicFormField, DynamicFormFieldType, FormContext } from '@bigcommerce/checkout/ui';
 import { FormField } from '@bigcommerce/checkout-sdk';
 import getPaypalCommerceRatePayValidationSchema from './validation-schemas/getPaypalCommerceRatePayValidationSchema';
+import { PaymentContext, PaymentContextProps } from '../../core/src/app';
```
serhii-tkachenko, line 11: We should not use any core code in other packages...

packages/core/src/app/ui/loading/LoadingSpinnerOverlay.tsx
serhii-tkachenko, line 1: I think it's better to move this component to UI package

serhii-tkachenko (commented):
I think that loading overlay might be useful for any part of the project... So maybe we should step back from PaymentContext and create another one for the loading overlay?

@bc-peng @animesh1987 do you have some thoughts about that?

packages/core/src/app/ui/loading/LoadingSpinnerOverlay.tsx
```diff
@@ -0,0 +1,46 @@
...
+interface LoadingSpinnerOverlayProps {
+    isLoading: boolean;
+    baseColor?: string;
+    spinnerColor?: string;
+    overlayColor?: string;
+}
+
+const SPINNER_ANIMATION = 'embedded-checkout-loading-spinner-overlay-rotation';
+
+const LoadingSpinnerOverlay: FunctionComponent<LoadingSpinnerOverlayProps> = ({
```
animesh1987, line 12: Is there a reason we are creating a new indicator? Can we not use the existing loading indicator already present in checkout js? We should try and keep one style of loading indicator to keep a uniform design.

packages/paypal-commerce-integration/src/PaypalCommerceRatePayPaymentMethod.tsx
```diff
@@ -67,6 +68,8 @@ const PaypalCommerceRatePayPaymentMethod: FunctionComponent<any> = ({
     const fieldsValues = useRef<Partial<RatePayFieldValues>>({});
     const isPaymentDataRequired = checkoutState.data.isPaymentDataRequired();
 
+    const { toggleLoadingOverlay } = useContext(PaymentContext) as PaymentContextProps;
```
animesh1987, line 71: No `as` please.

packages/core/src/app/payment/PaymentContext.tsx
```diff
@@ -19,6 +19,7 @@ export interface PaymentContextProps {
         schema: ObjectSchema<Partial<PaymentFormValues>> | null,
     ): void;
     hidePaymentSubmitButton(method: PaymentMethod, hidden?: boolean): void;
+    toggleLoadingOverlay?(isLoading: boolean): void;
```
animesh1987, line 22: Does this need to be in the context? Can we implement this only in the component as this is used by only one component at this point?

serhii-tkachenko (approved):

packages/core/src/scss/components/bigcommerce/loading/_loading.scss
```diff
@@ -64,3 +64,13 @@
         width: 0;
     }
 }
+
+.embedded-checkout-loading-spinner-overlay {
+    position: fixed;
+    top: 0;
+    left: 0;
+    background-color: rgba(1,1,1,.3);
```
vitalii-koshovyi, line 72:
```suggestion
    background-color: rgba(1, 1, 1, .3);
```

vitalii-koshovyi (approved):

bc-nick (approved):

packages/core/src/scss/components/bigcommerce/loading/_loading.scss
```diff
@@ -64,3 +64,13 @@
         width: 0;
     }
 }
+
+.embedded-checkout-loading-spinner-overlay {
+    position: fixed;
+    top: 0;
+    left: 0;
+    background-color: rgba(1, 1, 1, .3);
```
bc-peng, line 72: Could you please move the colour code to the SCSS setting?
https://github.com/bigcommerce/checkout-js/blob/master/packages/core/src/scss/settings/global/overlay/_overlay.scss

andriiVitvitskyi1990: moved

animesh1987 (approved):

packages/core/src/scss/components/bigcommerce/loading/_loading.scss
bc-peng: Please just move `rgba(1, 1, 1, .3)`  so that we can reuse this colour as a setting value.

serhii-tkachenko:
Its better to create another branch for prerelease versions

PR #1522, opened 2023-09-26
fix(payment): CHECKOUT-7761 Remove wallet buttons if no payment is required

animesh1987:
## What?
Remove wallet buttons in customer step if payment is not required for an order

## Why?
Presence of wallet buttons in checkout flow if payment is not required for an order causes confusion and causes errors. Hence remove wallet buttons from customer step if payment is not required for an order.

## Testing / Proof
https://github.com/bigcommerce/checkout-js/assets/7134802/6f960cd2-cb0e-42cd-9948-86ae293efb4b

@bigcommerce/team-checkout


serhii-tkachenko (approved):
Nice work! :)

bc-peng (approved):
LGTM.👏👏

PR #1540, opened 2023-10-04
feat(payment): PI-633 add stored cards component to square v2 payment method

bc-wpietrzak:
## What?
Added components for displaying/updating stored credit cards to Square v2 payment form.

## Why?
[PI-633](https://bigcommercecloud.atlassian.net/browse/PI-633)
As a part of stored credit cards implementation for Square v2.

## Testing / Proof
manual + unit testing

Screenshot 2023-10-24 at 10 44 32
Screenshot 2023-10-24 at 12 26 54
Screenshot 2023-10-24 at 12 27 33

@bigcommerce/team-checkout

[PI-633]: https://bigcommercecloud.atlassian.net/browse/PI-633?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/squarev2-integration/src/SquareV2PaymentMethod.tsx
```diff
@@ -178,12 +179,14 @@ const SquareV2PaymentMethod: FunctionComponent<PaymentMethodProps> = ({
...
-                <div id={containerId} style={{ minHeight: '100px' }} />
-            </LoadingOverlay>
-        </div>
+        <SquareV2Form
+            checkoutState={checkoutState}
+            containerId={containerId}
+            deinitializePayment={deinitializePayment}
+            initializePayment={initializePayment}
+            method={method}
+            renderPlaceholderFields={renderPlaceholderFields}
```
serhii-tkachenko, line 188: You can move `renderPlaceholderFields` inside `SquareV2Form`, because it is not related to any props from parent component

packages/squarev2-integration/src/SquareV2Form.tsx
```diff
@@ -0,0 +1,119 @@
...
+                    data-test={containerId}
+                    id={containerId}
+                    style={{
+                        display: !shouldShowCreditCardFieldset ? 'none' : undefined,
+                    }}
+                    tabIndex={-1}
+                >
+                    {renderPlaceholderFields()}
+                    <div id={containerId} />
+                </div>
```
serhii-tkachenko, line 109: Is there any reason why do you use the same id for both divs?

packages/squarev2-integration/src/SquareV2PaymentMethod.tsx
bc-wpietrzak: fixed

serhii-tkachenko (approved):
Nice work!

packages/squarev2-integration/src/SquareV2PaymentMethod.spec.tsx
```diff
@@ -6,14 +6,28 @@ import {
     PaymentInitializeOptions,
     PaymentRequestOptions,
 } from '@bigcommerce/checkout-sdk';
-import { mount } from 'enzyme';
```
animesh1987, line 9: Can we please convert this test to RTL test.

animesh1987 (approved):

packages/squarev2-integration/src/SquareV2Form.tsx
```diff
@@ -0,0 +1,140 @@
...
+        <LoadingOverlay
+            data-test="squarev2_loading_overlay"
+            hideContentWhenLoading
+            isLoading={isLoadingInstruments()}
+        >
+            <div className="paymentMethod--hosted">
+                {shouldShowInstrumentFieldset && (
+                    <CardInstrumentFieldset
+                        // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+                        instruments={instruments as CardInstrument[]}
```
animesh1987, line 113: minor comment, can we avoid using `as`

bc-wpietrzak:
@animesh1987 @serhii-tkachenko since your last review I've added some changes, would love to get some opinions on them as well

packages/squarev2-integration/src/SquareV2Form.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+    useEffect(() => {
+        if (isInstrumentFeatureAvailable) {
+            void checkoutService.loadInstruments();
+        }
+    }, [checkoutService, isInstrumentFeatureAvailable]);
+
+    const instruments = useMemo(() => getInstruments(method) || [], [getInstruments, method]);
+    const shouldShowInstrumentFieldset = isInstrumentFeatureAvailable && instruments.length > 0;
+    const shouldShowCreditCardFieldset = !shouldShowInstrumentFieldset || isAddingNewCard;
```
serhii-tkachenko, line 49: I might recommend to check with someone:
do you need to render credit cards for cases, when the customer has untrusted shipping address for the instruments?

This is a part of extra security to avoid fraud when paying with stored cards

bc-wpietrzak: Checked with the team, it seems our project requirements do not include handling untrusted shipping addresses.

bc-peng (commented):
Could you please update the codeowners so that this package won't require a checkout team's review again?
https://github.com/bigcommerce/checkout-js/blob/master/.github/CODEOWNERS

bc-wpietrzak:
@bc-peng updated codeowners

packages/squarev2-integration/src/SquareV2Form.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+    const getDefaultInstrumentId = useCallback(() => {
```
animesh1987, line 51: Here function is called `getDefaultInstrumentId` but it returns a token for the instrument? Do we need to rename it?

packages/squarev2-integration/src/__snapshots__/SquareV2PaymentMethod.spec.tsx.snap
```diff
@@ -3,41 +3,37 @@
 exports[`SquareV2 payment method should render placeholder form fields 1`] = `
 <div
   data-test="squarev2_placeholder_form"
```
animesh1987, line 5: Do we still need this, since the test has moved to RTL?

bc-wpietrzak: We still use this id in one of the RTL tests

packages/squarev2-integration/src/SquareV2Form.tsx
bc-wpietrzak: While a different name may be a better choice, the name `instrumentId` for a token is used in other components, so I'd leave the name as it is to avoid confusion.

PR #1542, opened 2023-10-05
feat(payment): PI-875 [Stripe] Error while try to save Stripe Link stored card as BC vaulted instrument

bc-sebastianszafraniec:
## What?
Now for StripeLink when customer is logged in BC account, the "save this card for future transaction" is not displayed.

## Why?
Shopper should not have ability to save Stripe Link instruments to BC. Remove checkbox.

## Testing / Proof
Before changes
image

After changes
Zrzut ekranu 2023-10-27 o 08 58 25

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/payment/paymentMethod/HostedWidgetPaymentMethod.tsx
+++ b/packages/core/src/app/payment/paymentMethod/HostedWidgetPaymentMethod.tsx
@@ -2,6 +2,7 @@ import {
     AccountInstrument,
     CardInstrument,
     CheckoutSelectors,
+    Customer,
     CustomerInitializeOptions,
     CustomerRequestOptions,
     Instrument,
@@ -9,6 +10,7 @@ import {
     PaymentInstrument,
     PaymentMethod,
     PaymentRequestOptions,
+    StoreConfig,
 } from '@bigcommerce/checkout-sdk';
 import { memoizeOne } from '@bigcommerce/memoize';
 import classNames from 'classnames';
@@ -39,6 +41,7 @@ import withPayment, { WithPaymentProps } from '../withPayment';
 import SignOutLink from './SignOutLink';
 
 export interface HostedWidgetPaymentMethodProps {
+    shouldSavingCardsBeEnabled?: boolean | undefined;
     additionalContainerClassName?: string;
     buttonId?: string;
     containerId: string;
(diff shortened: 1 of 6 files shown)


packages/core/src/app/payment/storedInstrument/isInstrumentFeatureAvailable.ts
```diff
@@ -5,20 +5,23 @@ export interface IsInstrumentFeatureAvailableState {
...
     paymentMethod: PaymentMethod;
+    shouldSavingCardsBeEnabled?: boolean;
 }
 
 export default function isInstrumentFeatureAvailable({
     config,
     customer,
     isUsingMultiShipping,
     paymentMethod,
+    shouldSavingCardsBeEnabled = true,
```
PavlenkoM, line 16: Maybe better to change this value to `undefined` by default and make it like `shouldSavingCardsBeDisabled`  or `disabledSavingCards` ? what do you think?

PavlenkoM:
Can you please add some tests to cover new logic?

packages/core/src/app/payment/storedInstrument/isInstrumentFeatureAvailable.ts
bc-sebastianszafraniec: As we discussed on private message, I personally like keeping `shouldSavingCardsBeEnabled` set to `true` as default as it indicates that checkbox should be visible by default. But it can be also set as `undefined` if there is such a need

bc-sebastianszafraniec:
> Can you please add some tests to cover new logic?

Sure, I will add tests

vitalii-koshovyi (approved):

PavlenkoM (approved):

packages/core/src/app/payment/paymentMethod/HostedWidgetPaymentMethod.tsx
```diff
@@ -332,9 +338,30 @@ class HostedWidgetPaymentMethod extends Component<
         );
     }
 
+
+    private isInstrumentFeatureAvailable() {
```
animesh1987, line 342: Do we need to create this method here? Seems like we can call the method in `storedInstruments` folder directly?

packages/core/src/app/payment/paymentMethod/StripeUPEPaymentMethod.tsx
```diff
@@ -97,6 +98,14 @@ const StripeUPEPaymentMethod: FunctionComponent<
         );
     };
 
+    const shouldSavingCardsBeEnabled = (): boolean => {
```
animesh1987, line 101: We have all these details in HostedWidgetPaymentMethod component, can we move this logic there?

bc-sebastianszafraniec: The idea was to make it more reusable, and give the possibility to pass shouldSavingCardsBeEnabled prop from payment method (if HostedWidgetPaymentMethod is used). For StripeUPE the conditions are described in shouldSavingCardsBeEnabled method and this way with passing shouldSavingCardsBeEnabled props gives other payment methods possibility to specify their own conditions.

packages/core/src/app/payment/paymentMethod/HostedWidgetPaymentMethod.tsx
bc-sebastianszafraniec: Yes, it can be called directly. I will change it.

animesh1987 (approved):

PR #1544, opened 2023-10-06
feat(checkout): CHECKOUT-7763 Update Sentry config

bc-peng:
## What?
Update the sentry config so that we can turn on 100% Sentry logging for certain brands.

## Why?
To track all issues for certain stores, so we can get better visibility on issues.

## Depends On (already released)
- https://github.com/bigcommerce/bigcommerce/pull/54892

## Testing / Proof
- CI checks.

### DIFF
--- a/packages/core/src/app/checkout/CheckoutApp.tsx
+++ b/packages/core/src/app/checkout/CheckoutApp.tsx
@@ -5,13 +5,11 @@ import ReactModal from 'react-modal';
 
 import { AnalyticsProvider } from '@bigcommerce/checkout/analytics';
 import { ExtensionProvider } from '@bigcommerce/checkout/checkout-extension';
-import { ErrorBoundary, ErrorLogger } from '@bigcommerce/checkout/error-handling-utils';
 import { getLanguageService, LocaleProvider } from '@bigcommerce/checkout/locale';
 import { CheckoutProvider } from '@bigcommerce/checkout/payment-integration-api';
 
 import '../../scss/App.scss';
 
-import { createErrorLogger } from '../common/error';
 import {
     createEmbeddedCheckoutStylesheet,
     createEmbeddedCheckoutSupport,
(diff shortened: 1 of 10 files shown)


packages/core/src/app/checkout/CheckoutApp.tsx
```diff
@@ -55,23 +40,20 @@ export default class CheckoutApp extends Component<CheckoutAppProps> {
 
     render() {
         return (
-            <ErrorBoundary logger={this.errorLogger}>
```
animesh1987, line 58: Howcome we are removing the `ErrorBoundary` here?

animesh1987: Ignore I see it's moved to the child component.

bc-peng: Moved to `checkout.tsx`, that's when we can have a read on the feature flag.
And I don't think we'll compromise on the capture scope, maybe a little.
I was going to comment here, another method is to keep the `ErrorBoundary` here and set it to always 100%. Only use the feature flag to control the error logger in `checkout.tsx`.

packages/core/src/app/order/OrderConfirmation.tsx
```diff
@@ -39,10 +39,11 @@ import { MobileView } from '../ui/responsive';
 import getPaymentInstructions from './getPaymentInstructions';
 import mapToOrderSummarySubtotalsProps from './mapToOrderSummarySubtotalsProps';
 import OrderConfirmationSection from './OrderConfirmationSection';
+import OrderIncompleteHeader from './OrderIncompleteHeader';
 import OrderStatus from './OrderStatus';
 import PrintLink from './PrintLink';
 import ThankYouHeader from './ThankYouHeader';
-import OrderIncompleteHeader from './OrderIncompleteHeader';
+import { BrowserOptions } from '@sentry/browser';
```
animesh1987, line 46: Import order not respected here?

animesh1987 (approved):

PR #1557, opened 2023-10-18
feat(checkout): CHECKOUT-7763 Remove convertcart filter

bc-peng:
## What?
- Remove `convertcart` filter for Sentry logging.
- Revert back to the previous structure of `ErrorBoundary`.
- Read sentry sample rate from BCapp template.

## Why?
We are using [Sentry server side filter ](https://docs.sentry.io/product/accounts/quotas/manage-event-stream-guide/#5-applying-workflows)to discard unwanted issues now.
> If you ["Delete & Discard"](https://docs.sentry.io/product/accounts/quotas/manage-event-stream-guide/#delete--discard) an issue, then future error events with the same fingerprint will not count toward your quota.

## Testing / Proof
- CI checks.

### DIFF
--- a/packages/core/src/app/checkout/Checkout.tsx
+++ b/packages/core/src/app/checkout/Checkout.tsx
@@ -261,24 +261,6 @@ class Checkout extends Component<
 
             window.addEventListener('beforeunload', this.handleBeforeExit);
 
-            const isSentryLoggingAll =
-                data.getConfig()?.checkoutSettings.features[
-                    'CHECKOUT-7764.Increase_sentry_logging_to_100_percent'
-                ] ?? false;
-
-            if (isSentryLoggingAll) {
-                const { sentryConfig, publicPath } = this.props;
-
-                this.errorLogger = createErrorLogger(
-                    { sentry: sentryConfig },
-                    {
-                        errorTypes: ['UnrecoverableError'],
-                        publicPath,
-                        sampleRate: 1,
-                    },
-                );
-            }
-
             if (isExtensionEnabled()) {
                 await extensionService.loadExtensions();
             }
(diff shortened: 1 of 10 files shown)


packages/core/src/app/checkout/Checkout.tsx
```diff
@@ -288,6 +287,24 @@ class Checkout extends Component<
...
 
+            const isSentryLoggingAll =
+                data.getConfig()?.checkoutSettings.features[
+                    'CHECKOUT-7764.Increase_sentry_logging_to_100_percent'
+                ] ?? false;
+
+            if (isSentryLoggingAll) {
+                const { sentryConfig, publicPath } = this.props;
+
+                this.errorLogger = createErrorLogger(
```
bc-peng, line 298: According to the Sentry doc, changing the error sample rate requires re-deployment.
https://docs.sentry.io/platforms/php/guides/laravel/configuration/sampling/#sampling-error-events

packages/core/src/app/common/error/SentryErrorLogger.ts
```diff
@@ -147,16 +146,6 @@ export default class SentryErrorLogger implements ErrorLogger {
     }
 
     private handleBeforeSend: (event: Event, hint?: EventHint) => Event | null = (event, hint) => {
-        const isConvertcart = Boolean(
```
bc-peng, line 150: The issue is discarded on the Sentry server side.

packages/core/src/app/checkout/CheckoutApp.tsx
bc-peng, line 1: All changes here were made by reverting the previous commit, https://github.com/bigcommerce/checkout-js/commit/3a0a4676ef342475b8f5c118098033e1c5e925d9.

animesh1987 (approved):

PR #1559, opened 2023-10-18
fix(checkout): CHECKOUT-5833 fix date issue in different timezones

richapatel1510:
## What?
Changed the format of date provided to `new Date(dateString)` function to be separated by year, month and day, i.e. `new Date(year, monthIndex, day)`

## Why?
As the `new Date(dateString)` gives the time in UTC timezone it displays incorrectly (1 day back) in some timezones, so the change is to display correct date in those timezones.

## Testing / Proof
**Manual test**

**BEFORE**

https://github.com/bigcommerce/checkout-js/assets/141614330/17d569d1-c561-44bd-ba3b-0ecb6fff308d

**AFTER**

https://github.com/bigcommerce/checkout-js/assets/141614330/268a1774-a311-4d78-a78b-ec55fec986b3

### DIFF
--- a/packages/core/src/app/address/mapAddressToFormValues.ts
+++ b/packages/core/src/app/address/mapAddressToFormValues.ts
@@ -69,7 +69,7 @@ function getValue(
     }
 
     if (fieldType === DynamicFormFieldType.date && typeof fieldValue === 'string') {
-        if(fieldValue){
+        if (fieldValue) {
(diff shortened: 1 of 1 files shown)


snaderiBC (approved):
👏

bc-shawnwang (approved):
`new Date(year, monthIndex, day)`

only month is as index. 😱

animesh1987 (approved):

PR #1563, opened 2023-10-18
feat(payment): PI-921 Amazon Pay Button needs to be changed to one with WITHOUT microtext

bc-sebastianszafraniec:
## What?
AmazonPay button width is adjusted properly.

## Why?
AmazonPay button width in Customer checkout step was not adjusted properly.

## Testing / Proof
1.Before changes 1

Before changes

2. After changes 1

After changes

And 

3. Before changes 2
Before changes 2

4. After changes  2

After changes 2

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/customer/customWalletButton/AmazonPayV2Button.tsx
+++ b/packages/core/src/app/customer/customWalletButton/AmazonPayV2Button.tsx
@@ -5,7 +5,7 @@ import CheckoutButton, { CheckoutButtonProps } from '../CheckoutButton';
 const AmazonPayV2Button: FunctionComponent<CheckoutButtonProps> = (props) => {
     useEffect(() => {
         beautifyAmazonButton();
-    }, [props]);
+    }, []);
 
     return (
         <div className="AmazonPayContainer">
@@ -26,10 +26,8 @@ const beautifyAmazonButton = (): void => {
 
     if (container && amazonButton) {
         amazonButton.style.height = '36px';
-        container.style.width = '100%';
-        
         return;
     }
 
-    setTimeout(beautifyAmazonButton, 200);
+    setTimeout(beautifyAmazonButton, 10);
 }
--- a/packages/core/src/app/customer/customWalletButton/AmazonPayV2Button.tsx
+++ b/packages/core/src/app/customer/customWalletButton/AmazonPayV2Button.tsx
@@ -5,7 +5,7 @@ import CheckoutButton, { CheckoutButtonProps } from '../CheckoutButton';
 const AmazonPayV2Button: FunctionComponent<CheckoutButtonProps> = (props) => {
     useEffect(() => {
         beautifyAmazonButton();
-    }, []);
+    }, [props]);
 
     return (
         <div className="AmazonPayContainer">
@@ -26,8 +26,10 @@ const beautifyAmazonButton = (): void => {
 
     if (container && amazonButton) {
         amazonButton.style.height = '36px';
+        container.style.width = '100%';
+        
         return;
     }
 
-    setTimeout(beautifyAmazonButton, 10);
+    setTimeout(beautifyAmazonButton, 200);
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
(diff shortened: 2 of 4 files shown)


packages/core/src/app/customer/customWalletButton/AmazonPayV2Button.tsx
```diff
@@ -5,7 +5,7 @@ import CheckoutButton, { CheckoutButtonProps } from '../CheckoutButton';
 const AmazonPayV2Button: FunctionComponent<CheckoutButtonProps> = (props) => {
     useEffect(() => {
         beautifyAmazonButton();
-    }, []);
+    }, [props]);
```
animesh1987, line 8: Do we need to pass `props` here? Don't think it's used in the effect?

serhii-tkachenko (approved):
LGTM!

packages/core/src/app/customer/customWalletButton/AmazonPayV2Button.tsx
```diff
@@ -26,8 +26,10 @@ const beautifyAmazonButton = (): void => {
 
     if (container && amazonButton) {
         amazonButton.style.height = '36px';
+        container.style.width = '100%';
+        
         return;
     }
 
-    setTimeout(beautifyAmazonButton, 10);
+    setTimeout(beautifyAmazonButton, 200);
```
animesh1987, line 34: Why are we changing the timeout?

packages/core/src/app/customer/customWalletButton/AmazonPayV2Button.tsx
bc-sebastianszafraniec: I think yes because container width which we depend on here is set later and wen trying to use 
```
useEffect(() => {
        beautifyAmazonButton();
    }, []);
```
the propper width is not set yet.

packages/core/src/app/customer/customWalletButton/AmazonPayV2Button.tsx
bc-sebastianszafraniec: I changed timeout to reduce beautifyAmazonButton function calls. I set it to 200 as it still looks ok then.

packages/core/src/app/customer/customWalletButton/AmazonPayV2Button.tsx
bc-peng: The interface of props does not seem to contain the width.
```
export interface CheckoutButtonProps {
    containerId: string;
    methodId: string;
    isShowingWalletButtonsOnTop?: boolean;
    deinitialize(options: CustomerRequestOptions): void;
    initialize(options: CustomerInitializeOptions): void;
    onError?(error: Error): void;
}
```
Could you please specify which prop it depends on?

bc-sebastianszafraniec: Actually It doesn’t depend on any specific information in props. It rather depends on that AmazonPayV2Button component is receiving new props object, which cause the `beautifyAmazonButton()` function call.
Let me explain:
Previously `beautifyAmazonButton()` was called after first render:

```
useEffect(() => {
        beautifyAmazonButton();
    }, []);
```

But it was to early to set container width to 100% as in this moment its parent div has not yet set proper width. In next steps AmazonPayV2Button component is receiving new props object (in mean time parent div is getting proper width) and then when `beautifyAmazonButton()` function is called and the container is getting proper 100% width.

bc-peng: Can I know how many times will this `useEffect` be called in a normal checkout page load after passing `prop` ?

bc-sebastianszafraniec: I checked and as I see this useEffect is called 13 times

bc-peng: I think we would all agree 13 is a big number.
It's not a blocker, but could you please try use just few props to reduce the rendering times?

animesh1987: Is this 13, with adding the props object? If yes then I align with Peng, maybe we need to rethink how the width is set as this feels like a hack and can cause some side effects.

bc-peng (approved):

PR #1565, opened 2023-10-18
chore(payment): PAYPAL-00 fix of e2e test

bc-nick:
## What?

Fix of e2e test

## Why?

Because of https://app.circleci.com/pipelines/github/bigcommerce/checkout-js/8215/workflows/8a2dda50-012b-4191-b698-e9538034d7c7/jobs/28189

## Testing / Proof

All tests pased

@bigcommerce/team-checkout


animesh1987 (approved):

packages/google-pay-integration/e2e/GooglePayBraintreeInPaymentStep.spec.ts
```diff
@@ -51,6 +53,20 @@ test.describe('Google Pay Braintree', () => {
...
+                if (request.url().includes(`checkoutId=${mockCartId}`)) {
+                    return route.fulfill({
+                        ...responseProps,
+                        body: JSON.stringify(checkoutSettings),
+                    });
+                }
+
+                await route.fallback();
+            },
+        );
```
bc-peng, line 68: We should avoid mocking `checkout-settings` because if the structure of `checkout-settings` changes, we can convert all old recordings into new format here.
https://github.com/bigcommerce/checkout-js/blob/25633d5cc3cae0f800cbfdd101fca9487c70e235/packages/test-framework/src/fixture/pageObject/playwright/PollyObject.ts#L121
A mocking response like this will just break the test in the scenario above.

bc-nick: Thanks! I got it. I will think about the different way to solve it

PR #1568, opened 2023-10-24
fix(checkout): CHECKOUT-7422 fix address localization on Checkout

richapatel1510:
## What?

1. Refetch country names while localizing if they are not available.
2. Load billing address countries as part of loading shipping form.

## Why?

1. Country names should appear in local language set for storefront.
3. We have a bug at this moment due to which billing form is displayed even if the setting My billing address is the same as my shipping address. is checked true, but postal code is not required for a country.

This happens due to https://github.com/bigcommerce/checkout-sdk-js/blob/master/packages/core/src/form/form-selector.ts#L130 line resolving to and [] since we have not loaded billing countries yet. Which results in postal code being marked as required on the form and hence the form being incomplete which results in user loading the form.

Hence if we fetch billing countries earlier in the process and update the state we can get accurate form field validations as per the countries data.

## Testing / Proof

- CI checks
- Manual testing

**BEFORE:**

https://github.com/bigcommerce/checkout-js/assets/141614330/c3174b17-560b-417a-8a2a-8e46e16255d2

**AFTER:**

https://github.com/bigcommerce/checkout-js/assets/141614330/7cac71b8-de90-445c-953c-0aa9fb31d9ba

### DIFF
--- a/packages/core/src/app/address/StaticAddress.tsx
+++ b/packages/core/src/app/address/StaticAddress.tsx
@@ -45,7 +45,9 @@ const StaticAddress: FunctionComponent<
         const fetchData = async () => {
             await loadAddressFields();
         }
-        if (!countries) fetchData();
+        if (!countries) {
+            fetchData();
+        }
     }, []);
 
     const address = localizeAddress(addressWithoutLocalization, countries);
--- a/packages/core/src/app/address/StaticAddress.tsx
+++ b/packages/core/src/app/address/StaticAddress.tsx
@@ -7,7 +7,7 @@ import {
 } from '@bigcommerce/checkout-sdk';
 import classNames from 'classnames';
 import { isEmpty } from 'lodash';
-import React, { FunctionComponent, memo, useEffect } from 'react';
+import React, { FunctionComponent, memo } from 'react';
 
 import { CheckoutContextProps } from '@bigcommerce/checkout/payment-integration-api';
 import { isPayPalConnectAddress, usePayPalConnectAddress } from '@bigcommerce/checkout/paypal-connect-integration';
@@ -33,23 +33,12 @@ export interface StaticAddressEditableProps extends StaticAddressProps {
 interface WithCheckoutStaticAddressProps {
     countries?: Country[];
     fields?: FormField[];
-    loadAddressFields: () => Promise<CheckoutSelectors>
 }
 
 const StaticAddress: FunctionComponent<
     StaticAddressEditableProps & WithCheckoutStaticAddressProps
-> = ({ countries, fields, loadAddressFields, address: addressWithoutLocalization }) => {
+> = ({ countries, fields, address: addressWithoutLocalization }) => {
     const { isPayPalAxoEnabled, paypalConnectAddresses } = usePayPalConnectAddress();
-
-    useEffect(() => {
-        const fetchData = async () => {
-            await loadAddressFields();
-        }
-        if (!countries) {
-            fetchData();
-        }
-    }, []);
-
     const address = localizeAddress(addressWithoutLocalization, countries);
     const isValid = !fields
         ? !isEmpty(address)
@@ -117,9 +106,6 @@ export function mapToStaticAddressProps(
         checkoutState: {
             data: { getBillingCountries, getShippingCountries, getBillingAddressFields, getShippingAddressFields },
         },
-        checkoutService: {
-            loadBillingAddressFields, loadShippingAddressFields
-        }
     } = context;
 
     return {
@@ -132,9 +118,6 @@ export function mapToStaticAddressProps(
                 : type === AddressType.Shipping
                 ? getShippingAddressFields(address.countryCode)
                 : undefined,
-        loadAddressFields: type === AddressType.Billing
-            ? loadBillingAddressFields
-            : loadShippingAddressFields
     };
 }
 
--- a/packages/core/src/app/address/StaticAddress.tsx
+++ b/packages/core/src/app/address/StaticAddress.tsx
@@ -7,7 +7,7 @@ import {
 } from '@bigcommerce/checkout-sdk';
 import classNames from 'classnames';
 import { isEmpty } from 'lodash';
-import React, { FunctionComponent, memo } from 'react';
+import React, { FunctionComponent, memo, useEffect } from 'react';
 
 import { CheckoutContextProps } from '@bigcommerce/checkout/payment-integration-api';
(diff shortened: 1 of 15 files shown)


packages/core/src/app/address/StaticAddress.tsx
```diff
@@ -104,18 +113,26 @@ export function mapToStaticAddressProps(
...
+        checkoutService: {
+            loadBillingAddressFields, loadShippingAddressFields
+        }
     } = context;
 
     return {
-        countries: getBillingCountries(),
+        countries: type === AddressType.Billing
+            ? getBillingCountries()
+            : getShippingCountries(),
```
snaderiBC, line 126: Is this fixing another issue?
Looks like before this change we ignored the type and always returned billing countries. But billing countries are not necessarily the same as shipping countries. Is my understanding correct?

richapatel1510: Not sure if it is solving something else as well here. But the issue here was that for country name translations we need countries data from API and since Billing form was not yet rendered (in case where billing is same as shipping or refreshing checkout page after filling both forms) the API to get billing countries wasn't triggered.  

Yes, you are right, billing countries are not necessarily the same as shipping countries because we have different API's to fetch them.

packages/core/src/app/address/StaticAddress.tsx
```diff
@@ -33,12 +33,21 @@ export interface StaticAddressEditableProps extends StaticAddressProps {
...
     StaticAddressEditableProps & WithCheckoutStaticAddressProps
-> = ({ countries, fields, address: addressWithoutLocalization }) => {
+> = ({ countries, fields, loadAddressFields, address: addressWithoutLocalization }) => {
     const { isPayPalAxoEnabled, paypalConnectAddresses } = usePayPalConnectAddress();
+
+    useEffect(() => {
+        const fetchData = async () => {
+            await loadAddressFields();
+        }
+        if (!countries) fetchData();
```
animesh1987, line 48: Can you please write this as 
```suggestion
        if (!countries) {
            fetchData()
        };
```

packages/core/src/app/address/StaticAddress.tsx
snaderiBC: I guess I was thinking about this: https://github.com/bigcommerce/checkout-js/pull/1571
These two PRs are solving very similar problems 🤔

richapatel1510: Yes, apparently #1571 would solve this translation problem partially.

snaderiBC: @animesh1987 I don't know if we need two PRs here?
Maybe changes you're adding in your PR can be included in this PR instead. What do you think?

animesh1987: Not entirely sure, as @richapatel1510 is adding support to `getShippingCountries` to StaticAddress. Where as in my PR I am preloading billing countries in shipping step.
So they might look similar but are different solutions. We can remove the call to loadCountries here but will need the change https://github.com/bigcommerce/checkout-js/pull/1568/files#diff-ded4d841a24b30f3c3ba48ec964a53a61045c8156846829a1aab4bb7ee986ed9R126 to get correct list of countries.

animesh1987: But @snaderiBC to your point both bugs are result of the same symptom that we don't have the correct data loaded. I have added a detailed description in my PR, see if that makes sense.

packages/core/src/app/shipping/Shipping.tsx
```diff
@@ -241,6 +243,7 @@ class Shipping extends Component<ShippingProps & WithCheckoutShippingProps, Ship
             !isEqualAddress(updatedShippingAddress, billingAddress) &&
             !hasRemoteBilling
         ) {
+            promises.push(loadBillingAddressFields())
```
richapatel1510, line 246: We only need to make this API call in shipping step when billing address is same as shipping address and we need to fetch the countries for billing address so that `optional postal code` for some countries is considered and validation is not triggered.

packages/core/src/app/address/StaticAddress.tsx
```diff
@@ -33,12 +33,23 @@ export interface StaticAddressEditableProps extends StaticAddressProps {
...
-> = ({ countries, fields, address: addressWithoutLocalization }) => {
+> = ({ countries, fields, loadAddressFields, address: addressWithoutLocalization }) => {
     const { isPayPalAxoEnabled, paypalConnectAddresses } = usePayPalConnectAddress();
+
+    useEffect(() => {
+        const fetchData = async () => {
+            await loadAddressFields();
+        }
+        if (!countries) {
+            fetchData();
```
richapatel1510, line 49: Need to fetch data here in case where countries are not fetched, i.e. on refreshing the checkout page when shipping and billing forms are filled already.

animesh1987: Ahh I see, maybe then we need to move the call back to shipping step mount. As we should try and avoid making API calls from low level components as it can lead to side effects and make debugging harder.

richapatel1510: Moving back to shipping step mount will result in all e2e tests failing. And we really don't need that API call at Shipping step mount. We only need it there in case where billing is same as shipping and billing step is not complete, just to avoid postal code validation. 
Moreover this will be called only in case someone refreshes checkout after reaching payment step. So we need to decide whether we want to update all the recording files for e2e tests and move this to mount or go with the current solution. 
CC: @bc-peng

bc-peng: We noticed this extra call while investigating why the majority of e2e tests failed in previous solutions.
The extra call is avoidable if we apply a stricter rule of making the request.
I think we should be able to run `loadAddressFields();` at any point within the shipping step, just need to figure out the conditions.

packages/test-framework/src/fixture/pageObject/playwright/PlaywrightHelper.ts
```diff
@@ -54,6 +89,8 @@ export class PlaywrightHelper {
             void route.fulfill({ status: 200, path: `${this.srcPath}/support/product.png` });
         });
 
+        // TODO: Remove these mocks when updating HAR files
+        await this.mockRoutes();
```
bc-peng, line 93:
```suggestion
        await this.mockRoutes();
        
```

bc-peng (approved):
LGTM👏👏 Nice fix.

animesh1987 (approved):

PR #1570, opened 2023-10-24
fix(payment): PAYPAL-3144 fixed the issue with PayPal Connect logo style glitch

serhii-tkachenko:
## What?
There was an issue with custom theme that affects PayPal Connect logo styling with display flex

## Why?
PayPal Connect brainding elements looks wrong on merchants checkout page

## Testing / Proof
Unit tests
Manual tests

Before:
Screenshot 2023-10-24 at 16 54 47

After:
Screenshot 2023-10-24 at 16 52 59

Nothing changed for default Cornerstone Checkout page:
Screenshot 2023-10-24 at 16 52 49

### DIFF
--- a/packages/core/src/app/shipping/StaticConsignment.scss
+++ b/packages/core/src/app/shipping/StaticConsignment.scss
@@ -1,5 +1,9 @@
 @import '../ui/Base';
 
+.staticConsignment {
+    width: 100%;
+}
+
 .staticConsignmentContainer {
     ul {
         list-style: none;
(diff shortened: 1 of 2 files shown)


andriiVitvitskyi1990 (approved):

bc-nick (approved):

bc-dronov (approved):

packages/core/src/app/shipping/StaticConsignment.scss
```diff
@@ -1,5 +1,9 @@
 @import '../ui/Base';
 
+.staticConsignment {
```
animesh1987, line 3: Just want to check if this causes issues with existing components? I don't think so but just want to check.

serhii-tkachenko: This element is display: block by default so it fills up 100% of parent block width. However, for custom themes some common css classes can have display flex on child blocks, so the staticConsignment changes its width to a size, that fits the content.

animesh1987 (approved):

PR #1571, opened 2023-10-25
fix(billing): CHECKOUT-7803 Load billing address countries when shipping form loads

animesh1987:
## What?
Load billing address countries as part of loading shipping form.

## Why?
We have a bug at this moment due to which billing form is displayed even if the setting `My billing address is the same as my shipping address.` is checked `true`, but postal code is not required for a country. 

This happens due to https://github.com/bigcommerce/checkout-sdk-js/blob/master/packages/core/src/form/form-selector.ts#L130 line resolving to and `[]` since we have not loaded billing countries yet. Which results in postal code being marked as required on the form and hence the form being incomplete which results in user loading the form.

Hence if we fetch billing countries earlier in the process and update the state we can get accurate form field validations as per the countries data.

## Testing / Proof
- CI

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/shipping/Shipping.tsx
+++ b/packages/core/src/app/shipping/Shipping.tsx
@@ -73,6 +73,7 @@ export interface WithCheckoutShippingProps {
     deleteConsignments(): Promise<Address | undefined>;
     getFields(countryCode?: string): FormField[];
     initializeShippingMethod(options: ShippingInitializeOptions): Promise<CheckoutSelectors>;
+    loadBillingAddressFields(): Promise<CheckoutSelectors>;
     loadShippingAddressFields(): Promise<CheckoutSelectors>;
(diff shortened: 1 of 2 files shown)


packages/core/src/app/shipping/Shipping.tsx
```diff
@@ -386,6 +388,9 @@ export function mapToShippingProps({
         hasMultiShippingEnabled && !methodId && shippableItemsCount > 1;
     const countriesWithAutocomplete = ['US', 'CA', 'AU', 'NZ'];
 
+    console.log('get billing address');
+    // checkoutService.loadBillingAddressFields();
+
```
richapatel1510, line 393: logs

animesh1987:
Closing this PR as it will be addressed in https://github.com/bigcommerce/checkout-js/pull/1568

PR #1577, opened 2023-10-30
feat(payment): PAYPAL-2887 removed unnecessary sort order

bc-nick:
## What?

Removed unnecessary sort order

## Why?

The order must match the order in the `remoteCheckoutProviders` array

@bigcommerce/team-checkout


packages/core/src/app/customer/__snapshots__/CheckoutButtonContainer.spec.tsx.snap
```diff
@@ -65,6 +59,12 @@ exports[`CheckoutButtonContainer displays wallet buttons for guest checkout 1`]
         <div
           id="googlepayauthorizenetCheckoutButton"
         />
+        <div
```
serhii-tkachenko, line 62: It might be better to update CheckoutButtonContainer.spec file and mock availableMethodIds variable in a right order. This snapshot may confuse other developers and it shows wrong information, because paypal related providers returned from BE will be first in the list

animesh1987: Good point we should not be editing snapshot files manually.

serhii-tkachenko (approved):

animesh1987 (approved):

PR #1584, opened 2023-11-03
feat(checkout): CHECKOUT-7838 preload extension with link tag

richapatel1510:
## What?
Optimize Extension Loading with Preloading via `<link>` tag

## Why?
In the extension project context, we can add all extension URLs to `<head>` within `<link>` tag as soon as we finish loading extensions so that we can shorten the time it takes to display an extension.

## Testing / Proof

- Manual testing

https://github.com/bigcommerce/checkout-js/assets/141614330/16f7df69-112d-43d9-8510-305bb74ba492

### DIFF
--- a/packages/checkout-extension/src/ExtensionService.ts
+++ b/packages/checkout-extension/src/ExtensionService.ts
@@ -22,6 +22,35 @@ export class ExtensionService {
         await this.checkoutService.loadExtensions();
     }
 
+    preloadExtensions(): void {
+        const state = this.checkoutService.getState();
+        const extensions = state.data.getExtensions();
+        const cartId = state.data.getCart()?.id;
+        const parentOrigin = state.data.getConfig()?.links.siteLink;
+
+        if (!cartId || !parentOrigin) return;
+        extensions?.forEach((extension) => {
+            const url = new URL(extension.url);
+
+            url.searchParams.set('extensionId', extension.id);
+            url.searchParams.set('cartId', cartId);
+            url.searchParams.set('parentOrigin', parentOrigin);
+
+            const link = document.createElement('link');
+
+            // Set the attributes for the link element
+            link.rel = 'preload';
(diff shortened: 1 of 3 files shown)


packages/checkout-extension/src/ExtensionService.ts
bc-peng, line 1: Can you please remove the comments from the file and add a unit test for the new method?

packages/checkout-extension/src/ExtensionService.ts
```diff
@@ -22,6 +22,35 @@ export class ExtensionService {
         await this.checkoutService.loadExtensions();
     }
 
+    preloadExtensions(): void {
+        const state = this.checkoutService.getState();
+        const extensions = state.data.getExtensions();
+        const cartId = state.data.getCart()?.id;
+        const parentOrigin = state.data.getConfig()?.links.siteLink;
+
+        if (!cartId || !parentOrigin) return;
```
animesh1987, line 31:
```suggestion
        if (!cartId || !parentOrigin) {
             return;
        }
```

packages/checkout-extension/src/ExtensionService.ts
```diff
@@ -22,6 +22,35 @@ export class ExtensionService {
         await this.checkoutService.loadExtensions();
     }
 
+    preloadExtensions(): void {
+        const state = this.checkoutService.getState();
+        const extensions = state.data.getExtensions();
+        const cartId = state.data.getCart()?.id;
```
animesh1987, line 28: Can we use `getCartOrThrow` here?

richapatel1510: I don't think that method is available in state data `CheckoutStoreSelector`.

animesh1987 (approved):

bc-peng (approved):

PR #1589, opened 2023-11-07
feat(checkout): CHECKOUT-7837 Optimize `loadCheckout` and `loadExtensions` for Parallel Execution

richapatel1510:
## What?
Remove `isExtensionEnabled` flag and load extensions in parallel to load checkout.

## Why?
Since the extension is rolled out to 100% now and extensions can be now loaded concurrently to checkout settings.

## Testing / Proof

- CI checks
- Manual testing


snaderiBC:
Just curious, is the performance improvement by concurrent request measurable? 🤔

animesh1987 (approved):

animesh1987:
> Just curious, is the performance improvement by parallel loading measurable? 🤔

There will be some benefit in checkout load, since as of now it's a separate blocking call. So moving this with checkout load call will give some benefit straight away. Somewhere between 500ms-1s.

PR #1591, opened 2023-11-08
feat(payment): PAYPAL-2953 bump checkout-sdk version

bc-nick:
## What?
...

## Why?
...

## Testing / Proof
...

@bigcommerce/team-checkout


animesh1987:
Looks ok, but just wanted to check how come there are other packages updated with this as well?

bc-nick:
> Looks ok, but just wanted to check how come there are other packages updated with this as well?

All dependencies updated after ran `npm install`. It is the only command that i used

animesh1987:
> > Looks ok, but just wanted to check how come there are other packages updated with this as well?
> 
> All dependencies updated after ran `npm install`. It is the only command that i used

Can you please run npm install@bigcommerce/checkout-sdk again

bc-nick:
> > > Looks ok, but just wanted to check how come there are other packages updated with this as well?
> > 
> > 
> > All dependencies updated after ran `npm install`. It is the only command that i used
> 
> Can you please run npm install@bigcommerce/checkout-sdk again

The same. Because those dependencies came from `checkout-sdk`

Screenshot 2023-11-08 at 11 55 06

animesh1987 (approved):

PR #1600, opened 2023-11-24
feat(payment): PI-1036 removed Stripe link inititializations for not supported cart amounts

vitalii-koshovyi:
## What?
Removed Stripe link inititializations for not supported cart amounts

## Why?
Because with Stripe Link will crash checkout when amount of the card is smaller then Stripe minimum supported value

## Testing / Proof
Before:
image
After:

https://github.com/bigcommerce/checkout-js/assets/79574476/a9f75d5e-7398-4aa8-b60f-5520061c7428

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/checkout/getCheckoutStepStatuses.ts
+++ b/packages/core/src/app/checkout/getCheckoutStepStatuses.ts
@@ -2,6 +2,8 @@ import { CheckoutSelectors } from '@bigcommerce/checkout-sdk';
 import { compact } from 'lodash';
 import { createSelector } from 'reselect';
 
+import { shouldUseStripeLinkByMinimumAmount } from '@bigcommerce/checkout/instrument-utils';
+
 import { isValidAddress } from '../address';
 import { EMPTY_ARRAY } from '../common/utility';
 import { SUPPORTED_METHODS } from '../customer';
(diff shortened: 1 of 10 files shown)


packages/instrument-utils/src/guards/shouldUseStripeLinkByMinimumAmount.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+        cartAmount: 0.4,
+    };
+
+    it('returns true if cart fullfill Stripe requirements', () => {
+        expect(shouldUseStripeLinkByMinimumAmount(cart)).toBe(true);
+    });
+
+    it('returns false if cart fullfill Stripe requirements', () => {
+        expect(shouldUseStripeLinkByMinimumAmount(notSupportedCart)).toBe(false);
+    });
```
bc-peng, line 18: Could we make the condition more specific?
Now it appears that the same condition can have two outcomes.

packages/instrument-utils/src/guards/shouldUseStripeLinkByMinimumAmount.ts
```diff
@@ -0,0 +1,47 @@
+import { Cart } from '@bigcommerce/checkout-sdk';
+
+enum stripeLinkMinAmount {
```
bc-peng, line 3: Could you provide the basis for setting the value？

vitalii-koshovyi: @bc-peng it was taken from the Stripe docs. I'll add a link for a Stripe documentation

PavlenkoM (approved):

bc-peng (approved):

packages/core/src/app/checkout/getCheckoutStepStatuses.ts
```diff
@@ -39,7 +42,7 @@ const getCustomerStepStatus = createSelector(
         // when reloading the checkout page it's necessary to refill the stripe components with the information saved.
         // In this step, we require that the customer strategy be reloaded the first time.
         const isUsingStripeLinkAndCheckoutPageIsReloaded = !isUsingWallet &&
-            config?.checkoutSettings.providerWithCustomCheckout === PaymentMethodId.StripeUPE && hasEmail && isGuest;
+            config?.checkoutSettings.providerWithCustomCheckout === PaymentMethodId.StripeUPE && hasEmail && isGuest && cart && shouldUseStripeLinkByMinimumAmount(cart);
```
animesh1987, line 45: Can we please move this check in a separate method?

serhii-tkachenko (approved):

animesh1987 (approved):

PR #1602, opened 2023-11-28
feat(payment): CHECKOUT-7901 Fetch button config when checkout loads

animesh1987:
## What?
Fetch button config when checkout loads

## Why?
In order to load button payment configs in parallel, move logic to fetch button configs to checkout component.

## Testing / Proof
- CI
- Manual

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/checkout/Checkout.tsx
+++ b/packages/core/src/app/checkout/Checkout.tsx
@@ -8,9 +8,9 @@ import {
     EmbeddedCheckoutMessenger,
     EmbeddedCheckoutMessengerOptions,
     FlashMessage,
+    PaymentMethod,
     Promotion,
-    RequestOptions,
-} from '@bigcommerce/checkout-sdk';
+ RequestOptions } from '@bigcommerce/checkout-sdk';
 import classNames from 'classnames';
 import { find, findIndex } from 'lodash';
 import React, { Component, lazy, ReactNode } from 'react';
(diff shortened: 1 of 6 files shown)


animesh1987:
There are two approaches that we are trying in this PR for fetching button configs:

- Load button configs in checkout component
Pros: Button configs are fetched for both settings for wallet buttons i.e. in customer component and wallet buttons on top.
Cons: The drawback here is that we need to add logic so button list component doesn't start fetching button configs, leading to extra API calls which can be slow things down.

- Load button config for the buttons on top of page component
Pros: Logic is contained in ButtonContainer component and not distributed across the application.
Cons: Optimisation is only present for buttons on top setting.

animesh1987:
Build is failing due to SDK not merged.

bc-peng (approved):
LGTM, nice work👏👏

packages/core/src/app/checkout/Checkout.tsx
```diff
@@ -8,9 +8,9 @@ import {
     EmbeddedCheckoutMessenger,
     EmbeddedCheckoutMessengerOptions,
     FlashMessage,
+    PaymentMethod,
     Promotion,
-    RequestOptions,
-} from '@bigcommerce/checkout-sdk';
+ RequestOptions } from '@bigcommerce/checkout-sdk';
```
serhii-tkachenko, line 13: Here might be a eslint warning:

```suggestion
    RequestOptions,
 } from '@bigcommerce/checkout-sdk';
```

serhii-tkachenko (approved):
Nice work!

PR #1605, opened 2023-11-30
feat(payment): PAYPAL-2620 paypal-express-integration package

bc-nick:
## What?

Created `paypal-express-integration` package

## Why?

To separate and move related code

@bigcommerce/team-checkout


serhii-tkachenko (approved):

packages/paypal-express-integration/src/index.ts
bc-peng, line 1: Can you please explain why are you making this skeleton package here?
Would it be possible to create a PR with actual content in it?

packages/paypal-express-integration/project.json
```diff
@@ -0,0 +1,25 @@
+{
+  "root": "packages/paypal-express-integration",
+  "sourceRoot": "packages/paypal-express-integration/src",
+  "projectType": "library",
+  "targets": {
+    "lint": {
+      "executor": "@nrwl/linter:eslint",
+      "outputs": ["{options.outputFile}"],
+      "options": {
+        "lintFilePatterns": ["packages/paypal-express-integration/**/*.ts"]
```
bc-peng, line 10:
```suggestion
        "lintFilePatterns": ["packages/paypal-express-integration/**/*.{ts,tsx}"]
```

bc-peng:
Can we add an e2e test to the package?

packages/paypal-express-integration/project.json
bc-nick: done

bc-nick:
> Can we add an e2e test to the package?

@bc-peng would be better to create another ticket for writing e2e test and do it separately. Because i already have combined a two tickets to one PR.
I will create a ticket for this and take it to work in one of the next sprints

.github/CODEOWNERS
```diff
@@ -56,6 +56,7 @@
 /packages/braintree-integration @bigcommerce/team-paypal
 /packages/paypal-commerce-integration @bigcommerce/team-paypal
 /packages/paypal-connect-integration @bigcommerce/team-paypal
+/packages/paypal-express-integration @bigcommerce/team-integrations
```
serhii-tkachenko, line 59:
```suggestion
/packages/paypal-express-integration @bigcommerce/team-paypal 
```

serhii-tkachenko (approved):

bc-peng (approved):
LGTM. 👏👏

vitalii-koshovyi (changes requested):
We need to make this package generic. This functionality should be used by other providers. For example https://github.com/bigcommerce/checkout-js/blob/master/packages/bluesnap-direct-integration/src/BlueSnapDirectAlternativePaymentMethod.tsx
 potentially it could be used by a lot of providers from the `core`

I believe we need to add e2e tests, to be sure that refactoring doesn't break anything.

My proposition is:

**1) Make this package generic for other providers**
You need to create module resolver here https://github.com/bigcommerce/checkout-js/tree/master/packages/hosted-payment-integration/src instead of creation separate module for pp-express
**We just need to figure out what to do with provider specific initialization props** 
I mean this one 
```
paypalexpress: {
     useRedirectFlow: true,
},
```
**2) Create separate package paypal-express-integration** that will contain e2e tests

**3) Remove Bluesnap** [file](https://github.com/bigcommerce/checkout-js/blob/master/packages/bluesnap-direct-integration/src/BlueSnapDirectAlternativePaymentMethod.tsx) and add bluesnap resolver [{ gateway: 'bluesnapdirect' }] to the hosted-payment-integration toResolvableComponent function call.

cc @serhii-tkachenko @PavlenkoM

vitalii-koshovyi:
@bc-peng @animesh1987 @serhii-tkachenko @bc-nick @PavlenkoM  do you agree with my idea that https://github.com/bigcommerce/checkout-js/tree/master/packages/hosted-payment-integration/src should have its own resolver? I think it should be generic for other providers as well. To avoid code duplication.
But we should have a provider specific package that will have e2e tests.
You can check my comment above

serhii-tkachenko:
> @bc-peng @animesh1987 @serhii-tkachenko @bc-nick @PavlenkoM do you agree with my idea that https://github.com/bigcommerce/checkout-js/tree/master/packages/hosted-payment-integration/src should have its own resolver? I think it should be generic for other providers as well. To avoid code duplication. But we should have a provider specific package that will have e2e tests. You can check my comment above

Yes, I agree with the idea, that we probably should reduce amount of code (packages). We can make the same as we have in CheckoutButton - generic component for buttons on checkout page.

vitalii-koshovyi (approved):
After discussion with @serhii-tkachenko and @bc-nick we decided to leave it as it is.
Because providers have unique initializationData and we didn't find a way to make it generic for all providers. 
Current approach is a simplest from other variants that we discussed. 

I would recommend to add e2e test for this paypal-express package.

Well done!

PR #1609, opened 2023-12-01
feat(payment): PI-1148 Hide card holder name from hosted fields form

PavlenkoM:
## What?
Add ability to hide card holder name field when payment provider doesn't use it

## Why?
For new TD Bank integration we don't need this field and should not show unused fields

## Testing / Proof
Unit tests and manually tested
Screenshot 2023-12-04 at 11 03 25

### Show all fields 
• When options for specific fields wasn't set
Screenshot 2023-12-01 at 18 07 17
• When card holder name field was enabled manually
Screenshot 2023-12-01 at 18 08 06
### Hide card holder name
• When only card holder name field was disabled
Screenshot 2023-12-01 at 18 08 42
Screenshot 2023-12-04 at 10 33 35
Screenshot 2023-12-04 at 10 36 16
• When card holder name was disabled and card code field was disabled
Screenshot 2023-12-01 at 18 09 57
### Card code field was disabled
card holder name field enabled manually or it's value wasn't set
Screenshot 2023-12-01 at 18 09 28

@bigcommerce/team-checkout

### DIFF
--- a/packages/hosted-credit-card-integration/src/components/HostedCreditCardFieldset/HostedCreditCardFieldset.tsx
+++ b/packages/hosted-credit-card-integration/src/components/HostedCreditCardFieldset/HostedCreditCardFieldset.tsx
@@ -1,3 +1,4 @@
+import classNames from 'classnames';
 import React, { FunctionComponent, ReactNode } from 'react';
 
 import { TranslatedString } from '@bigcommerce/checkout/locale';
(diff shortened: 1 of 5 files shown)


packages/hosted-credit-card-integration/src/components/HostedCreditCardFieldset/HostedCreditCardFieldset.tsx
```diff
@@ -34,7 +35,12 @@ const HostedCreditCardFieldset: FunctionComponent<HostedCreditCardFieldsetProps>
             </Legend>
         }
     >
-        <div className="form-ccFields">
+        <div
```
animesh1987, line 38: Is there a test for this?

animesh1987 (approved):

PR #1613, opened 2023-12-08
fix(payment): PI-345 prevent actions with certificate while processing payment

PavlenkoM:
## What?
Remove certificate action button while payment in process

## Why?
To prevent changing amount by shopper after submitting payment

## Testing / Proof
Unit test and manually tested

https://github.com/bigcommerce/checkout-js/assets/9430298/ecf51842-5e22-45ba-a7cd-a9a58cf62a63

@bigcommerce/team-checkout


packages/core/src/app/order/OrderSummarySubtotals.test.tsx
```diff
@@ -37,31 +48,31 @@ describe('OrderSummarySubtotals', () => {
...
         const couponElements = screen.getAllByTestId('cart-coupon');
         const [firstCouponElement, secondCouponElement] = couponElements;
 
         expect(couponElements).toHaveLength(2);
-        expect(firstCouponElement).toHaveTextContent('-$5.00');
+        expect(firstCouponElement).toHaveTextContent('5');
         expect(firstCouponElement).toHaveTextContent('20% off each item');
         expect(firstCouponElement).toHaveTextContent('savebig2015');
-        expect(secondCouponElement).toHaveTextContent('-$5.00');
+        expect(secondCouponElement).toHaveTextContent('5');
```
bc-peng, line 75: Those format should be out of scope and untouched.
Could we update `jest.mock('./OrderSummaryPrice')` to avoid uncessary changes?

PavlenkoM: Hi Peng.
I am agree with you, this formatting and component in general should be out of scope for changes of child component.
In case of unit test we should test each unit separately. So if this format changed inside `OrderSummaryPrice` this logic should be tested inside unit test for `OrderSummaryPrice`.
And `OrderSummarySubtotals.test.tsx` should not test logic of child components or have any other dependency to child components. So this formatting check should not be here at all
But if you think that our unit test should be integration test, I will keep this formatting

PavlenkoM: @bc-peng 
I rolled back this formatting.
Can you look, please?

packages/core/src/app/order/OrderSummarySubtotals.test.tsx
```diff
@@ -23,17 +25,19 @@ describe('OrderSummarySubtotals', () => {
...
-                    coupons={order.coupons}
-                    discountAmount={order.discountAmount}
-                    giftCertificates={order.payments && mapFromPayments(order.payments)}
-                    handlingAmount={order.handlingCostTotal}
-                    shippingAmount={order.shippingCostTotal}
-                    subtotalAmount={order.orderAmount}
-                    taxes={order.taxes}
-                />
-            </LocaleContext.Provider>
+            <CheckoutProvider checkoutService={createCheckoutService()}>
```
bc-peng, line 28: We have introduced a custom render to wrap all contexts.
`import { fireEvent, render, screen } from '@bigcommerce/checkout/test-utils';`
Try it out to avoid making context on your own.

PavlenkoM: Thanks @bc-peng, thanks for this solution. I changed it for my test which uses react testing library

packages/core/src/app/order/OrderSummaryTotal.spec.tsx
```diff
@@ -19,13 +21,15 @@ describe('OrderSummaryTotal', () => {
...
         beforeEach(() => {
             orderSummaryTotal = mount(
-                <LocaleContext.Provider value={localeContext}>
-                    <OrderSummaryTotal
-                        orderAmount={100}
-                        shopperCurrencyCode="USD"
-                        storeCurrencyCode="USD"
-                    />
-                </LocaleContext.Provider>,
+                <CheckoutProvider checkoutService={createCheckoutService()}>
```
bc-peng, line 24: Thanks for updating the testWrapper.
This line and L53 appear to be overlooked.

packages/core/src/app/order/OrderSummaryPrice.spec.tsx
```diff
@@ -1,23 +1,51 @@
...
-    let orderSummaryPrice: ShallowWrapper;
+    let orderSummaryPrice: any;
+    let checkoutService: CheckoutService;
+    let checkoutState: CheckoutSelectors;
+
+    const mountTestComponent = (
+        props: OrderSummaryPriceProps & WithCheckoutOrderSummaryPriceProps & { children?: React.ReactNode }
+    ) => {
+        const orderSummaryPriceContainer = mount(
+            <CheckoutProvider checkoutService={checkoutService}>
```
bc-peng, line 24: Another one.

packages/core/src/app/order/OrderSummaryTotal.spec.tsx
PavlenkoM: Hm,  but I think we should use new render only for new test with react testing library and not for enzyme?

packages/core/src/app/order/OrderSummaryPrice.spec.tsx
PavlenkoM: the same as before, it is a test with enzyme

packages/core/src/app/order/OrderSummaryPrice.tsx
```diff
@@ -146,4 +153,12 @@ class OrderSummaryPrice extends Component<OrderSummaryPriceProps, OrderSummaryPr
...
-export default OrderSummaryPrice;
+function mapToOrderSummaryPriceProps({ checkoutState }: CheckoutContextProps): WithCheckoutOrderSummaryPriceProps {
+    const { statuses: { isSubmittingOrder } } = checkoutState;
+
+    return {
+        isSubmittingOrder: isSubmittingOrder(),
+    };
+}
+
+export default withCheckout(mapToOrderSummaryPriceProps)(OrderSummaryPrice);
```
bc-peng, line 164: Could we avoid using HOC and creating new enzyme tests?
Because we are going to upgrade to react 18 next year, all these work will be outdated at that time.

animesh1987:
Hey @PavlenkoM another note on this PR, instead of hiding the button can we just `disable` it?

PavlenkoM:
Hi @animesh1987 @bc-peng 
As you requested before:
* I migrated this component to functional and migrated unit test for this component to react testing library 
* Also make "remove" action disabled instead of hiding

can you please look?

And also have a question, how can I get color for the button from theme? https://github.com/bigcommerce/checkout-js/pull/1613/files#diff-911fc412c6e9e36a7b387dc31dec0dc4eb19f4927c32fa1b8fbd4db3823dab47R114

packages/core/src/app/order/OrderSummaryPrice.tsx
```diff
@@ -41,109 +42,102 @@ function isNumberValue(displayValue: number | ReactNode): displayValue is number
...
-                            {superscript && (
-                                <sup data-test="cart-price-value-superscript">{superscript}</sup>
+                        )}
+                        {onActionTriggered && actionLabel && (
+                            <span className="cart-priceItem-link">
+                                <button
+                                    data-test="cart-price-callback"
+                                    disabled={isDisabled}
+                                    onClick={preventDefault(onActionTriggered)}
+                                    style={{color: isDisabled ? '#ddd' : '#4496f6'}}
```
bc-peng, line 114: To get the colour from the theme, you might want to use a class name from the theme's CSS file.
I think the related class names are either:
```
.optimizedCheckout-buttonPrimary[disabled],.optimizedCheckout-buttonPrimary[disabled].active,.optimizedCheckout-buttonPrimary[disabled].is-active,.optimizedCheckout-buttonPrimary[disabled].is-loading,.optimizedCheckout-buttonPrimary[disabled]:focus,.optimizedCheckout-buttonPrimary[disabled]:hover
```
or
```
.optimizedCheckout-buttonSecondary[disabled],.optimizedCheckout-buttonSecondary[disabled].active,.optimizedCheckout-buttonSecondary[disabled].is-active,.optimizedCheckout-buttonSecondary[disabled].is-loading,.optimizedCheckout-buttonSecondary[disabled]:focus,.optimizedCheckout-buttonSecondary[disabled]:hover
```
All class names can be found at [here](https://cdn11.bigcommerce.com/s-pyfdp1ymny/stencil/55c3c420-d9d0-013a-571a-12ed63bdf519/css/optimized-checkout-5fef6070-0fdd-013b-da77-062f8167c164.css).

PavlenkoM: Yes, I saw this classes.
But the problem is that for now we display this button like a link `<a href=... />` and `optimizedCheckout-button...` classes is looks like a button. So by adding this classes we will change appearance what is not good.
But for link (`<a />`) we don't have a disable state and styles for this state.

So, @animesh1987 , maybe we can leave it as it is with a `<a href=... />` and hide this button in case user can't click on it?

Examples with current implementation and button classes
Screenshot 2023-12-18 at 12 51 02

animesh1987: Feel free to add disabled styles for `<a />` tag.

PavlenkoM: Hi @animesh1987 @bc-peng 
I have added styles for disabled link(video updated)
can you please check?

packages/core/src/scss/components/foundation/buttons/_buttons.scss
```diff
@@ -60,4 +60,10 @@
     .button--slab {
         width: 100%;
     }
+
+    .link--disabled {
```
bc-peng, line 64: Is it possbile to use L31 to avoid making a new class?

PavlenkoM: No, It is impossible, because of [`<a />`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) doesn't have a 'disabled' state.

bc-peng (approved):

PR #1634, opened 2023-12-27
Revert "CHECKOUT-6524: Fix Sentry client initialization to avoid conflict"

bc-peng:
# What

Reverts bigcommerce/checkout-js#1610.

# Why

Reverting this PR because `this.hub = createHub(clientOptions);` is not working in production and stopping Sentry from logging any issue.

An example of a production error message:
```
Refused to create a worker from 'blob:https://checkout.solostove.com/5837bd07-3018-47d1-8af1-50e16ed86d04' because it violates the following Content Security Policy directive: "script-src 'none'". Note that 'worker-src' was not explicitly set, so 'script-src' is used as a fallback.
```

Screenshot 2023-12-27 at 11 29 13 am


snaderiBC (approved):
@bc-peng Could you explain a bit in the description section why we are reverting and link to the issue? Thanks

bc-peng:
> @bc-peng Could you explain a bit in the description section why we are reverting and link to the issue? Thanks

Sure, updated.

richapatel1510 (approved):

PR #1657, opened 2024-01-12
feat(payment): PAYPAL-3446 Shipping address form isn't validated when Accelerated Checkout is enabled

bc-nick:
## What?

Added `shouldHaveCustomValidation` function to define the validation flow

## Why?

Accelerated Checkout should use the default validation flow

## Testing / Proof

Manual testing

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/shipping/SingleShippingForm.tsx
+++ b/packages/core/src/app/shipping/SingleShippingForm.tsx
@@ -28,6 +28,7 @@ import {
     mapAddressToFormValues,
 } from '../address';
 import { getCustomFormFieldsValidationSchema } from '../formFields';
+import { PaymentMethodId } from '../payment/paymentMethod';
 import { Fieldset, Form } from '../ui/form';
 
 import BillingSameAsShippingField from './BillingSameAsShippingField';
@@ -76,6 +77,14 @@ interface SingleShippingFormState {
     isResettingAddress: boolean;
     isUpdatingShippingData: boolean;
     hasRequestedShippingOptions: boolean;
+}
+
+function shouldHaveCustomValidation(methodId?: string): boolean {
+    return !(
+        methodId === PaymentMethodId.BraintreeAcceleratedCheckout ||
+        methodId === PaymentMethodId.PayPalCommerceAcceleratedCheckout ||
+        !methodId
+    );
 }
 
 export const SHIPPING_AUTOSAVE_DELAY = 1700;
@@ -334,7 +343,7 @@ export default withLanguage(
             getFields,
             methodId,
         }: SingleShippingFormProps & WithLanguageProps) =>
-            methodId
+            shouldHaveCustomValidation(methodId)
                 ? object({
                       shippingAddress: lazy<Partial<AddressFormValues>>((formValues) =>
                           getCustomFormFieldsValidationSchema({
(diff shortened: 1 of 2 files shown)


serhii-tkachenko (approved):

packages/core/src/app/shipping/SingleShippingForm.tsx
```diff
@@ -78,6 +79,14 @@ interface SingleShippingFormState {
     hasRequestedShippingOptions: boolean;
 }
 
+function shouldHaveCustomValidation(methodId?: string): boolean {
+    return !(
```
vitalii-koshovyi, line 83: For me checking methodId at first, seems easier to read 💅
```
 methodId && !(methodId === PaymentMethodId.BraintreeAcceleratedCheckout || methodId === PaymentMethodId.PayPalCommerceAcceleratedCheckout)
```

vitalii-koshovyi (approved):

animesh1987:
Can we please add tests to cover this change.

bc-nick:
> Can we please add tests to cover this change.

@animesh1987 tests have been successfully added. take a look :pray:

packages/core/src/app/shipping/SingleShippingForm.tsx
animesh1987: This is a bit confusing to read, as I see if the methodId is either `PaymentMethodId.BraintreeAcceleratedCheckout` or `PaymentMethodId.PayPalCommerceAcceleratedCheckout` then we should return `false` is that correct?

animesh1987: If that is the case can we move the ids to an array.
```
const methodIdsWithoutCustomValidation = [PaymentMethodId.BraintreeAcceleratedCheckout, PaymentMethodId.PayPalCommerceAcceleratedCheckout];

return methodId && !methodIdsWithoutCustomValidation.includes(methodId);
```

bc-nick: @animesh1987 yours is better. take a look pls

packages/core/src/app/shipping/SingleShippingForm.tsx
```diff
@@ -334,7 +344,7 @@ export default withLanguage(
             getFields,
             methodId,
         }: SingleShippingFormProps & WithLanguageProps) =>
-            methodId
```
bc-peng, line 337: Could you please explain what's wrong with the previous control flow?
It looks to me that now we only apply `getCustomFormFieldsValidationSchema` to two IDs listed here, previously, we apply it to all IDs.

animesh1987: AFAIK we are ignoring `PaymentMethodId.BraintreeAcceleratedCheckout and PaymentMethodId.PayPalCommerceAcceleratedCheckout` from custom validation, hence this method.

Is that the correct understanding @bc-nick ?

bc-nick: We do not need **getCustomFormFieldsValidationSchema** for both methods because it filters by `custom` field and validation works incorrectly

You can take a look here: https://github.com/bigcommerce/checkout-js/blob/aa4bee8cece2b5efe775c731eb6481d4da480c6c/packages/core/src/app/formFields/getCustomFormFieldsValidationSchema.ts#L47

So we need to use default validation

animesh1987 (approved):

PR #1685, opened 2024-02-01
fix(payment): PAYPAL-3519 fixed the issue with analytic tracking of selected payment method on payment step load

serhii-tkachenko:
## What?
Fixed the issue with analytic tracking of selected payment method on payment step load

## Why?
When the form is loaded we have different payment method selected by default, but it is not the same that provided to analytics tracking method on payment step first load. This issue occurs because prev default payment method provided to tracking method instead of actual one.

## Testing / Proof
Manual tests
CI

Before:
Screenshot 2024-02-01 at 12 49 02

After:
Screenshot 2024-02-01 at 12 44 34

### DIFF
--- a/packages/core/src/app/payment/Payment.tsx
+++ b/packages/core/src/app/payment/Payment.tsx
@@ -547,14 +547,14 @@ class Payment extends Component<
 
     private async loadPaymentMethodsOrThrow(): Promise<void> {
         const {
-            defaultMethod,
             loadPaymentMethods,
             onUnhandledError = noop,
         } = this.props;
-        const { selectedMethod = defaultMethod } = this.state;
 
         try {
             await loadPaymentMethods();
+
+            const selectedMethod = this.state.selectedMethod || this.props.defaultMethod;
 
             if (selectedMethod) {
                 this.trackSelectedPaymentMethod(selectedMethod);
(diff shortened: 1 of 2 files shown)


packages/core/src/app/payment/Payment.tsx
```diff
@@ -547,15 +547,15 @@ class Payment extends Component<
...
-            defaultMethod,
             loadPaymentMethods,
             onUnhandledError = noop,
         } = this.props;
-        const { selectedMethod = defaultMethod } = this.state;
 
         try {
             await loadPaymentMethods();
 
+            const selectedMethod = this.state.selectedMethod || this.props.defaultMethod;
```
serhii-tkachenko, line 557: We get `selectedMethod` and `defaultMethod` from `this.state` and `this.props` directly because they may change after async `loadPaymentMethods` method call

PavlenkoM (approved):
Thanks, for fixing the issue with analytics! 👍

packages/core/src/app/payment/Payment.tsx
animesh1987: This seems fine for now, do you think there is an opportunity to refactor this in future?

animesh1987 (approved):

packages/core/src/app/payment/Payment.tsx
serhii-tkachenko: We tried to move this implementation inside tracking method, but it looks a bit complicated. I guess it might be fixed when Payment component will be updated from class component to functional

PR #1689, opened 2024-02-06
feat(payment): PAYPAL-3453 Support saving PPCP PayPal Account to the vault

bc-nick:
## What?

Support saving PPCP PayPal Account to the vault on Checkout page

## Why?

To be able to save PPCP PayPal Account for future transactions

Depends on: https://github.com/bigcommerce/checkout-sdk-js/pull/2352

## Testing / Proof

Screenshot 2024-02-19 at 13 39 00

Manual testing
Unit tests

@bigcommerce/team-checkout


bc-dronov (approved):

packages/paypal-commerce-integration/.eslintrc.json
```diff
@@ -6,6 +6,7 @@
     {
       "files": ["*.ts", "*.tsx"],
       "rules": {
+        "@typescript-eslint/no-unsafe-member-access": "off",
```
serhii-tkachenko, line 9: Why do you need this rule?

packages/paypal-commerce-integration/src/components/PayPalCommercePaymentMethodComponent.tsx
```diff
@@ -57,19 +68,47 @@ const PayPalCommercePaymentMethodComponent: FunctionComponent<
...
     const initializePayment = async () => {
         try {
             await checkoutService.initializePayment({
                 gatewayId: method.gateway,
                 methodId: method.id,
                 [providerOptionsKey]: {
                     container: '#checkout-payment-continue',
-                    onRenderButton: () => {
-                        paymentForm.hidePaymentSubmitButton(method, true);
+                    shouldRenderPayPalButtonOnInitialization: false,
```
serhii-tkachenko, line 109: Should this be false for cases if vaulting feature is disabled or if a guest wants to purchase something with this payment method?

packages/paypal-commerce-integration/.eslintrc.json
bc-nick: https://github.com/bigcommerce/checkout-js/pull/1689/files#diff-a00ef9fab04a2b411a0abba10dc95be92b4769e4e6eb1d225f50cf3a7fa6d8b9R35


Unsafe member access .isComplete on an `any` value

packages/paypal-commerce-integration/src/components/PayPalCommercePaymentMethodComponent.tsx
bc-nick: It should be `false` for both because of the logic of appearing button

packages/paypal-commerce-integration/.eslintrc.json
serhii-tkachenko: I don't think that it is a good idea to set this rule for whole package. I recommend to use 1 line disable rule for such cases or even avoid such rule and fix the issue.

packages/paypal-commerce-integration/src/components/PayPalCommercePaymentMethodComponent.tsx
serhii-tkachenko: I don't really sure. may I ask you to record a proof video with guest flow?

bc-nick: You're right. I am on my way to fixing it. Thank you

serhii-tkachenko: I don't understand what was fixed..

serhii-tkachenko: it should be false only when BC merchant go through checkout steps and has vaulted instruments otherwise it should be true in my opinion.

serhii-tkachenko (approved):

PR #1697, opened 2024-02-13
fix(checkout): Add missing Order Comments input id

elaleprieto:
## What?
Added the missing ID attribute to the Order Comment input

## Why?
To match the label and fix accessibility issues

## Testing / Proof

@bigcommerce/team-checkout


bc-0dp:
@elaleprieto could you amend your commit message to something like `fix(checkout): CHECKOUT-000 Add missing Order Comments input id`

elaleprieto:
Done! ✅

animesh1987 (approved):

Issue #1412, opened 2023-08-02
Accessibility Issue: "Sign in now" link is not focusable


mattcoy-arcticleaf:
"Sign in now" link cannot be accessed with the keyboard via tab as it is an `<a>` tag without an href. This could cause other usability issues as well. It should be a `<button>` tag since it is not actually a link.

To replicate, go to checkout in Cornerstone and tab through the elements. You will see that the Sign in now button never receives focus!

animesh1987:
Hey @mattcoy-arcticleaf thanks for raising this, will have a look at this in next week and get back to on this.

bc-0dp:
@bigcommerce/team-checkout was this resolved?

mattcoy-arcticleaf:
@bc-0dp , nope, this was not resolved.

bc-0dp:
@mattcoy-arcticleaf if you want you are very welcome to create a PR. I've made a draft one for review and consideration.

PR #1698, opened 2024-02-14
fix(checkout): CHECKOUT-000 Add focus to checkout-customer-login

bc-0dp:
## What?
Add `tabIndex` and `role="button"` to `#checkout-customer-index`

## Why?
As proposed in #1412 adding these properties will allow screen readers and keyboard users to focus the 'Sign in now` Link

## Testing / Proof
Screenshot 2024-02-14 at 11 34 21

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/customer/GuestForm.tsx
+++ b/packages/core/src/app/customer/GuestForm.tsx
@@ -102,8 +102,7 @@ const GuestForm: FunctionComponent<
                     <p>
                         <TranslatedString id="customer.login_text" />{' '}
                         <a
-                            href="#"
-                            tabIndex="0"
+                            tabIndex={0}
                             role="button"
                             data-test="customer-continue-button"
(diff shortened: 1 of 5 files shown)


mattcoy-arcticleaf:
@bc-0dp I think this solves it! I like that you defined the role as well :)

animesh1987 (approved):

bc-peng:
There is a `tabindex={0}` now, added by this PR. https://github.com/bigcommerce/checkout-js/pull/881
Perhaps we need to replace one with `tabindex={1}`.
cc @animesh1987

bc-0dp:
@bc-peng According to MDN
> You are recommended to only use 0 and -1 as tabindex values. 

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

animesh1987 (approved):

Issue #1700, opened 2024-02-15
Order Summary option label and value should be visually separated


mattcoy-arcticleaf:
When you add an item to cart with options and reach the checkout, the order summary shows the option label and value on a single line in a single HTML element with no separation. This can be very confusing to customers. These should be separated by a colon, or at least placed into separate HTML tags so a developer can style them differently.

#### Current implementation:
<img width="412" alt="Screenshot 2024-02-15 at 9 00 37 AM" src="https://github.com/bigcommerce/checkout-js/assets/47044676/91f67663-2544-492e-ab38-13e85047e9d2">
<img width="762" alt="Screenshot 2024-02-15 at 9 02 54 AM" src="https://github.com/bigcommerce/checkout-js/assets/47044676/b1f3dac5-b321-46ad-9f1f-9b27104857d9">

#### Proposed implementation 1:
<img width="764" alt="Screenshot 2024-02-15 at 9 16 43 AM" src="https://github.com/bigcommerce/checkout-js/assets/47044676/cae0c583-47c4-471a-8876-d30cba147de7">
<img width="394" alt="Screenshot 2024-02-15 at 9 16 47 AM" src="https://github.com/bigcommerce/checkout-js/assets/47044676/ea5f2dae-455f-47bf-a075-4977010a682e">

#### Proposed implementation 2:
<img width="768" alt="Screenshot 2024-02-15 at 9 18 05 AM" src="https://github.com/bigcommerce/checkout-js/assets/47044676/ebdbf5dd-a9fa-46c4-913b-9b06b7b65491">
<img width="415" alt="Screenshot 2024-02-15 at 9 18 08 AM" src="https://github.com/bigcommerce/checkout-js/assets/47044676/5e38db40-972c-4730-b412-b8a1c514626e">

bc-0dp:
That's cool!
As a minimum split them up in option value spans.

developerscott:
This would be super helpful to us and our clients as well. 
I prefer option 2 (to start) as it gives us more control than option 1.

animesh1987:
Sounds like a nice improvement, we will go with option 2.

PR #1701, opened 2024-02-16
feat(checkout): CHECKOUT-000 Add spans for product-title, product-item-quantity, product-option-label, and product-option-value

bc-0dp:
## What?
As requested by community members here, https://github.com/bigcommerce/checkout-js/issues/1700
we've added wrapping spans around product quantity and product title, as well as adding wrapping spans around product options label and value in the cart-summary.

## Why?
This improves customization options for developers as they can target these values now purely with CSS

## Testing / Proof

Screenshot 2024-02-16 at 11 55 05
Screenshot 2024-02-16 at 11 51 50

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/order/OrderSummaryItem.tsx
+++ b/packages/core/src/app/order/OrderSummaryItem.tsx
@@ -17,7 +17,9 @@ export interface OrderSummaryItemProps {
 
 export interface OrderSummaryItemOption {
     testId: string;
-    content: ReactNode;
+    content?: ReactNode;
+    optionLabel?: ReactNode;
+    optionValue?: ReactNode;
 }
 
 const OrderSummaryItem: FunctionComponent<OrderSummaryItemProps> = ({
@@ -37,7 +39,8 @@ const OrderSummaryItem: FunctionComponent<OrderSummaryItemProps> = ({
                 className="product-title optimizedCheckout-contentPrimary"
                 data-test="cart-item-product-title"
             >
-                {`${quantity} x ${name}`}
+                <span className="product-item-quantity">{quantity}</span>
+                <span className="product-item-name">{name}</span>
             </h4>
             {productOptions && productOptions.length > 0 && (
                 <ul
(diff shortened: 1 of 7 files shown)


packages/core/src/app/order/OrderSummaryItem.tsx
```diff
@@ -37,7 +39,8 @@ const OrderSummaryItem: FunctionComponent<OrderSummaryItemProps> = ({
                 className="product-title optimizedCheckout-contentPrimary"
                 data-test="cart-item-product-title"
             >
-                {`${quantity} x ${name}`}
+                <span className="product-item-quantity">{quantity}</span>
```
animesh1987, line 42: Do we need to change this?

bc-0dp: Technically no. But the community elaborated, and mentioned that it's the same improvement and would be nice to be able to target qty and name separately as well. Enabling dev's to style those individually with CSS without the need to introduce JS to do so

packages/core/src/app/order/OrderSummaryItem.tsx
```diff
@@ -46,7 +49,8 @@ const OrderSummaryItem: FunctionComponent<OrderSummaryItemProps> = ({
                 >
                     {productOptions.map((option, index) => (
                         <li className="product-option" data-test={option.testId} key={index}>
-                            {option.content}
+                            <span className="product-option-label">{option.optionLabel}</span>
```
animesh1987, line 52: Maybe add a check here for content, just in case for customisations? Something like
```
{Boolean(content) ? 
    {option.content} : 
    <>
       <span className="product-option-label">{option.optionLabel}</span>
       <span className="product-option-value">{option.optionValue}</span>
    </>
 }
```

bc-peng:
Hi @bc-0dp , thanks for putting this together and for the follow-up on the review feedback.

This PR has been open for some time, and the codebase has moved on significantly since then.

To keep the PR queue tidy, we're going to close this one.
- If this change is still needed, could you please raise it as a feature request. That way it can be properly prioritized and implemented against the current codebase.

Thanks again for the contribution!

PR #1730, opened 2024-02-29
feat(payment): PAYPAL-3447 Allow editing of Billing Address for eWallet Buttons flow

bc-nick:
## What?

Allow editing of Billing Address for eWallet Buttons flow

## Why?

Because of PayPal is not returning all required billing information for buyer accounts and hence upon order placement shoppers receive an error as they can’t return to the billing step and edit the billing information

## Testing / Proof

Screenshot 2024-03-01 at 16 22 01

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/checkout/getCheckoutStepStatuses.ts
+++ b/packages/core/src/app/checkout/getCheckoutStepStatuses.ts
@@ -124,6 +124,31 @@ const getBillingStepStatus = createSelector(
             };
         }
 
+        const isUsingPaypal =
+            checkout && checkout.payments
+                ? checkout.payments.some(
+                    (payment) =>
+                        [
+                            'braintreepaypal',
+                            'braintreepaypalcredit',
+                            'braintreevenmo',
+                            'paypalcommerce',
+                            'paypalcommercecredit',
+                            'paypalcommercevenmo'
+                        ]
+                            .includes(payment.providerId))
+                : false;
+
+        if (isUsingPaypal) {
+            return {
+                type: CheckoutStepType.Billing,
+                isActive: false,
+                isComplete: hasAddress,
+                isEditable: hasAddress,
+                isRequired: true,
+            };
+        }
+
         return {
             type: CheckoutStepType.Billing,
             isActive: false,


andriiVitvitskyi1990 (approved):

serhii-tkachenko (approved):

packages/core/src/app/checkout/getCheckoutStepStatuses.ts
```diff
@@ -90,16 +90,10 @@ const getBillingStepStatus = createSelector(
             : EMPTY_ARRAY;
     },
     (checkout, billingAddress, billingAddressFields) => {
-        const hasAddress = billingAddress
+        const isComplete = billingAddress
```
animesh1987, line 93: Have we checked what happens for stripe in this case?

packages/core/src/app/checkout/getCheckoutStepStatuses.ts
```diff
@@ -90,16 +90,10 @@ const getBillingStepStatus = createSelector(
...
             ? isValidAddress(billingAddress, billingAddressFields)
             : false;
-        const isUsingWallet =
-            checkout && checkout.payments
-                ? checkout.payments.some(
-                      (payment) => SUPPORTED_METHODS.indexOf(payment.providerId) >= 0,
-                  )
-                : false;
-        const isComplete = hasAddress || isUsingWallet;
```
bc-dronov, line 102: Correct me if I'm wrong. As I understand from the task description, editing of billing address should be available only for PayPal. But as I see this solution affects other providers as well.

bc-dronov (approved):

packages/core/src/app/checkout/getCheckoutStepStatuses.ts
bc-nick: @animesh1987 attached the screen recording with stripe payment process using GP:

https://github.com/bigcommerce/checkout-js/assets/99336044/6b10ea49-c788-45af-beeb-9eb8a8e336a3

bc-peng:
> PayPal is not returning all required billing information for buyer accounts and hence upon order placement shoppers receive an error as they can’t return to the billing step and edit the billing information

Can we upload a video to show the PayPal button e2e workflow after the change?

bc-nick:
> > PayPal is not returning all required billing information for buyer accounts and hence upon order placement shoppers receive an error as they can’t return to the billing step and edit the billing information
> 
> Can we upload a video to show the PayPal button e2e workflow after the change?

@bc-peng attached screen recording of the PayPal button workflow:

https://github.com/bigcommerce/checkout-js/assets/99336044/b824c18f-87fb-42a3-9c33-623546b0d95c

packages/core/src/app/checkout/getCheckoutStepStatuses.ts
bc-peng: Thank you for posting PayPal e2e video in another comment.
Because we are not sure the impact o other payment wallet buttons, does it make sense to limit the change to PayPal only?

bc-nick: As far as i understood this is the general change for all payment wallet buttons. But i think i need to make some clarification about it. I will be back with information.

bc-nick: @bc-peng i have made changes related to PayPal only

bc-peng (approved):
Thank you for your patience and nice work.👏👏

PR #1731, opened 2024-03-01
fix(payment): PI-1671 added translations to the AccountInstrumentSele…

vitalii-koshovyi:
## What?
Added translations to the AccountInstrumentSelect component

## Why?
Due the absence of translations 

## Testing / Proof
image
image

@bigcommerce/team-checkout


serhii-tkachenko (approved):
Nice work 👍

packages/locale/src/translations/en.json
```diff
@@ -320,10 +320,12 @@
             "ideal_bic_required": "You must select an issuer to place the order",
             "instrument_text": "Stored credit cards",
             "instrument_add_card_action": "Use a different card",
+            "instrument_account_number_ending": "Account number ending in: {accountNumber}",
             "instrument_default_ending_in_text": "Card ending in {endingIn}",
             "instrument_ending_in_text": "{cardTitle} ending in {endingIn}",
             "instrument_expired_text": "Expired {expiryDate}",
             "instrument_expires_text": "Expires {expiryDate}",
+            "instrument_issuer": "Issuer",
```
bc-peng, line 328:
```suggestion
            "instrument_issuer": "Issuer: {issuer}",
```
Can we make `instrument_issuer` the same way as `instrument_account_number_ending`?

animesh1987: Think the logic has been changed in the component level in this case.

animesh1987: I agree with @bc-peng that might be good to keep things consistent for the two fields.

vitalii-koshovyi: are you sure about it?
What if I need to use `Issuer` without `:` . Will need to add another translation?

animesh1987 (approved):

packages/locale/src/translations/en.json
animesh1987: Discussed offline with @vitalii-koshovyi and makes sense to keep translation as is. Client components can implement translations correctly. Serves for the case where we might need display something like
`Issuer test`

PR #1739, opened 2024-03-04
fix(payment): PI-1670 fixed default payment method checkbox for the p…

vitalii-koshovyi:
…ayment methods ouside the core package

## What?
Fixed default payment method checkbox for the payment methods ouside the core package

## Why?
"Use this card as the default payment method" checkbox appears even when there aren't previously stored cards.
It happens when the Payment gateway has already stored other payment methods, like SEPA, ACH, etc.
Because the endpoint returns all stored payment instruments for that Payment gateway, and on the checkbox component side we don’t have filtering regarding the type of the instrument. In that case even if shopper doesn't have stored cards, but checkbox component recognises other stored instrument and renders "Use this card as the default payment method" checkbox.
We need to add additional filtering by the stored instrument type on the checkbox component.

## Testing / Proof
Before fix:
image

After fix:
BraintreeACH (BraintreeAchPaymentForm)
image

Adyen(HostedWidgetPaymentComponent):
image
image

Bluesnap (BlueSnapDirectEcpPaymentMethod):
image

Credit card(HostedCreditCardPaymentMethod/CreditCardPaymentMethodComponent):
image

Square([SquareV2Form.tsx](https://github.com/bigcommerce/checkout-js/pull/1739/files#diff-af9f63171a9f550d55c1ac7707657ff501ad412c43940a0d02a130c6cc849959)):
image

@bigcommerce/team-checkout


packages/instrument-utils/src/storedInstrument/StoreInstrumentFieldset/StoreInstrumentFieldset.tsx
```diff
@@ -23,17 +23,10 @@ interface WithStorageSettings {
...
     paymentForm: PaymentFormService,
 ): WithStorageSettings => {
-    const {
-        checkoutState: {
-            data: { getInstruments },
-        },
-    } = context;
-
-    const allInstruments = getInstruments();
+    const allInstruments = props.instruments;
```
serhii-tkachenko, line 29: It might be nicer to get `instruments` from `props` on line 33 and update `allInstruments` with `instruments`.

serhii-tkachenko (approved):
LGTM 👍

animesh1987 (approved):

PavlenkoM (approved):

PR #1742, opened 2024-03-07
feat(payment): PI-1723 [KlarnaV2] Refactor the existing `KlarnaV2` co…

bc-sebastianszafraniec:
…mponents in Checkout JS to use the new checkout payment integration JS API

## What?
Move KlarnaV2 to a separate package

## Why?
To use the new checkout payment integration API

## Testing / Proof
Manually tested

https://github.com/bigcommerce/checkout-js/assets/130039975/fb9f3134-cfca-4afc-935f-0ed441e734a2

Test coverage
Zrzut ekranu 2024-03-7 o 13 27 15

@bigcommerce/team-checkout


packages/klarna-integration/project.json
```diff
@@ -0,0 +1,23 @@
...
+      "options": {
+        "lintFilePatterns": ["packages/klarna-integration/**/*.{ts,tsx}"]
+      }
+    },
+    "test": {
+      "executor": "@nrwl/jest:jest",
+      "outputs": ["coverage/packages/klarna-integration"],
+      "options": {
+        "jestConfig": "packages/klarna-integration/jest.config.js",
+        "passWithNoTests": true
```
serhii-tkachenko, line 18: Do you need this option for current package?

packages/klarna-integration/src/klarnav2/KlarnaV2PaymentMethod.tsx
```diff
@@ -0,0 +1,83 @@
...
+            setValidationSchema={setValidationSchema}
+            signOut={checkoutService.signOutCustomer}
+            {...rest}
+        />
+    );
+};
+
+export default toResolvableComponent<PaymentMethodProps, PaymentMethodResolveId>(
+    KlarnaV2PaymentMethod,
+    [{ gateway: 'klarna' }],
```
serhii-tkachenko, line 82: Should we use id instead gateway here? May I ask you to double check this part?

packages/klarna-integration/src/klarnav2/KlarnaV2PaymentMethod.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+    const {
+        data: { getCheckout, isPaymentDataRequired },
+        statuses: { isLoadingInstruments },
+    } = checkoutState;
+
+    const checkout = getCheckout();
+    const customer = checkoutState.data.getCustomer();
+    const isSignedIn = customer?.isGuest;
+    const isInstrumentFeatureAvailable = !isSignedIn && Boolean(method.config.isVaultingEnabled);
```
serhii-tkachenko, line 53: It might be better to update `isSignedIn` variable name to `isGuestCustomer`

packages/klarna-integration/src/klarnav2/KlarnaV2PaymentMethod.spec.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+        const container = mount(<PaymentMethodTest {...defaultProps} method={method} />);
+        const component = container.find(HostedWidgetPaymentComponent);
+
+        expect(component.props()).toEqual(
+            expect.objectContaining({
+                containerId: `${method.id}Widget`,
+                // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+                deinitializePayment: expect.any(Function),
+                // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
serhii-tkachenko, line 128: Its ok to turn off this eslint rule for whole package for spec and test files. Also, as far as I remember we should use `.test.tsx` instead of `.spec.tsx`

bc-sebastianszafraniec: I will set these rules for whole package and change extension from .spec.tsx to .test.tsx

packages/klarna-integration/src/klarnav2/KlarnaV2PaymentMethod.tsx
bc-sebastianszafraniec: `Id` `klarna` is used by klarnaV1. KlarnaV2 is working with `gateway` `klarna`

packages/klarna-integration/project.json
bc-sebastianszafraniec: Looks like yes. Without it CI checks are failing.

serhii-tkachenko (approved):
LGTM 👍

packages/klarna-integration/project.json
bc-peng: If so, that menas there is no tests inside this package, but we do have a `test.tsx` here?
Can you please double check the root cause of that CI failure?

bc-peng:
Ideally, we should have e2e tests for all integration packages.
Could you please at least post a manual testing video for the payment method?

bc-sebastianszafraniec:
I added video from manual testing

animesh1987 (approved):

bc-peng (approved):

PR #1744, opened 2024-03-08
fix(checkout): CHECKOUT-8049 Use session storage to save subscription status in frontend

animesh1987:
## What?
Use session storage to save subscription status.

## Why?
Once a user checks/unchecks the subscription checkbox and move and moves to next step. We send the relevant call to subscription API, but if the page refreshes we lose the state of subscription.

If the user goes back to customer step the state of checkbox is defaults to initial settings, which can be confusing for the user. 

## Testing / Proof
- CI 
- Screencast 

https://github.com/bigcommerce/checkout-js/assets/7134802/77923f83-73cb-4212-b0ea-a6c97b862a16

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/checkout/Checkout.tsx
+++ b/packages/core/src/app/checkout/Checkout.tsx
@@ -36,6 +36,7 @@ import {
     CustomerViewType,
 } from '../customer';
 import { getSupportedMethodIds } from '../customer/getSupportedMethods';
+import { SubscribeSessionStorage } from '../customer/SubscribeSessionStorage';
 import { EmbeddedCheckoutStylesheet, isEmbedded } from '../embeddedCheckout';
 import { PromotionBannerList } from '../promotion';
 import { hasSelectedShippingOptions, isUsingMultiShipping, StaticConsignment } from '../shipping';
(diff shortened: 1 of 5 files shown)


bc-donfran (approved):
Looks good to me :D good use of session storage.

packages/core/src/app/customer/Customer.tsx
```diff
@@ -406,8 +407,11 @@ class Customer extends Component<CustomerProps & WithCheckoutCustomerProps & Ana
                     : updateSubscriptionWhenUnchecked,
             });
 
+
```
bc-peng, line 410:
```suggestion
```

packages/core/src/app/checkout/Checkout.tsx
```diff
@@ -609,6 +610,8 @@ class Checkout extends Component<
             this.embeddedMessenger.postComplete();
         }
 
+        SubscribeSessionStorage.removeSubscribeStatus();
```
bc-peng, line 613: I see we only remove the storage value here, which seems enough. 🤔

bc-peng (approved):
Nice work👏👏

packages/core/src/app/checkout/Checkout.tsx
animesh1987: Yes, ideally should only remove it when a payment is successful so just removing it before navigating away to order-confirmation page.

bc-peng: I was thinking about the same sessionStorage, but different checkouts. It should be OK.

animesh1987: I think should be ok for this instance.

PR #1757, opened 2024-03-18
refactor(payment): PAYPAL-000 updated analytics tests files from spec.tsx to test.txs

serhii-tkachenko:
## What?
Updated analytics tests files from spec.tsx to test.txs

## Why?
As part of global tests refactoring

## Testing / Proof
CI


animesh1987:
Noticed that we are only changing the file name here, but we will need to rewrite the test with `.tsx` by removing https://github.com/bigcommerce/checkout-js/blob/master/packages/analytics/src/AnalyticsProvider.spec.tsx#L2C10-L2C15 with react testing library equivalent.

animesh1987 (approved):

PR #1775, opened 2024-03-27
feat(payment): PAYPAL-3580 PAYPAL-3591 added PayPal Fastlane Icon implementation

serhii-tkachenko:
## What?
Added PayPal Fastlane Icon implementation

## Why?
As part of PayPal Connect to PayPal Fastlane rebranding

## Testing / Proof
Unit tests
Manual tests
CI

Screenshot 2024-03-27 at 17 53 42


packages/core/src/app/address/StaticAddress.test.tsx
```diff
@@ -18,7 +18,7 @@ import StaticAddress, { StaticAddressProps } from './StaticAddress';
 
 jest.mock('@bigcommerce/checkout/ui', () => ({
     ...jest.requireActual('@bigcommerce/checkout/ui'),
-    IconPayPalConnectSmall: () => (<div data-test="pp-fastlane-icon">IconPayPalFastlaneSmall</div>)
+    IconPayPalConnectSmall: () => (<div data-test="pp-connect-icon">IconPayPalFastlaneSmall</div>)
```
animesh1987, line 21: We don't need snapshot tests anymore, can we please refactor this test to get rid of enzyme.

serhii-tkachenko: Let me make it in separate PR, since this feature is a bit urgent for us

bc-dronov (approved):

packages/paypal-fastlane-integration/src/is-paypal-fastlane-address.ts
```diff
@@ -28,9 +28,14 @@ const isEqualToPayPalFastlaneAddress = (
     paypalFastlaneAddress: Partial<ComparableAddress>,
 ): boolean => isEqual(normalizeAddress(bcAddress), normalizeAddress(paypalFastlaneAddress));
 
-const isPayPalFastlaneAddress = (address: Address, addresses: Address[]): boolean =>
-    addresses.some((paypalFastlaneAddress) =>
+const isPayPalFastlaneAddress = (address: Address, addresses?: Address[]): boolean => {
+    if (!addresses || addresses.length === 0) {
```
bc-dronov, line 32: `if (!addresses || !addresses.length)`

bc-nick (approved):

packages/paypal-fastlane-integration/src/is-paypal-fastlane-address.ts
serhii-tkachenko: `addresses.length === 0` looks more recognizable to me, so I can easily detect why I need this condition. I prefer to keep it as it is, if you don't mind

animesh1987 (commented):
For this PR can we please rename the `****.test.tsx` files to `****.spec.tsx` since test suffix is supposed to be for files that use react testing library.

animesh1987 (approved):

PR #1785, opened 2024-04-03
feat(extension): CHECKOUT-000 Pass consignment includes to reload checkout command

animesh1987:
## What?
Pass consignment includes to reload checkout command.

## Why?
Based on feedback of the extension is able to select shipping option, then reload command can take the shopper back to shipping step since shipping options are not reloaded on calling `reloadCheckout` command.

## Testing / Proof
- CI

@bigcommerce/team-checkout

### DIFF
--- a/packages/checkout-extension/src/handlers/createReloadCheckoutHandler.ts
+++ b/packages/checkout-extension/src/handlers/createReloadCheckoutHandler.ts
@@ -8,7 +8,12 @@ export function createReloadCheckoutHandler({
     return {
         commandType: ExtensionCommandType.ReloadCheckout,
         handler: () => {
-            void checkoutService.loadCheckout(checkoutService.getState().data.getCheckout()?.id);
+            void checkoutService.loadCheckout(checkoutService.getState().data.getCheckout()?.id, {
+                params: {
+                    // eslint-disable-next-line
+                    include: ['consignments.availableShippingOptions'] as any,
+                },
+            });
         },
     };
 }


packages/checkout-extension/src/handlers/createReloadCheckoutHandler.ts
```diff
@@ -8,7 +8,12 @@ export function createReloadCheckoutHandler({
     return {
         commandType: ExtensionCommandType.ReloadCheckout,
         handler: () => {
-            void checkoutService.loadCheckout(checkoutService.getState().data.getCheckout()?.id);
+            void checkoutService.loadCheckout(checkoutService.getState().data.getCheckout()?.id, {
+                params: {
+                    // eslint-disable-next-line
+                    include: ['consignments.availableShippingOptions'] as any,
```
davidchin, line 14:
```suggestion
                    include: [CheckoutIncludes.AvailableShippingOptions],
```

davidchin (commented):
Also, can you check if it fixes the issue @petercossey's describing?

packages/checkout-extension/src/handlers/createReloadCheckoutHandler.ts
animesh1987: I was going by this https://github.com/bigcommerce/checkout-js/blob/master/packages/core/src/app/checkout/Checkout.tsx#L203 but just tried to use the enum and seems to work fine. Seems like we have fixed the enum export issue overtime 😄

animesh1987: For some reason this enum usage throws an undefined error where `CheckoutIncludes` is not defined. Might need to investigate it later.

davidchin (approved):

animesh1987:
Completed as part of this https://github.com/bigcommerce/checkout-js/pull/2146

PR #1799, opened 2024-04-10
feat(payment): PAYPAL-3997 added delay for PPCP credit card spinner

bc-dronov:
## What?
Added delay for PPCP Credit Card spinner

## Why?
We should have this delay, because there is a strange pre-loaders from PayPal sdk when we initialize a strategy.
We can't remove it using PayPal api, so for now we hide it with a little loader delay.

## Testing / Proof
Before changes:

https://github.com/bigcommerce/checkout-js/assets/130665807/143394f5-2e21-42e4-ad07-0bde6bb38ee7

After changes:

https://github.com/bigcommerce/checkout-js/assets/130665807/6089e24b-80ae-470b-a4c4-b37d44ec5b0d

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/payment/paymentMethod/PaypalCommerceCreditCardPaymentMethod.tsx
+++ b/packages/core/src/app/payment/paymentMethod/PaypalCommerceCreditCardPaymentMethod.tsx
@@ -1,4 +1,4 @@
-import React, { FunctionComponent, useCallback } from 'react';
+import React, { FunctionComponent, useCallback, useEffect, useState } from 'react';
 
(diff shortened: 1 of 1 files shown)


serhii-tkachenko (changes requested):
I think we should find another way how to deal with PayPal Card Fields loading issue, because PayPal Commerce Credit Cards strategy initialisation may take more then 1900 ms for customers with slow internet connection, so the issue will be showed to them.. It's more about hiding an issue instead of fixing it.

PR #1800, opened 2024-04-11
fix(checkout): CHECKOUT-8183 set shouldSaveAddress flag to false if it's from selected address

bc-shawnwang:
## What?
bump checkout sdk to include

https://github.com/bigcommerce/checkout-sdk-js/pull/2462

## Why?
include the changes of sdk to set shouldSaveAddress flag to false if it's from selected address

## Testing / Proof

@bigcommerce/team-checkout

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
-        "@bigcommerce/checkout-sdk": "^1.582.0",
+        "@bigcommerce/checkout-sdk": "^1.582.3",
         "@bigcommerce/citadel": "^2.15.1",
         "@bigcommerce/form-poster": "^1.2.2",
         "@bigcommerce/memoize": "^1.0.0",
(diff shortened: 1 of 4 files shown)


packages/core/src/app/shipping/Shipping.tsx
```diff
@@ -244,6 +246,7 @@ class Shipping extends Component<ShippingProps & WithCheckoutShippingProps, Ship
             !isEqualAddress(updatedShippingAddress, billingAddress) &&
             !hasRemoteBilling
         ) {
+            console.log('updated shipping address value', updatedShippingAddress);
```
bc-shawnwang, line 249: if `billingSameAsShipping` and other condition met, it will use shipping address to update billing address

packages/core/src/app/shipping/SingleShippingForm.tsx
```diff
@@ -270,8 +270,11 @@ class SingleShippingForm extends PureComponent<
         this.setState({ isResettingAddress: true });
 
         try {
+            address.shouldSaveAddress = false;
             await updateAddress(address);
 
+            // await updateAddress({...address, shouldSaveAddress: false})
```
bc-shawnwang, line 276: billing address will use the value if we selected "same as billing address"

animesh1987 (approved):

bc-peng (approved):

PR #1834, opened 2024-04-22
feat(payment): PI-1961 E2E tests for `checkoutcom` integration

bc-ania:
## What?
E2E tests for `checkoutcom` integration

## Testing / Proof
tests

### DIFF
--- a/packages/checkoutcom-integration/e2e/__har__/Credit-Card-with-Checkoutcom-in-Payment-Step_3718073430/recording.har
+++ b/packages/checkoutcom-integration/e2e/__har__/Credit-Card-with-Checkoutcom-in-Payment-Step_3718073430/recording.har
@@ -1,0 +1,1714 @@
+{
+  "log": {
+    "_recordingName": "Credit Card with Checkoutcom in Payment Step",
+    "creator": {
+      "comment": "persister:CustomFSPersister",
+      "name": "Polly.JS",
+      "version": "6.0.5"
+    },
+    "entries": [
+      {
+        "_id": "3f87bb23d79c8bef5c6840e20cf65aac",
+        "_order": 0,
+        "cache": {},
+        "request": {
+          "bodySize": 0,
+          "cookies": [],
+          "headers": [
+            {
+              "name": "accept-language",
+              "value": "en-US"
+            },
+            {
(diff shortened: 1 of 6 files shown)


packages/test-framework/src/fixture/pagePreset/PaymentStepAsGuestEGP.ts
bc-peng, line 1: I see we have a `PaymentStepAsGuestEU.ts` and here comes another currency.
Is it possbile to refactor `PaymentStepAsGuest.ts` so that it can accept a currency as a prop?

vitalii-koshovyi: Agree with @bc-peng here. You can create a preset where you gonna pass currency and shipping/billing country code.

vitalii-koshovyi (approved):

packages/checkoutcom-integration/src/CheckoutcomCustomPaymentMethod.tsx
```diff
@@ -0,0 +1,64 @@
...
+}
+
+const CheckoutcomCustomPaymentMethod: FunctionComponent<PaymentMethodProps> = ({
+    language,
+    method,
+    checkoutService,
+    checkoutState,
+    ...rest
+}) => {
+    const checkoutCustomMethod = method.id as checkoutcomPaymentMethods;
```
animesh1987, line 28: Can we avoid `as` here

packages/checkoutcom-integration/src/CheckoutcomCustomFormFields.test.tsx
```diff
@@ -8,28 +8,19 @@ import { EventEmitter } from 'events';
 import { Formik } from 'formik';
 import { noop } from 'lodash';
 import React, { FunctionComponent } from 'react';
-
+import {
```
animesh1987, line 11: Since this test is still using enzyme, I will suggest we keep the suffix to be `spec` as compared to `test`. Since we want to keep `test` for RTL tests and use that for migrating tests.

animesh1987 (approved):

PR #1841, opened 2024-04-24
feat(payment): PAYPAL-4066 localizeAddress moved to shared package

bc-dronov:
## What?
localizeAddress moved from `core` to shared `locale` package

## Why?
Because we need to use this method out of `core` package

## Testing / Proof
All tests have been passed

@bigcommerce/team-checkout

### DIFF
--- a/packages/locale/src/address.mock.ts
+++ b/packages/locale/src/address.mock.ts
@@ -1,0 +1,19 @@
+import { Address } from '@bigcommerce/checkout-sdk';
+
+export function getAddress(): Address {
+    return {
+        firstName: 'Test',
+        lastName: 'Tester',
+        company: 'Bigcommerce',
+        address1: '12345 Testing Way',
+        address2: '',
+        city: 'Some City',
+        stateOrProvince: 'California',
+        stateOrProvinceCode: 'CA',
+        country: 'United States',
+        countryCode: 'US',
+        postalCode: '95555',
+        phone: '555-555-5555',
+        customFields: [],
+    };
+}
--- a/packages/locale/src/countries.mock.ts
+++ b/packages/locale/src/countries.mock.ts
@@ -1,0 +1,41 @@
+import { Country } from '@bigcommerce/checkout-sdk';
+
+export function getCountries(): Country[] {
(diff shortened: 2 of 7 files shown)


andriiVitvitskyi1990 (approved):

packages/locale/src/localizeAddress.ts
```diff
@@ -1,7 +1,10 @@
 import { Address, Country } from '@bigcommerce/checkout-sdk';
 import { find, isEmpty } from 'lodash';
 
-import { LocalizedGeography } from '../geography';
+interface LocalizedGeography {
```
serhii-tkachenko, line 4: Should we get access ti this interface outside of this file? (somewhere in other package)

packages/locale/src/address.mock.ts
```diff
@@ -0,0 +1,19 @@
+import { Address } from '@bigcommerce/checkout-sdk';
+
+export function getAddress(): Address {
```
serhii-tkachenko, line 3: payment-integration-test-utils package already has this mock, no need to duplicate it, better import it from there

packages/locale/src/countries.mock.ts
```diff
@@ -0,0 +1,41 @@
+import { Country } from '@bigcommerce/checkout-sdk';
+
+export function getCountries(): Country[] {
```
serhii-tkachenko, line 3: payment-integration-test-utils package already has this mock, no need to duplicate it, better import it from there

serhii-tkachenko (approved):

packages/locale/src/address.mock.ts
bc-dronov: This package is located in checkout-sdk project but not here.

packages/locale/src/countries.mock.ts
bc-dronov: This package is located in checkout-sdk project but not here.

packages/locale/src/localizeAddress.ts
bc-dronov: I've checked `LocalizedGeography` uses only `localizeAddress` function

animesh1987 (approved):

PR #1844, opened 2024-04-26
feat(payment): PAYPAL-3520 Max capture amount failure

bc-nick:
## What?

Handling an specific error related to max capture failure

## Why?

To show the error message with an appropriate information for users

Depends on: https://github.com/bigcommerce/checkout-sdk-js/pull/2485

## Testing / Proof

Manual testing

@bigcommerce/team-checkout


bc-dronov (approved):

packages/core/src/app/payment/paymentMethod/HostedPaymentMethod.tsx
```diff
@@ -168,6 +164,34 @@ class HostedPaymentMethod extends Component<
...
+
+    private getPaymentOptions(): PaymentInitializeOptions {
+        const { onUnhandledError, method, language } = this.props;
+
+        const options = {
+            gatewayId: method.gateway,
+            methodId: method.id,
+        };
+
+        if (method.id === 'braintreepaypal') {
```
bc-peng, line 176: I think we agreed to avoid adding further payment method specific logic to `core`.
Is it possbile to place this in its own repo?

bc-nick: @bc-peng yes, it is possible. I have updated PR with an appropriate changes. Take a look pls

animesh1987 (approved):

PR #1875, opened 2024-05-27
fix(payment): PI-2103 fix bluesnap direct styling

vitalii-koshovyi:
## What?
Fixed hosted input styling for providers that doesn't need cvv verification

## Why?
Due to our mechanics to get Stencil styles in the checkout we need to mount hidden input and from that we're taking styles.
We use cvv input as an element where we should mount that hidden input for the stored card .
Some providers like BluesnapDirect don't have such input so we should mount it near the card number input.

## Testing / Proof
Before fix:
Screenshot 2024-05-28 at 16 29 23

After fix:
Screenshot 2024-05-28 at 16 24 30

@bigcommerce/team-checkout

### DIFF
--- a/packages/hosted-credit-card-integration/src/HostedCreditCardPaymentMethod.tsx
+++ b/packages/hosted-credit-card-integration/src/HostedCreditCardPaymentMethod.tsx
@@ -64,11 +64,17 @@ const HostedCreditCardPaymentMethod: FunctionComponent<PaymentMethodProps> = ({
             const isInstrumentCardCodeRequired = selectedInstrument
                 ? isInstrumentCardCodeRequiredProp(selectedInstrument, method)
                 : false;
-            const styleContainerId = selectedInstrument
-                ? isInstrumentCardCodeRequired
-                    ? getHostedFieldId('ccCvv')
-                    : undefined
-                : getHostedFieldId('ccNumber');
+            let styleContainerId;
+
+            if (selectedInstrument) {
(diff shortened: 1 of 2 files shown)


packages/hosted-credit-card-integration/src/HostedCreditCardPaymentMethod.tsx
```diff
@@ -64,11 +64,17 @@ const HostedCreditCardPaymentMethod: FunctionComponent<PaymentMethodProps> = ({
             const isInstrumentCardCodeRequired = selectedInstrument
                 ? isInstrumentCardCodeRequiredProp(selectedInstrument, method)
                 : false;
-            const styleContainerId = selectedInstrument
-                ? isInstrumentCardCodeRequired
-                    ? getHostedFieldId('ccCvv')
-                    : undefined
-                : getHostedFieldId('ccNumber');
+            let styleContainerId;
```
animesh1987, line 67: Can we please update tests to cover this change.

vitalii-koshovyi: @animesh1987 yeah, forgot about it. I haven't refactored whole test file, due to small estimation of this ticket. We can create a separate ticket for the refactoring to RTL

bc-peng (approved):
LGTM.

richapatel1510 (approved):

animesh1987 (approved):

PR #1884, opened 2024-06-05
fix(payment): PAYPAL-4248 updated customer step loading state to be able to continue with checkout even if wallet button is in loading state

serhii-tkachenko:
## What?
Updated customer step loading state to be able to continue with checkout even if wallet button is in loading state

## Why?
Some users have pretty bad internet connection which might increase time for waiting on wallet buttons initialization. Therefore, it might be better to make wallet buttons loading state as an optional, to avoid blocking user from checkout.

## Testing / Proof
Manual tests was made with Network throttling 3g slow to emulate slow internet connection
CI

[Before] Continue button is disabled until wallet buttons finish their initialisation process: 

https://github.com/bigcommerce/checkout-js/assets/25133454/00b9c3e5-1965-4824-9231-2fbdf9b7839d

[After] Continue button is not disabled and can be used by customer even if wallet buttons still initialising on the page:

https://github.com/bigcommerce/checkout-js/assets/25133454/d8c55ebf-56e6-4467-bed7-27bd8b5e865c

### DIFF
--- a/packages/core/src/app/customer/Customer.tsx
+++ b/packages/core/src/app/customer/Customer.tsx
@@ -206,9 +206,7 @@ class Customer extends Component<CustomerProps & WithCheckoutCustomerProps & Ana
             onError={onUnhandledError}
           />;
 
-        const isLoadingGuestForm = isWalletButtonsOnTop ?
-            isContinuingAsGuest || isExecutingPaymentMethodCheckout :
-            isContinuingAsGuest || isInitializing || isExecutingPaymentMethodCheckout;
+        const isLoadingGuestForm = isContinuingAsGuest || isExecutingPaymentMethodCheckout;
 
         return (
             shouldRenderStripeForm ?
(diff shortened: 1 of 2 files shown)


packages/core/src/app/customer/Customer.tsx
```diff
@@ -206,9 +206,7 @@ class Customer extends Component<CustomerProps & WithCheckoutCustomerProps & Ana
             onError={onUnhandledError}
           />;
 
-        const isLoadingGuestForm = isWalletButtonsOnTop ?
-            isContinuingAsGuest || isExecutingPaymentMethodCheckout :
-            isContinuingAsGuest || isInitializing || isExecutingPaymentMethodCheckout;
+        const isLoadingGuestForm = isContinuingAsGuest || isExecutingPaymentMethodCheckout;
```
animesh1987, line 209: I think we still need the `isWalletButtonsOnTop` check.

serhii-tkachenko: I don't think so, because we have the same condition for both cases

animesh1987: Do we need to update tests as part of this?

serhii-tkachenko: Updated PR with extra tests

animesh1987 (approved):

PR #1886, opened 2024-06-06
feat(payment): PI-2177 Pay with Amazon Button display issues at checkout

bc-sebastianszafraniec:
## What?
Adjusted AmazonPay button styling.

## Why?
After changing button design back to one with microtext, at Customer step the microtext is overlaping the button. This only occurs in Customer step as here we are setting all accelerated checkout buttons (paypal, gpay, amazonPay) height to 36px. 

**Before:**
Zrzut ekranu 2024-06-6 o 11 48 25

**After:**
Zrzut ekranu 2024-06-6 o 12 15 36

## Testing / Proof
Tested manually

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/scss/components/checkout/checkoutRemote/_checkoutRemote.scss
+++ b/packages/core/src/scss/components/checkout/checkoutRemote/_checkoutRemote.scss
@@ -17,6 +17,7 @@
 
     > .AmazonPayContainer {
         width: remCalc(200px);
+        height: 51px !important;
         #amazonpayCheckoutButton > div{
             width: 100% !important;
             height: 51px !important;
--- a/packages/core/src/scss/components/checkout/checkoutRemote/_checkoutRemote.scss
+++ b/packages/core/src/scss/components/checkout/checkoutRemote/_checkoutRemote.scss
@@ -19,6 +19,7 @@
         width: remCalc(200px);
         #amazonpayCheckoutButton > div{
             width: 100% !important;
+            height: 51px !important;
         }
     }
 


vitalii-koshovyi (approved):

packages/core/src/scss/components/checkout/checkoutRemote/_checkoutRemote.scss
```diff
@@ -19,6 +19,7 @@
         width: remCalc(200px);
         #amazonpayCheckoutButton > div{
             width: 100% !important;
+            height: 51px !important;
```
animesh1987, line 22: Is this button for top of checkout page? I believe we want to have a specific height for all the buttons.

bc-sebastianszafraniec: Yes, this is for AmazonPay button in Customer step. In this place all buttons have height set to 36px (rectangular yellow part of AmazonPay button too). Recently we changed AmazonPay button design to one with microtext and as I know it was Amazon demand. Previously in all places AmazonPay button had design without microtext. 
We need this change to move microtext a little bit down from the rectangular part because without it the microtext is overflowing it.

animesh1987:
Thanks @bc-sebastianszafraniec for the detailed information. I will say if possible we should remove the microtext to conform to the layout of buttons in this case.

cc @vitalii-koshovyi

vitalii-koshovyi:
@animesh1987 we couldn't remove it, it's a recommendation from Amazon. We recently had a specific ticket to return it😁

bc-sebastianszafraniec:
@animesh1987 As Vitaliy mentioned we need to keep the microtext on AmazonPay button. Please find my solution in last commit. 
**Here is short video how it behaves.**

https://github.com/bigcommerce/checkout-js/assets/130039975/9ef885c8-b08e-4804-9ac7-2cd73ab5f65d

animesh1987 (approved):

PR #1900, opened 2024-06-19
feat(checkout): CHECKOUT-2368 Display Custom Shipping Method

bc-peng:
## What?
Display custom shipping method created by manual order UI on the storefront checkout page.

## Why?
Support the use of a custom shipping method.

### When the experiment is ON
Opening a draft order with a custom shipping method will jump to the payment step directly.

https://github.com/bigcommerce/checkout-js/assets/88361607/78e2ac4b-1245-41d4-9897-d39b820e6cc3

### When the experiment is OFF
The edit button of the shipping step appears on the page.

image

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/checkout/Checkout.test.tsx
+++ b/packages/core/src/app/checkout/Checkout.test.tsx
@@ -16,7 +16,10 @@ import {
 } from '@bigcommerce/checkout/analytics';
 import { ExtensionProvider } from '@bigcommerce/checkout/checkout-extension';
 import { getLanguageService, LocaleProvider } from '@bigcommerce/checkout/locale';
-import { CHECKOUT_ROOT_NODE_ID, CheckoutProvider } from '@bigcommerce/checkout/payment-integration-api';
+import {
+    CHECKOUT_ROOT_NODE_ID,
+    CheckoutProvider,
+} from '@bigcommerce/checkout/payment-integration-api';
 import { CheckoutPageNodeObject } from '@bigcommerce/checkout/test-framework';
 
 import { createErrorLogger } from '../common/error';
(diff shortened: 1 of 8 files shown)


packages/core/src/app/checkout/Checkout.test.tsx
```diff
@@ -255,6 +258,20 @@ describe('Checkout', () => {
...
 
+        it('renders custom shipping method and locks shipping component', async () => {
+            checkout.use('CartWithCustomShippingAndBilling');
+
+            render(<CheckoutTest {...defaultProps} />);
+
+            await checkout.waitForPaymentStep();
+
+            expect(screen.getByText('Manual Order Custom Shipping Method')).toBeInTheDocument();
+            expect(screen.getAllByRole('button', { name: 'Edit' })).toHaveLength(2);
```
bc-peng, line 269: We are asserting these two `Edit` buttons here.

image

packages/core/src/app/checkout/getCheckoutStepStatuses.ts
```diff
@@ -180,12 +180,18 @@ const getShippingStepStatus = createSelector(
             cart && consignments ? hasUnassignedLineItems(consignments, cart.lineItems) : true;
         const isComplete = hasAddress && hasOptions && !hasUnassignedItems;
         const isRequired = itemsRequireShipping(cart, config);
+        const isCompleteWithCustomShipping =
```
animesh1987, line 183: Can we please just call it `hasCustomShippingSelected`

packages/core/src/app/common/utility/isCustomShippingEnabled.ts
```diff
@@ -0,0 +1,9 @@
+import { CheckoutSettings } from '@bigcommerce/checkout-sdk';
+
+export default function isCustomShippingEnabled(
```
animesh1987, line 3: Can we refactor this and have it generic so we can pass an experiment name and fallback is `true`.

packages/core/src/app/shipping/hasSelectedShippingOptions.ts
```diff
@@ -9,12 +9,14 @@ export default function hasSelectedShippingOptions(consignments: Consignment[]):
...
-            ).length,
+            (consignment.selectedShippingOption &&
+                consignment.selectedShippingOption.id &&
+                // Selected option is available
+                consignment.availableShippingOptions &&
+                consignment.availableShippingOptions.filter(
+                    ({ id }) => id === consignment.selectedShippingOption?.id,
+                ).length) ||
+            (consignment.selectedShippingOption?.type === 'custom' &&
+                consignment.selectedShippingOption.cost >= 0),
```
animesh1987, line 20: Don't think we need the cost comparison.

bc-peng: Thanks for the feedback.
All comments have been addressed in the last commit, https://github.com/bigcommerce/checkout-jas/pull/1900/commits/b83b798c9769fe3c9ec5e5d87cc5270312bda14a.

animesh1987 (approved):

PR #1903, opened 2024-06-21
refactor(checkout): PI-2297 moved WalletButton component to separate package

vitalii-koshovyi:
## What?
Moved WalletButton Component to the separate package.
I haven't removed WalletButton Component from the core yet. It will be removed after the refactoring of the "core" payment methods that use him.
It's easier to review when you compare new module with [older one](https://github.com/bigcommerce/checkout-js/blob/master/packages/core/src/app/payment/paymentMethod/WalletButtonPaymentMethod.tsx)
## Why?
Due to the monorepo refactoring, it needed for further GooglePay, VisaCheckout, etc refactoring.

## Testing / Proof
image

@bigcommerce/team-checkout

### DIFF
--- a/packages/ui/src/loading/LoadingSpinner.tsx
+++ b/packages/ui/src/loading/LoadingSpinner.tsx
@@ -10,7 +10,12 @@ const LoadingSpinner: FunctionComponent<LoadingSpinnerProps> = ({ isLoading }) =
     }
 
     return (
-        <div className="loadingSpinner loadingOverlay-container" style={{ height: 100 }}>
+        <div
+            aria-busy="true"
+            className="loadingSpinner loadingOverlay-container"
+            role="status"
+            style={{ height: 100 }}
+        >
             <div className="loadingOverlay optimizedCheckout-overlay" />
         </div>
     );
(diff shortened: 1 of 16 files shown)


packages/ui/src/loading/LoadingSpinner.tsx
```diff
@@ -10,7 +10,11 @@ const LoadingSpinner: FunctionComponent<LoadingSpinnerProps> = ({ isLoading }) =
     }
 
     return (
-        <div className="loadingSpinner loadingOverlay-container" style={{ height: 100 }}>
+        <div
+            className="loadingSpinner loadingOverlay-container"
+            data-test="loading-spinner"
```
bc-peng, line 15: Instead of utilizing test-id, can we create a user accessibility role and use it for testing?

bc-peng: The rest looks good to me.

vitalii-koshovyi: Yeah, I've tried to do that but it's not comfortable to work with it in that case.
We have a loading spinner and have loading overlay(parent element) and I don't see any variants how to rearrange roles in that case. 
So, I'll add `role="status"` to spinner element, and will leave LoadingOverlay element with test-id.
If you have better ideas, please share

bc-peng: Thank you.
I think `role="status"` is good as we don't have any other status role component, then it can be asserted by `screen.getByRole('status');`.

bc-peng (approved):
LGTM 👏👏

PR #1916, opened 2024-07-02
chore(other): CHECKOUT-8410 Update to use node 20

animesh1987:
## What?
Update to use node 20.

## Why?
Node 18 is already in maintenance mode, so it makes sense to move to next active version.

## Testing / Proof
- CI

@bigcommerce/team-checkout @bigcommerce/team-payments


richapatel1510 (approved):
👏

bc-peng (approved):

harnerdesigns:
What version of Node 20 comes with NPM 9? `nvm use 20` right now installs `node v20.15.1 (npm v10.7.0)`

Also the `.nvmrc` is now incorrect after this PR.

PR #1926, opened 2024-07-15
feat(checkout): CHECKOUT-8300 Improve Extension Rendering Performance

bc-peng:
## What?
Improve the rendering time of a checkout extension.

### Limitations
- Applicable to regions not initially visible on the checkout page only.
- Requires an extension to enable server-side cache headers (e.g., `Cache-Control: public, max-age=3600`) for modern browsers to load the extension from cache on subsequent visits.

## Why?
Enhance the user experience of the checkout extension.

### Manual Testing

https://github.com/user-attachments/assets/76ada22a-e16d-4bab-a31b-5c28d0c0d7ef

@bigcommerce/team-checkout

### DIFF
--- a/packages/checkout-extension/src/ExtensionService.ts
+++ b/packages/checkout-extension/src/ExtensionService.ts
@@ -32,22 +32,36 @@ export class ExtensionService {
             return;
         }
 
+        const disabledPreloadRegions = [
+            ExtensionRegion.SummaryAfter,
+            ExtensionRegion.SummaryLastItemAfter,
+        ];
+
         extensions?.forEach((extension) => {
+            if (disabledPreloadRegions.includes(extension.region)) {
+                return;
+            }
+
             const url = new URL(extension.url);
 
             url.searchParams.set('extensionId', extension.id);
             url.searchParams.set('cartId', cartId);
             url.searchParams.set('parentOrigin', parentOrigin);
 
-            const link = document.createElement('link');
+            const iframe = document.createElement('iframe');
 
-            link.rel = 'preload';
-            link.as = 'document';
-            link.href = url.toString();
+            iframe.src = url.toString();
+            iframe.style.display = 'none';
 
-            const head = document.head;
+            const removeIframe = () => {
+                if (iframe.parentNode) {
+                    iframe.parentNode.removeChild(iframe);
+                }
+            };
 
-            head.appendChild(link);
+            iframe.addEventListener('load', removeIframe);
+
+            document.body.appendChild(iframe);
         });
     }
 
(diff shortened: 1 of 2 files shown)


animesh1987 (approved):
Looks good to me @bc-peng, @davidchin can you please have a look as well.

packages/checkout-extension/src/ExtensionService.ts
```diff
@@ -32,22 +32,36 @@ export class ExtensionService {
...
+
+            iframe.src = url.toString();
+            iframe.style.display = 'none';
 
-            link.rel = 'preload';
-            link.as = 'document';
-            link.href = url.toString();
+            const removeIframe = () => {
+                if (iframe.parentNode) {
+                    iframe.parentNode.removeChild(iframe);
```
davidchin, line 58: Instead of loading the iframe in a temporary location, removing it, and then adding it back in the actual location, is it possible to just hide and show iframe that should be preloaded. This way we don't have the limitation of needing TTL caching, which may not work for extensions that have dynamic content.

bc-peng: > just hide and show iframe that should be preloaded

Thanks for your feedback @davidchin.

It appears that when `checkout-js` renders a region's container or its parent component (e.g., the selected shipping method region), that's when we need to present the extension. 

Therefore, using the hide and show method might not have the same effect as loading an extension at page load.

Additionally, we previously tested that moving an iframe to a different DOM element results in reloading, so the hide and show approach won't work in this case either.

davidchin: I thought the intention is to load the extension in the background and _show_ it immediately using CSS. As long as we don't attach the iframe to a different parent, it shouldn't reload. I suppose the trick is to make the extension region persistent so they don't remount when checkout steps are collapsed and expanded.

The concern I have with the current approach is that it may have a performance benefit for those extensions that can utilise TTL cache by warming up the cache, but it'd have a performance impact on those extensions that can't as they need to be loaded twice.

bc-peng: > load the extension in the background and show it immediately using CSS. 


> the trick is to make the extension region persistent so they don't remount when checkout steps are collapsed and expanded.

I explored this during an early spike trial. After positioning the region with CSS absolute positioning, the major challenge was creating a placeholder that occupies the same space as the extension in the actual area (e.g., a shipping method) and keeping the size in sync whenever either side changes its size. This approach felt hacky.

I agree the performance impact concern is valid. So far, we have only applied pre-loading to 3 out of 5 regions, and they are all in the shipping step. Developers will be informed that setting a TTL cache header would allow this extension to appear like a native component. In the future, we might introduce an API flag to determine if an extension wants pre-loading.

cc @bc-charlesho

davidchin: > After positioning the region with CSS absolute positioning, the major challenge was creating a placeholder that occupies the same space as the extension in the actual area

I understand if the parent of an extension region needs to be unmounted/re-mounted, we won't be able to simply hide and show the extension. We will need to use a workaround like what you tried before. Another idea that I think we can try is to lift up the extension regions so they are closer to the root node (i.e.: `checkout`). That way we can avoid unmounting/re-mounting the regions.

bc-peng:
Closing the PR as we have decided to explore alternative solutions that do not require any actions from merchants or partners and do not introduce any side effects, such as double loading.

PR #1930, opened 2024-07-18
feat(payment): STRIPE-383 japanese states added

bc-dronov:
## What?
Added Japanese states to the Stripe mapper

## Why?
Because we cannot proceed to the payment step without it.

## Testing / Proof
Before:

https://github.com/user-attachments/assets/3aef57ea-9eb1-4f82-88f7-3546dc15ba14

After:

https://github.com/user-attachments/assets/a5321835-be9f-48d1-bb90-0ee8df3d9f63

@bigcommerce/team-checkout


PavlenkoM (approved):

serhii-tkachenko (approved):

animesh1987 (commented):
Should this list be moved to some stripe specific package?

bc-dronov:
> Should this list be moved to some stripe specific package?

Hi @animesh1987 ! We don't have a separate Stripe package for now. It could be refactored as a separate task with replacing all Stripe staff to the separate one.

bc-peng (approved):
I share Animesh's viewpoint that this issue can bypass a review from the checkout team by being moved to a stripe package.

animesh1987:
Thanks @bc-dronov will be ideal to plan and move the code across as follow up. This is good for now 😄

animesh1987 (approved):

PR #1949, opened 2024-08-05
chore(other): CHECKOUT-8509 Remove experiment wrapping code for CHECKOUT-4183.checkout_google_address_autocomplete_uk

animesh1987:
## What?
Remove experiment wrapping for code to add autocomplete for UK addresses.

## Why?
We have rolled out google maps support for UK addresses for one year now, so seems like a good opportunity for removing the experiment code wrapping the change.

## Testing / Proof
- CI 

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/billing/Billing.tsx
+++ b/packages/core/src/app/billing/Billing.tsx
@@ -143,13 +143,9 @@ function mapToBillingProps({
         return null;
     }
 
-    const { enableOrderComments, googleMapsApiKey, features } = config.checkoutSettings;
+    const { enableOrderComments, googleMapsApiKey } = config.checkoutSettings;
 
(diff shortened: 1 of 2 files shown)


.nvmrc
```diff
@@ -1 +1 @@
-18.17
+20.6.0
```
TomA-R, line 1: Seems like it should be a separate PR?

TomA-R (approved):

.nvmrc
animesh1987: Yes it can be, I am just taking the change from this https://github.com/bigcommerce/checkout-js/pull/1933 and applying it. Seems an easy and straightforward one.

TomA-R: For anyone watching, we decided to move this to a different PR

PR #1950, opened 2024-08-05
chore(other): GITHUB-000 Update nvmrc

animesh1987:
## What?
Matches the node version in config.yml. Reference https://github.com/bigcommerce/checkout-js/pull/1933

## Why?
Specify a specific version to be used. Plus matches with what we use in our CI orb.

## Testing / Proof
- CI

@bigcommerce/team-checkout


TomA-R (approved):
lgtm once commit validation is happy

PR #1964, opened 2024-08-14
refactor(payment): STRIPE-423 StripeV3 moved to separate package

bc-dronov:
## What?
StripeV3 moved to separate package

## Testing / Proof
Manual tests/ unit tests
@bigcommerce/team-checkout


packages/stripe-integration/.eslintrc.json
```diff
@@ -6,7 +6,8 @@
       "rules": {
         "@typescript-eslint/no-unsafe-argument": "off",
         "@typescript-eslint/no-unsafe-assignment": "off",
-        "@typescript-eslint/no-unsafe-member-access": "off"
+        "@typescript-eslint/no-unsafe-member-access": "off",
+        "@typescript-eslint/consistent-type-assertions": "off"
```
PavlenkoM, line 10: if we need this rule only for test, will be better to keep this rule only for test files
```
{
      "files": ["*.test.tsx"],
      "rules": {
          "@typescript-eslint/consistent-type-assertions": "off"
        }
}
```

packages/stripe-integration/src/stripev3/StripePaymentMethod.spec.tsx
```diff
@@ -1,59 +1,37 @@
+/* eslint @typescript-eslint/no-floating-promises: 0 */
```
PavlenkoM, line 1: we should migrate old `enzyme` test to `@testing-library/react`. But you can do this in separate PR. Create please a task for this, and for Stripe UPE as well.

PavlenkoM (approved):

packages/stripe-integration/src/stripev3/StripePaymentMethod.spec.tsx
vitalii-koshovyi: If you don't have problems with deadlines, I would suggest to implement it in this PR.

bc-dronov: No problem, I can do it in this PR

bc-dronov: @vitalii-koshovyi If you don't mind, I'll create a separate task for this.

bc-peng (approved):
LGTM.

PR #1978, opened 2024-08-27
fix(customer): CHECKOUT-8559 Remove string override for customer password validation error

animesh1987:
## What?
Remove string override for regex validation

## Why?
We return the correct translated message as part of password requirements from checkout settings API, and the translation in checkout app can be incorrect so it's better to use API message. Merchants can still override theme files.

## Testing / Proof
- CI

Screenshot 2024-08-27 at 3 39 19 PM

@bigcommerce/team-checkout


bc-peng (approved):
It took me a while to figure out where does `password must match the following: "/[a-zA-Z]/"` come from 😂 👏

PR #1979, opened 2024-08-28
CHECKOUT-8300: Reset Formik form once initial value is loaded in the shipping step

davidchin:
## What?
Add the `withFormikExtended` HOC, which extends the behaviour of the default `withFormik` HOC. This enhancement allows a Formik form to be reset once its initial value is loaded, enabling it to render even before the initial value is ready. Without this functionality, the form might not be able to submit if its `isInitialValid` check requires the initial value to be present.

## Why?
This is necessary because some checkout extensions are located inside Formik forms. To render these extensions before their parent forms are fully loaded, we need a way to reinitialise the forms once the required data is ready, as explained above.

### Before
This video shows that the shopper cannot submit the form because it is rendered before the initial value is loaded. Consequently, the `isInitialValid` check marks the form as invalid and prevents submission.

https://github.com/user-attachments/assets/09187d1e-5de2-417d-b83d-af470dbfc980

### After
This video shows that the shopper can submit the form because it is reinitialised after the initial value is loaded. As a result, the `isInitialValid` function has the necessary data to perform its check, allowing the form to be submitted if it is valid.

https://github.com/user-attachments/assets/7558bc2f-64d8-45c7-b195-3c77f357657e

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/common/form/index.ts
+++ b/packages/core/src/app/common/form/index.ts
@@ -1,2 +1,3 @@
 export { default as connectFormik } from './connectFormik';
 export { default as ConnectFormikProps, WithFormikProps } from './ConnectFormikProps';
+export { default as withFormikExtended, WithFormikExtendedProps } from './withFormikExtended';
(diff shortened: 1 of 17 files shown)


bc-peng (approved):
LGTM👏👏

PR #1985, opened 2024-09-01
feat(payment): PI-1572 Write integration tests to verify all `klarna`…

bc-sebastianszafraniec:
… and `klarnav2` components of Checkout SDK/JS are working together

## What?
Added E2E tests for Klarna and KlarnaV2 integration on Checkout JS

## Testing / Proof
All tests have passed

@bigcommerce/team-checkout

### DIFF
--- a/packages/test-framework/src/fixture/pageObject/Assertions.ts
+++ b/packages/test-framework/src/fixture/pageObject/Assertions.ts
@@ -22,10 +22,6 @@ export class Assertions {
         await expect(page.locator(element)).toBeVisible();
     }
 
-    shouldMadeRequest(request: boolean) {
-        expect(request).toBeTruthy();
-    }
-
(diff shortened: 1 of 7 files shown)


packages/test-framework/src/fixture/pageObject/Assertions.ts
```diff
@@ -16,6 +16,16 @@ export class Assertions {
...
 
+    async shouldSeeElement(element: string): Promise<void> {
+        const page = this.page;
+
+        await expect(page.locator(element)).toBeVisible();
+    }
+
+    shouldMadeRequest(request: boolean) {
+        expect(request).toBeTruthy();
+    }
```
bc-peng, line 27: This helper function seems not related to reading requests.
Maybe for the usecase of this PR, we can assert the value directly.

bc-peng (approved):

PR #2018, opened 2024-09-20
feat(payment): STRIPE-444 New Stripe UPE (PoC)

PavlenkoM:
## What?
PoC how we can implement custom view for payment method inside our default payment methods accordion.

In our case it is for new Stripe implementation which uses custom accordion view and group multiple payment methods in one payment element.
This custom accordion should be rendered inside our default payment methods list. Also it should be sorted by some additional logic and can be placed on any position in default accordion.
Current PoC does not contain all Stripe Element styling, because we are waiting for some fixes on stripe side.

Also Accordion ui and context was migrated to the shared UI package to be available for payment integration packages.

## Why?
...

## Testing / Proof

https://github.com/user-attachments/assets/e79bfd32-e011-4649-9c7e-c78763b57fad

@bigcommerce/team-checkout


animesh1987:
Hey @PavlenkoM is this still valid?

PR #2020, opened 2024-09-23
CHECKOUT-8519: Load scripts with integrity hashes to meet PCI4 requirements

davidchin:
## What?
Include integrity hashes in `manifest.json` to be added to the scripts and stylesheets used during checkout. Any script or stylesheet that doesn't match its corresponding subresource integrity hash ([SRI](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity)) will not be executed.

## Why?
To comply with the script integrity requirement ([6.4.3](https://www.pcisecuritystandards.org/wp-content/uploads/2023/09/05.Scaling-6.4.3-and-11.6.1-Browser-Script-Management-and-The-Large-Enterprise-Journey-to-Compliance.pdf)) of PCI4, which ensures that scripts executed during checkout have not been tampered with by a malicious attacker.

## Testing / Proof
Screenshot 2024-09-23 at 4 06 31 PM

@bigcommerce/team-checkout

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -6,12 +6,16 @@ const { join } = require('path');
 const StyleLintPlugin = require('stylelint-webpack-plugin');
 const { DefinePlugin } = require('webpack');
 const WebpackAssetsManifest = require('webpack-assets-manifest');
+const { isArray, mergeWith } = require('lodash');
 
-const { AsyncHookPlugin,
+const {
+    AsyncHookPlugin,
     BuildHookPlugin,
     getLoaderPackages: { aliasMap: alias, tsLoaderIncludes },
     getNextVersion,
-    transformManifest } = require('./scripts/webpack');
+    mergeManifests,
+    transformManifest,
+} = require('./scripts/webpack');
 
 const ENTRY_NAME = 'checkout';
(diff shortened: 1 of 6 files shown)


webpack.config.js
```diff
@@ -32,6 +36,7 @@ const BABEL_PRESET_ENV_CONFIG = {
     useBuiltIns: 'usage',
     modules: false,
 };
+const PRELOAD_ASSETS = ['billing', 'shipping', 'payment'];
```
animesh1987, line 39: What does this do? Does it mean we now load assets before we actually get on the step?

animesh1987 (approved):
Looks good to me code wise 👍

richapatel1510 (approved):

webpack.config.js
davidchin: @animesh1987 We used to preload all assets listed in the `dynamicChunks` section of `manifest.json`. However, as `manifest.json` now includes the dynamic chunks for `loader.js`, which contains translation JSON files in various languages (e.g., `translations-es-1.525.0.js`), we need to limit preloading to only the necessary ones (`billing`, `shipping` and `payment`).

bc-peng (approved):

PR #2023, opened 2024-09-24
fix(checkout): CHECKOUT-8006 Optimise Shipping Option Updates

bc-mackxu:
## What?  
Moved the shipping address form value change detection into the debounce function.

## Why?  
This ensures that the API call to fetch available shipping options is only triggered when the form values (like the postcode) have been updated and the user has stopped typing.

## Rollout/Rollback
Turn off experiment `CHECKOUT-8006.move_check_updated_address_value_within_debounced`

## Testing / Proof
- [x] unit test
- [x] manual test
### Setup  
- PostCode: `2000` - Free Shipping  
- PostCode: `20001` - Flat rate $6.65  

### Before  
When changing the postcode and switching back quickly, different shipping options were displayed due to immediate API calls.  

https://github.com/user-attachments/assets/54b3ec36-88dd-46ee-ba16-8ff692cb0e5c  

### After  
Now, the shipping options only update after typing has stopped, ensuring accuracy in the options displayed. Additionally, while the shipping options are refreshing, the shipping address form shows a loading icon to indicate that an update is in progress.  

https://github.com/user-attachments/assets/7674c645-ef6b-4844-bda0-789f77bc1012  

---

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/shipping/Shipping.tsx
+++ b/packages/core/src/app/shipping/Shipping.tsx
@@ -85,6 +85,7 @@ export interface WithCheckoutShippingProps {
     updateCheckout(payload: CheckoutRequestBody): Promise<CheckoutSelectors>;
     updateShippingAddress(address: Partial<Address>): Promise<CheckoutSelectors>;
     shouldRenderStripeForm: boolean;
+    moveCheckUpdatedAddressValueWithinDebouncedExperiment: boolean;
 }
 
 interface ShippingState {
@@ -134,6 +135,7 @@ class Shipping extends Component<ShippingProps & WithCheckoutShippingProps, Ship
             isFloatingLabelEnabled,
             shouldRenderStripeForm,
             shouldRenderWhileLoading,
+            moveCheckUpdatedAddressValueWithinDebouncedExperiment,
             ...shippingFormProps
         } = this.props;
 
@@ -180,6 +182,7 @@ class Shipping extends Component<ShippingProps & WithCheckoutShippingProps, Ship
                         isGuest={isGuest}
                         isInitialValueLoaded={shouldRenderWhileLoading ? !isInitializing : true}
                         isMultiShippingMode={isMultiShippingMode}
+                        moveCheckUpdatedAddressValueWithinDebouncedExperiment={moveCheckUpdatedAddressValueWithinDebouncedExperiment}
                         onMultiShippingSubmit={this.handleMultiShippingSubmit}
                         onSingleShippingSubmit={this.handleSingleShippingSubmit}
                         onUseNewAddress={this.handleUseNewAddress}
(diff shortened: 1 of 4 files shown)


packages/core/src/app/shipping/SingleShippingForm.tsx
```diff
@@ -111,10 +112,50 @@ class SingleShippingForm extends PureComponent<
...
+                                    'consignments.availableShippingOptions': includeShippingOptions,
+                                },
+                            },
+                        });
+    
+                        if (includeShippingOptions) {
+                            this.setState({ hasRequestedShippingOptions: true });
+                        }
+                    } finally {
+                        this.setState({ isUpdatingShippingData: false });
```
animesh1987, line 153: Do we need this https://github.com/bigcommerce/checkout-js/pull/2023/files#diff-738bdd4ad4c50c9374ff90b1691ec572ef3d86fe8457f2d942b1c7af745df69aR175?

packages/core/src/app/shipping/SingleShippingForm.tsx
```diff
@@ -111,10 +112,50 @@ class SingleShippingForm extends PureComponent<
     ) {
         super(props);
 
-        const { updateAddress } = this.props;
-
         this.debouncedUpdateAddress = debounce(
             async (address: Address, includeShippingOptions: boolean) => {
+                const {
```
animesh1987, line 117: My suggestion would be here to create another debounce function based on the experiment and use the experiment as a switch in `updateAddressWithFormData` function.

packages/core/src/app/shipping/Shipping.tsx
```diff
@@ -85,6 +85,7 @@ export interface WithCheckoutShippingProps {
     updateCheckout(payload: CheckoutRequestBody): Promise<CheckoutSelectors>;
     updateShippingAddress(address: Partial<Address>): Promise<CheckoutSelectors>;
     shouldRenderStripeForm: boolean;
+    moveCheckUpdatedAddressValueWithinDebouncedExperiment: boolean;
```
bc-peng, line 88: Thought it might be clearer to use a more action-oriented name instead of just an experiment name. For example, we could use `isDebouncedCheckAndUpdateAddress`.

packages/core/src/app/shipping/Shipping.tsx
```diff
@@ -434,6 +437,7 @@ export function mapToShippingProps({
         updateShippingAddress: checkoutService.updateShippingAddress,
         isFloatingLabelEnabled: isFloatingLabelEnabled(config.checkoutSettings),
         shouldRenderStripeForm: providerWithCustomCheckout === PaymentMethodId.StripeUPE && shouldUseStripeLinkByMinimumAmount(cart),
+        isDebouncedCheckAndUpdateAddressExperiment: features['CHECKOUT-8006.move_check_updated_address_value_within_debounced'] ?? false,
```
bc-peng, line 440: For better supporting expeirment removal, we now use this method to get the value.
https://github.com/bigcommerce/checkout-js/blob/master/packages/core/src/app/common/utility/isExperimentEnabled.ts

bc-peng (approved):

animesh1987 (approved):

PR #2032, opened 2024-09-30
feat(payment): PI-1595 Refactor the existing `AmazonPay` and `AmazonP…

bc-sebastianszafraniec:
…ayV2` components in Checkout JS to use the new checkout payment integration JS API

## What?
Move AmazonPayV2PaymentMethod and AmazonPayV2Button to a separate package

## Why?
To use the new checkout payment integration API

## Testing / Proof
Tested manually

https://github.com/user-attachments/assets/701cc46c-deae-42b2-aaed-7e724f95ef86

https://github.com/user-attachments/assets/7d84106d-91c0-41de-a796-86c853d293bf

@bigcommerce/team-checkout


packages/amazon-pay-v2-integration/src/AmazonPayV2PaymentMethod.tsx
```diff
@@ -0,0 +1,98 @@
...
+        !isGuestCustomer && Boolean(method.config.isVaultingEnabled);
+
+    const instruments = checkoutState.data.getInstruments(method) || [];
+
+    return (
+        <HostedWidgetPaymentComponent
+            {...rest}
+            buttonId="editButtonId"
+            containerId="paymentWidget"
+            deinitializeCustomer={undefined}
```
bc-nick, line 65: Why are u set the `undefined` value like this?

packages/amazon-pay-v2-integration/src/AmazonPayV2PaymentMethod.tsx
```diff
@@ -0,0 +1,98 @@
...
+            deinitializePayment={checkoutService.deinitializePayment}
+            disableSubmit={disableSubmit}
+            hidePaymentSubmitButton={hidePaymentSubmitButton}
+            hideWidget
+            initializeCustomer={undefined}
```
bc-nick, line 70: the same question

bc-peng:
Is it possbile to add e2e tests to cover the new package now?
I understand in order to do so, we need to mock amazon packages, just asking.

bc-sebastianszafraniec:
> Is it possbile to add e2e tests to cover the new package now? I understand in order to do so, we need to mock amazon packages, just asking.

Hi @bc-peng Technically it is possible. However we used to add e2e tests in separate tasks and in separate PR. I believe it will be next step in working on this package.

packages/amazon-pay-v2-integration/src/AmazonPayV2PaymentMethod.tsx
bc-sebastianszafraniec: This solution was copied with AmazonPayV2PaymentMethod from core, but after analyzing, looks like these props are not needed here to be pass as undefined. I removed them.

bc-peng (approved):

packages/amazon-pay-v2-integration/README.md
```diff
@@ -0,0 +1,15 @@
+# amazon-pay-v2-integration
+
+This library was generated with [Nx](https://nx.dev).
```
animesh1987, line 3: Can we please have a better readme as compared to the boilerplate.

PavlenkoM (approved):

animesh1987 (approved):

PR #2035, opened 2024-10-02
CHECKOUT-8592: Add address selector for new multishipping UI

richapatel1510:
## What?

- Create new components for new multishipping user interface.
- Add address selector and `Add Shipping Destination` button

## Why?
For new improved Multishipping UI/UX

## Testing / Proof

- CI checks
- Manual testing

https://github.com/user-attachments/assets/564b700e-2f20-4751-b1d2-b1df1abd6b2a

@bigcommerce/team-checkout


package.json
```diff
@@ -86,6 +86,7 @@
     "react-transition-group": "^2.5.3",
     "reselect": "^4.1.7",
     "shallowequal": "^1.1.0",
+    "styled-components": "^6.1.13",
```
animesh1987, line 89: We avoid using styled-component in checkout-js, since preference is to use sass files which merchants can override for customising. Maybe we can avoid these and use classes and sass?

package.json
```diff
@@ -14,7 +14,7 @@
     "build": "nx run core:build",
     "build:server": "http-server dist",
     "dev": "nx run core:dev",
-    "dev:server": "http-server build",
+    "dev:server": "http-server build --cors",
```
animesh1987, line 17: What does this do?

richapatel1510: Possibly after [this](https://github.com/bigcommerce/checkout-js/pull/2020) update, when we try and run `dev:server` it turns runs with `CORS: false` which does not allow dev store to make requests to localhost, so adding this flag makes that http server setting `CORS: true` true which allows dev store to run fine

packages/core/src/app/shipping/MultiShippingFormV2.scss
```diff
@@ -0,0 +1,17 @@
...
+}
+
+.consignment-header {
+    font-size: fontSize("large");
+    margin: 0 0 spacing("single");
+}
+
+.add-consignment-button {
+    color: color("greys", "dark");
+}
```
bc-peng, line 17: EOL

packages/core/src/app/shipping/MultiShippingFormV2Footer.tsx
```diff
@@ -0,0 +1,41 @@
...
+}
+
+const MultiShippingFormV2Footer: FunctionComponent<ShippingFormFooterProps> = ({
+    shouldShowOrderComments,
+    shouldDisableSubmit,
+    isLoading,
+}) => {
+    return (
+        <>
+            <Extension region={ExtensionRegion.ShippingShippingAddressFormAfter} />
```
bc-peng, line 23: Let's keep it for single shipping for now.

packages/core/src/app/address/AddressSelect.test.tsx
```diff
@@ -59,6 +60,15 @@ describe('AddressSelect component', () => {
         expect(screen.getByText(selectedAddress.lastName)).toBeInTheDocument();
     });
 
+    it('renders static address V2 when there is a selected address and showSingleLineAddress is true', () => {
```
bc-peng, line 63:
```suggestion
    it('renders single line static address when there is a selected address and showSingleLineAddress is true', () => {
```

packages/core/src/app/shipping/MultiShippingFormV2.tsx
```diff
@@ -0,0 +1,109 @@
...
+    onSubmit(values: MultiShippingFormV2Values): void;
+}
+
+const MultiShippingFormV2: FunctionComponent<MultiShippingFormV2Props> = ({
+    countriesWithAutocomplete,
+    defaultCountryCode,
+    isLoading,
+    onUnhandledError,
+}: MultiShippingFormV2Props) => {
+    const [addShippingDestination, setAddShippingDestination] = useState(false);
```
bc-peng, line 36: Does it make sense to change the state name to `isAddingShippingDestination`? I feel this is more in-line with [our existing naming usages](https://github.com/search?q=repo%3Abigcommerce%2Fcheckout-js+isadding&type=code).

bc-peng (approved):
LGTM👏👏

animesh1987 (approved):

PR #2037, opened 2024-10-03
CHECKOUT-8519: Upgrade yup to a version that does not trigger CSP error

davidchin:
## What?
Upgrade `yup` to a version that doesn't trigger CSP error.

## Why?
`yup` depends on an old version of `property-expr`, which evaluates arbitrary strings as JavaScript. The newer version of `yup` uses a version of `property-expr` that doesn't use `eval` anymore (see https://github.com/jquense/expr/pull/11).

I decided not to upgrade to the latest version of `yup` for now because it has several breaking changes (see https://github.com/jquense/yup/blob/master/CHANGELOG.md)

### Before
This screenshot shows that `property-expr` package triggers CSP error when there is`script-src` policy.

Screenshot 2024-10-03 at 4 02 08 PM

### After
This screenshot shows that `property-expr` package no longer triggers CSP error when there is `script-src` policy.

Screenshot 2024-10-03 at 3 52 38 PM

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/formFields/getCustomFormFieldsValidationSchema.ts
+++ b/packages/core/src/app/formFields/getCustomFormFieldsValidationSchema.ts
@@ -53,7 +53,6 @@ export default memoize(function getCustomFormFieldsValidationSchema({
                         schema[name] = date()
                             // Transform NaN values to undefined to avoid empty string (empty input) to fail date
                             // validation when it's optional
-                            .strict(true)
                             .nullable(true)
                             .transform((value, originalValue) =>
                                 originalValue === '' ? null : value,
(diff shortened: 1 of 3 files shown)


packages/core/src/app/formFields/getCustomFormFieldsValidationSchema.ts
```diff
@@ -62,7 +61,6 @@ export default memoize(function getCustomFormFieldsValidationSchema({
                         schema[name] = number()
                             // Transform NaN values to undefined to avoid empty string (empty input) to fail number
                             // validation when it's optional
-                            .strict(true)
```
davidchin, line 65: The `strict` option should never be used with `transform`, as it is designed to bypass the `transform` function. The reason why it worked before the version upgrade was due to a bug in the library. The bug has been fixed since version [0.29.1](https://github.com/jquense/yup/blob/master/CHANGELOG.md#bug-fixes-31).

bc-peng (approved):

animesh1987 (approved):

PR #2051, opened 2024-10-16
Add create certificate on checkout page 

freddyjgs:
## What?
...

## Why?
...

## Testing / Proof
...

@bigcommerce/team-checkout


.env-example
```diff
@@ -1,3 +1,8 @@
 PORT=8080
 MODE=replay
-STOREURL=
+ACCESS TOKEN: dgu5fvo5xjt64ocykla3a6gkh5ahkdy
```
bc-donfran, line 3: You may want to rotate this as secrets shouldn't be public

localhost.key
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
bc-donfran, line 1: You may want to rotate this (if used in your environment) as this would be considered a secret/sensitive.

PR #2055, opened 2024-10-21
feat(checkout): CHECKOUT-8599 Add Confirmation Modal for Multi-shipping

bc-peng:
## What?
Show the warning popup before redirecting to cart page when multi-address shipping is enabled along with the new UI.

*The modal is expected to be reused when switching from multi-shipping to single shipping.
## Why?
Improving multi-shipping UX.

### Mobile

https://github.com/user-attachments/assets/eaeb6c43-92e7-40de-9112-5c5fc6cfb920

### Desktop

https://github.com/user-attachments/assets/90aed54a-de69-4e68-b727-8448624c5457

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/cart/CartSummary.tsx
+++ b/packages/core/src/app/cart/CartSummary.tsx
@@ -16,15 +16,26 @@ export type WithCheckoutCartSummaryProps = {
     storeCurrency: StoreCurrency;
     shopperCurrency: ShopperCurrency;
     storeCreditAmount?: number;
+    isNewMultiShippingUIEnabled: boolean;
 } & RedeemableProps;
 
-const CartSummary: FunctionComponent<WithCheckoutCartSummaryProps> = ({ cartUrl, ...props }) => {
-    const headerLink = isBuyNowCart() ? null : <EditLink url={cartUrl} />;
+const CartSummary: FunctionComponent<
+    WithCheckoutCartSummaryProps & {
+        isMultiShippingMode: boolean;
+    }
+> = ({ cartUrl, isMultiShippingMode, isNewMultiShippingUIEnabled, ...props }) => {
+    const headerLink = isBuyNowCart() ? null : (
+        <EditLink
+            isMultiShippingMode={isNewMultiShippingUIEnabled && isMultiShippingMode}
+            url={cartUrl}
+        />
+    );
 
     return withRedeemable(OrderSummary)({
         ...props,
         cartUrl,
         headerLink,
+        isNewMultiShippingUIEnabled,
     });
 };
 
--- a/packages/core/src/app/cart/EditLink.tsx
+++ b/packages/core/src/app/cart/EditLink.tsx
@@ -14,9 +14,7 @@ const EditLink: FunctionComponent<EditLinkProps> = ({ className, url, isMultiShi
     const [isModalOpen, setIsModalOpen] = useState(false);
 
     const gotoCartPage = () => {
-        if (window.top) {
-            window.top.location.href = url;
-        }
+        window.location.assign(url);
     };
 
     if (isMultiShippingMode) {
--- a/packages/core/src/app/cart/EditLink.tsx
+++ b/packages/core/src/app/cart/EditLink.tsx
@@ -1,22 +1,58 @@
-import React, { FunctionComponent, memo } from 'react';
+import React, { FunctionComponent, memo, useState } from 'react';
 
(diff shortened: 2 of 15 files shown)


packages/core/src/app/checkout/Checkout.tsx
```diff
@@ -523,22 +523,24 @@ class Checkout extends Component<
     }
 
     private renderCartSummary(): ReactNode {
+        const { isMultiShippingMode } = this.state;
```
bc-peng, line 526: In addtion to the manual switch, we also read consignments to determine if the checkout is using multi-shipping during each page initialization.
https://github.com/bigcommerce/checkout-js/blob/e49e97ca3cb6b0a5ee90f782ae2a0bee3e52dbcc/packages/core/src/app/checkout/Checkout.tsx#L259

packages/core/src/app/cart/CartSummary.tsx
```diff
@@ -16,15 +16,26 @@ export type WithCheckoutCartSummaryProps = {
...
-const CartSummary: FunctionComponent<WithCheckoutCartSummaryProps> = ({ cartUrl, ...props }) => {
-    const headerLink = isBuyNowCart() ? null : <EditLink url={cartUrl} />;
+const CartSummary: FunctionComponent<
+    WithCheckoutCartSummaryProps & {
+        isMultiShippingMode: boolean;
+    }
+> = ({ cartUrl, isMultiShippingMode, isNewMultiShippingUIEnabled, ...props }) => {
+    const headerLink = isBuyNowCart() ? null : (
+        <EditLink
+            isMultiShippingMode={isNewMultiShippingUIEnabled && isMultiShippingMode}
```
animesh1987, line 29: I see this will be rolled out for all users regardless of the experiment, is that expected?

packages/core/src/app/cart/EditLink.tsx
```diff
@@ -1,22 +1,58 @@
...
-    >
-        <TranslatedString id="cart.edit_cart_action" />
-    </a>
-);
+const EditLink: FunctionComponent<EditLinkProps> = ({ className, url, isMultiShippingMode }) => {
+    const [isModalOpen, setIsModalOpen] = useState(false);
+
+    const gotoCartPage = () => {
+        if (window.top) {
+            window.top.location.href = url;
```
animesh1987, line 18: Howcome we don't use `window.location.assign()`?

packages/core/src/app/cart/CartSummary.tsx
animesh1987: Saw this https://github.com/bigcommerce/checkout-js/pull/2055/files#diff-ea06ebdcd56f95f158f566d4d907e9e54c508436fac6c5e65900160b0aedd7f7R28.

packages/ui/src/modal/ConfirmationModal.tsx
```diff
@@ -0,0 +1,56 @@
...
+    onRequestClose,
+}: ConfirmationModalProps) => {
+    return (
+        <Modal
+            additionalModalClassName="modal--confirm"
+            footer={
+                <Button
+                    onClick={action}
+                    size={ButtonSize.Small}
+                    testId="edit-multi-shipping-cart-confirm-button"
```
richapatel1510, line 32: Should this be dynamic from props as well?

packages/core/src/app/cart/EditLink.tsx
bc-peng: I noticed we used `window.top.location` three times in `checkout.tsx`. For example, jumping to the customer registration page.
https://github.com/bigcommerce/checkout-js/blob/366c1fcd6f2bda7e0529f29013468e0d1571c676/packages/core/src/app/checkout/Checkout.tsx#L753

It's good for the embedded checkout, and it behaves similarly to the `assign` method in terms of creating browsing history.

Do you think we need to use `assign` here?

animesh1987: I think possibly best to test with `assign`.

packages/ui/src/modal/ConfirmationModal.tsx
bc-peng: Good catch, I have removed this `testId` as we are able to assert the modal existence by other methods.

animesh1987 (approved):
Nice work 👏

richapatel1510 (approved):

PR #2056, opened 2024-10-21
CHECKOUT-8596 Add product allocation modal

richapatel1510:
## What?

- Add allocate items modal for new SF multishipping UI

## Why?

- So that products can be assigned to an address in the new multi shipping experience.

## Testing / Proof

- CI checks
- Manual testing

https://github.com/user-attachments/assets/149bcfc2-58ba-4941-b3a6-5147dbbd3244

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/shipping/AllocateItemsModal.tsx
+++ b/packages/core/src/app/shipping/AllocateItemsModal.tsx
@@ -1,7 +1,6 @@
 import { Address, ConsignmentLineItem } from "@bigcommerce/checkout-sdk";
 import { FormikProps } from "formik";
 import React, { FunctionComponent } from "react";
-import { number, object } from "yup";
 
 import { preventDefault } from "@bigcommerce/checkout/dom-utils";
 import { withLanguage, WithLanguageProps } from "@bigcommerce/checkout/locale";
@@ -13,8 +12,8 @@ import { Button } from "../ui/button";
 import { Form } from "../ui/form";
 import { Modal, ModalHeader } from "../ui/modal";
 
-import { MultiShippingTableItemWithType, UnassignedItems } from "./hooks/useMultishippingConsignmentItems";
 import LeftToAllocateItemsTable from "./LeftToAllocateItemsTable";
+import { UnassignedItems } from "./MultishippingV2Type";
 
 export interface AllocateItemsModalFormValues {
     [key: string]: number;
@@ -134,19 +133,5 @@ export default withLanguage(
             return values;
         },
         enableReinitialize: true,
-        validationSchema: ({ unassignedItems }: AllocateItemsModalProps) => {
-            // TODO: This will change when we get latest designs for error handling
-            const createItemSchema = (item: MultiShippingTableItemWithType) => {
-                return number()
-                    .integer()
-                    .min(0)
-                    .max(item.quantity, `You can't allocate more than ${item.quantity} items`)
-            };
-            const schemaObject = Object.fromEntries(
-                unassignedItems.lineItems.map((item) => [item.id.toString(), createItemSchema(item)]),
-            );
-
-            return object().shape(schemaObject);
-        }
     })(AllocateItemsModal),
 );
--- a/packages/core/src/app/shipping/AllocateItemsModal.tsx
+++ b/packages/core/src/app/shipping/AllocateItemsModal.tsx
@@ -13,7 +13,7 @@ import { Form } from "../ui/form";
 import { Modal, ModalHeader } from "../ui/modal";
 
 import LeftToAllocateItemsTable from "./LeftToAllocateItemsTable";
-import { UnassignedItems } from "./MultishippingV2Type";
+import { MultiShippingTableData } from "./MultishippingV2Type";
 
 export interface AllocateItemsModalFormValues {
     [key: string]: number;
@@ -24,7 +24,7 @@ interface AllocateItemsModalProps {
     isOpen: boolean;
     onRequestClose?(): void;
     address: Address;
-    unassignedItems: UnassignedItems;
+    unassignedItems: MultiShippingTableData;
     onAllocateItems(consignmentLineItems: ConsignmentLineItem[]): void;
 }
 
--- a/packages/core/src/app/shipping/AllocateItemsModal.tsx
+++ b/packages/core/src/app/shipping/AllocateItemsModal.tsx
@@ -1,0 +1,152 @@
+import { Address, ConsignmentLineItem } from "@bigcommerce/checkout-sdk";
+import { FormikProps } from "formik";
+import React, { FunctionComponent } from "react";
+import { number, object } from "yup";
+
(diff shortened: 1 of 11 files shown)


packages/core/src/app/shipping/hooks/useMultishippingConsignmentItems.ts
bc-peng, line 1: Perhaps we should break this file into smaller files so that each file serves a single purpose. 🤔

packages/core/src/app/shipping/MultiShippingFormV2.scss
```diff
@@ -8,10 +8,154 @@
...
+        border-left: none;
+        border-right: none;
+    
+        thead {
+            background: none;
+            tr {
+                border-top: container("border");
+                border-bottom: container("border");
+                th:first-child {
+                    width: 50%;
```
bc-peng, line 102: In order to manage all these numbers easily in the furture, can we create some variables and put them in one place? Maybe at top of this file?
https://github.com/bigcommerce/checkout-js/blob/c21be1be1b29d589e6473490112601ec469d8887/packages/core/src/scss/settings/global/layout/_layout.scss#L13

packages/core/src/app/shipping/NewConsignment.tsx
```diff
@@ -45,16 +54,63 @@ const NewConsignment = ({
...
+            }
+        } finally {
+            toggleAllocateItemsModal();
+        }
+    }
+
     return (
         <div className='consignment-container'>
-            <h3 className='consignment-header'>Shipping destination {consignmentNumber}</h3>
+            <h3 className='consignment-header'>Destination #{consignmentNumber}</h3>
```
bc-peng, line 84: Oooops, in order to support multi-lang, we need to put new strings into `en.json`, and use it in this way `<TranslatedString id="address.add_address_heading" />`. 
https://github.com/bigcommerce/checkout-js/blob/c21be1be1b29d589e6473490112601ec469d8887/packages/locale/src/translations/en.json#L2

This also applies to places like `No items allocated` and `Allocate items` below.

richapatel1510: I've created this ticket to localise all the strings for this project https://bigcommercecloud.atlassian.net/browse/CHECKOUT-8780

packages/core/src/app/shipping/hooks/useMultishippingConsignmentItems.ts
animesh1987: Good point. I think we need to try and break down components to small as possible kind of complying with single responsibility principle, for context https://dev.to/mikhaelesa/single-responsibility-principle-in-react-10oc

richapatel1510: This hook just calculates unassigned items and mapped items to consignment. I can move the types/interfaces to separate file.

packages/core/src/app/shipping/ConsignmentLineItem.tsx
```diff
@@ -0,0 +1,111 @@
+import { ConsignmentLineItem } from "@bigcommerce/checkout-sdk";
```
bc-peng, line 1: Not sure why our linter didn't pick this up. This file appears to have some eslint issues, like not using single quote.
I also observed that the linting rules were not applied consistently in `checkout-js`. 
Perhaps after this PR, we can investigate why the linter is behaving inconsistently and do a quick cleanup.
cc @animesh1987 

image

bc-peng: The rest looks good to me.

packages/core/src/app/shipping/hooks/useMultishippingConsignmentItems.ts
```diff
@@ -0,0 +1,110 @@
+import { Consignment, LineItemMap } from "@bigcommerce/checkout-sdk";
+
+import { useCheckout } from "@bigcommerce/checkout/payment-integration-api";
+
+import { LineItemType, MappedDataConsignment, MultiShippingTableItemWithType, UnassignedItems } from "../MultishippingV2Type";
+
+interface MultiShippingConsignmentItemsHook {
+    unassignedItems: UnassignedItems;
```
animesh1987, line 8: Can we have a different name for interface? Unassigned items seems very specific.

packages/core/src/app/shipping/MultishippingV2Type.ts
```diff
@@ -0,0 +1,32 @@
...
+    type: LineItemType;
+}
+
+export interface UnassignedItems {
+    lineItems: MultiShippingTableItemWithType[];
+    hasDigitalItems: boolean;
+    shippableItemsCount: number;
+};
+
+export interface MappedDataConsignment extends Consignment, UnassignedItems {
```
animesh1987, line 30: This could also be named better since by reading the name it's not clear what it does.

packages/core/src/app/shipping/AllocateItemsModal.tsx
```diff
@@ -0,0 +1,137 @@
...
+    address,
+    unassignedItems,
+    setValues,
+    dirty,
+    submitForm,
+}: AllocateItemsModalProps & FormikProps<AllocateItemsModalFormValues>) => {
+
+    const leftItemsTotal = unassignedItems.shippableItemsCount;
+
+    const allocationPendingMessage = `${leftItemsTotal} ${leftItemsTotal === 1 ? 'item left to allocate' : 'items left to allocate'
```
animesh1987, line 44: I guess this will be done as part of translation ticket?

packages/core/src/app/shipping/ConsignmentLineItem.tsx
animesh1987: Sounds good, maybe our config is incorrect.

packages/core/src/app/shipping/LeftToAllocateItemsTable.tsx
```diff
@@ -0,0 +1,47 @@
...
+    return <table className="table left-to-allocate-items-table">
+        <thead>
+            <tr>
+                <th>Item</th>
+                <th>Left to allocate</th>
+                <th>Quantity</th>
+            </tr>
+        </thead>
+        <tbody>
+            {items.map(item => (
```
animesh1987, line 21: This should be a separate component.

packages/core/src/app/shipping/NewConsignment.tsx
animesh1987: This is ok for now, but ideally we should localise the strings as we add them.

bc-peng: Thought we could do the locale ticket right after this PR in order to keep the gap small.

animesh1987 (approved):
Nice work 👏

bc-peng (approved):

PR #2074, opened 2024-10-30
feat(checkout): CHECKOUT-8601 Enhanced Multi-Address Shipping Summary View

bc-mackxu:
## What?
Introduce a new Multi-address Shipping summary view.

## Why?
Improve shopper experience.

### Before
image

### After
image

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/checkout/Checkout.tsx
+++ b/packages/core/src/app/checkout/Checkout.tsx
@@ -40,7 +40,7 @@ import { getSupportedMethodIds } from '../customer/getSupportedMethods';
 import { SubscribeSessionStorage } from '../customer/SubscribeSessionStorage';
 import { EmbeddedCheckoutStylesheet, isEmbedded } from '../embeddedCheckout';
 import { PromotionBannerList } from '../promotion';
-import { hasSelectedShippingOptions, isUsingMultiShipping, StaticConsignment } from '../shipping';
+import { hasSelectedShippingOptions, isUsingMultiShipping, ShippingSummary } from '../shipping';
 import { ShippingOptionExpiredError } from '../shipping/shippingOption';
 import { LazyContainer, LoadingNotification, LoadingOverlay } from '../ui/loading';
 import { MobileView } from '../ui/responsive';
@@ -138,6 +138,7 @@ export interface WithCheckoutProps {
     isPending: boolean;
     isPriceHiddenFromGuests: boolean;
     isShowingWalletButtonsOnTop: boolean;
+    isNewMultiShippingUIEnabled: boolean;
     loginUrl: string;
     cartUrl: string;
     createAccountUrl: string;
@@ -423,7 +424,7 @@ class Checkout extends Component<
     }
 
     private renderShippingStep(step: CheckoutStepStatus): ReactNode {
-        const { hasCartChanged, cart, consignments = [] } = this.props;
+        const { hasCartChanged, cart, consignments = [], isNewMultiShippingUIEnabled } = this.props;
 
         const { isBillingSameAsShipping, isMultiShippingMode } = this.state;
 
@@ -438,15 +439,7 @@ class Checkout extends Component<
                 key={step.type}
                 onEdit={this.handleEditStep}
                 onExpanded={this.handleExpanded}
-                summary={consignments.map((consignment) => (
-                    <div className="staticConsignmentContainer" key={consignment.id}>
-                        <StaticConsignment
-                            cart={cart}
-                            compactView={consignments.length < 2}
-                            consignment={consignment}
-                        />
-                    </div>
-                ))}
+                summary={<ShippingSummary cart={cart} consignments={consignments} isMultiShippingMode={isMultiShippingMode} isNewMultiShippingUIEnabled={isNewMultiShippingUIEnabled} />}
             >
                 <LazyContainer loadingSkeleton={<AddressFormSkeleton />}>
                     <Shipping
--- a/packages/core/src/app/shipping/ShippingSummary.tsx
+++ b/packages/core/src/app/shipping/ShippingSummary.tsx
@@ -1,0 +1,53 @@
+
+import { Cart, Consignment } from '@bigcommerce/checkout-sdk';
+import React, { FunctionComponent, memo } from 'react';
+
+import StaticConsignment from './StaticConsignment';
+import StaticMultiConsignment from './StaticMultiConsignment';
+
+
+interface ShippingSummaryProps {
+    isNewMultiShippingUIEnabled: boolean;
+    isMultiShippingMode: boolean;
+    consignments: Consignment[];
+    cart: Cart;
+}
+
+const ShippingSummary: FunctionComponent<ShippingSummaryProps> = ({
+    isNewMultiShippingUIEnabled,
+    isMultiShippingMode,
+    consignments,
+    cart
+}) => {
+    if (isNewMultiShippingUIEnabled && isMultiShippingMode) {
+        return (
+            <>
+                {consignments.map((consignment, index) => (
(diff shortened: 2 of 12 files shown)


packages/core/src/app/checkout/Checkout.tsx
```diff
@@ -423,30 +424,44 @@ class Checkout extends Component<
...
-        const { hasCartChanged, cart, consignments = [] } = this.props;
+        const { hasCartChanged, cart, consignments = [], isNewMultiShippingUIEnabled } = this.props;
 
         const { isBillingSameAsShipping, isMultiShippingMode } = this.state;
 
         if (!cart) {
             return;
         }
 
+        const isRenderMultiStaticConsignment = isNewMultiShippingUIEnabled && consignments.length > 1;
```
richapatel1510, line 435: Can we use `isMultiShippingMode` instead of checking consignments.length?

richapatel1510: Also what happens in case when someone chooses multi-shipping mode but only create 1 consignment, does that mean it is single shipping?

packages/core/src/app/shipping/StaticConsignmentItemList.tsx
```diff
@@ -28,7 +38,10 @@ const StaticConsignmentItemList: FunctionComponent<StaticConsignmentItemListProp
 
             <ul>
                 {lineItems.map((item) => (
-                    <li key={item.id}>{`${item.quantity} x ${item.name}`}</li>
+                    <li key={item.id}>
+                        <strong>{item.quantity} x </strong>{item.name} 
+                        {renderProductOptionDetails(item)}
+                    </li>
```
richapatel1510, line 44: The content seems similar to the one in ConsignmentLineItem, we can move it to a separate component and use it a both places.

packages/core/src/app/shipping/StaticMultiConsignment.tsx
richapatel1510, line 1: If the only difference in the component is styling and the heading, we can just add extra heading and apply appropriate className without needing new component so that we do not duplicate logic for stuff like paypalFastlane.

packages/core/src/app/checkout/Checkout.tsx
bc-peng: >Also what happens in case when someone chooses multi-shipping mode but only create 1 consignment, does that mean it is single shipping?

Choosing multi-shipping mode will update `isMultiShippingMode` as `true`, and the UI will stay in the multi-shipping mode until clicking the switch button.

packages/core/src/app/checkout/Checkout.tsx
```diff
@@ -423,30 +424,44 @@ class Checkout extends Component<
...
+
+        const summary = isRenderMultiStaticConsignment
```
bc-peng, line 437: Can we put this into a separate component file as `checkout.tsx` is quite big now?

packages/core/src/app/checkout/Checkout.tsx
bc-peng: Also, new changes should be behind the experiment in order to not break any existing multi-shipping (CSS) customizations merchants may have.

packages/core/src/app/shipping/StaticMultiConsignment.tsx
bc-mackxu: The address detail has a bit difference. 

#### Old one
image

#### New one
image

packages/core/src/app/checkout/Checkout.tsx
bc-mackxu: Thanks for suggestion. Looks like `isMultiShippingMode` already check the case only created 1 consignment, and will return false. I will change to `isNewMultiShippingUIEnabled && isMultiShippingMode`

https://github.com/bigcommerce/checkout-js/blob/CHECKOUT-8600/packages/core/src/app/shipping/isUsingMultiShipping.ts#L10-L21

packages/core/src/app/shipping/ConsignmentLineItemSummary.tsx
```diff
@@ -0,0 +1,33 @@
+import { PhysicalItem } from '@bigcommerce/checkout-sdk';
+import React, { FunctionComponent, memo } from 'react';
+
+import { MultiShippingTableItemWithType } from './MultishippingV2Type';
+
+export interface ConsignmentLineItemSummaryProps {
+    lineItems: MultiShippingTableItemWithType[] | PhysicalItem[]
+}
+
+const ConsignmentLineItemSummary: FunctionComponent<ConsignmentLineItemSummaryProps> = ({
```
richapatel1510, line 10: I suppose we can name it better since this is going to be used in allocate items modal as well. Something like `ConsignmentLineItemDetail`

packages/core/src/app/shipping/ShippingSummary.tsx
```diff
@@ -0,0 +1,55 @@
...
+    consignments.map((consignment) => (
+        <div className="staticConsignmentContainer" key={consignment.id}>
+            <StaticConsignment
+                cart={cart}
+                compactView={consignments.length < 2}
+                consignment={consignment}
+            />
+        </div>
+    ))
+}
```
richapatel1510, line 50: Can you please fix indentation for these lines.

packages/core/src/app/shipping/ShippingSummary.tsx
```diff
@@ -0,0 +1,55 @@
...
+                {consignments.map((consignment, index) => (
+                <div className="staticMultiConsignmentContainer" key={consignment.id}>
+                    <StaticMultiConsignment
+                        cart={cart}
+                        consignment={consignment}
+                        consignmentNumber={index + 1}
+                    />
+                </div>
+                ))}
+            </>
```
bc-peng, line 34: `<>` seens redundant.

```suggestion
                consignments.map((consignment, index) => (
                <div className="staticMultiConsignmentContainer" key={consignment.id}>
                    <StaticMultiConsignment
                        cart={cart}
                        consignment={consignment}
                        consignmentNumber={index + 1}
                    />
                </div>
                ))
```

packages/core/src/app/shipping/ShippingSummary.tsx
```diff
@@ -0,0 +1,55 @@
...
+        );
+    }
+
+    return (
+        <>
```
bc-peng, line 39: Same here.

packages/core/src/app/shipping/StaticMultiConsignment.tsx
```diff
@@ -0,0 +1,116 @@
...
+            data: { getShippingCountries },
+        },
+    } = context;
+
+    return {
+        countries: getShippingCountries(),
+    };
+}
+
+export default withCheckout(mapToStaticMultiConsignmentProps)(memo(StaticMultiConsignment));
```
bc-peng, line 116: As Richa suggested, we generally avoid using HOC in a new component.

packages/core/src/app/shipping/ShippingSummary.tsx
bc-mackxu: It is for a single parent wrapper. Wouldn't work without it.
image

packages/core/src/app/shipping/StaticMultiConsignment.tsx
bc-peng: Could we rename this file to `StaticMultiShippingConsignment.tsx` to emphasize that it is intended for multi-shipping use only?

bc-peng (approved):
LGTM.👏👏 Nice work.

richapatel1510 (approved):
Nice work 👏

PR #2075, opened 2024-10-30
feat(checkout): CHECKOUT-8790 Display detailed error when applying coupon fails on checkout

bc-shawnwang:
## What?
return and show detailed error message when applying coupon fails on checkout page

## Why?
now if coupon apply error is not on two specific error code, it always display "The gift certificate or coupon code is invalid".

This confuses our shoppers as they don’t always understand why a coupon code doesn’t apply, e.g because of currency not support it.

## Testing / Proof

**case 1: apply a coupon not for current cart currency**

before change:
image

🚫 too generic and not reflect the reason of failure

after change:
image

✅ which matches the server side detailed message ()

**case 2: apply a coupon that does not exists**

befor change:
image

after change:
image

✅ similar, but after change it fully reflect the error message service side send, not fixed.

**case 3: if apply coupon response has no detailed message or empty string**

after change:
Screenshot 2024-10-31 at 11 42 09 AM

✅ still use the default fallback value

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/cart/Redeemable.spec.tsx
+++ b/packages/core/src/app/cart/Redeemable.spec.tsx
@@ -24,7 +24,13 @@ describe('CartSummary Component', () => {
         errors: [{ code: 'min_purchase' }],
     } as RequestError;
     const appliedError = {
+        errors: [{ code: '', message: 'Specific error message'}],
+    } as RequestError;
+    const appliedErrorWithoutMessage = {
         errors: [{}],
+    } as RequestError;
+    const appliedErrorWitEmptyStringMessage = {
+        errors: [{code: '', message: ''}],
     } as RequestError;
 
     const RedeemableTestComponent = (props: RedeemableProps) => (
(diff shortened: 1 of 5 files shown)


packages/core/src/app/cart/Redeemable.tsx
```diff
@@ -136,7 +136,7 @@ const RedeemableForm: FunctionComponent<
                 return <TranslatedString id="redeemable.coupon_location_error" />;
 
             default:
-                return <TranslatedString id="redeemable.code_invalid_error" />;
+                return errorMessage;
```
bc-shawnwang, line 139: errorMessage from server side is already translated

bc-mackxu:
```
errorMessage?: string
```

It seems errorMessage could be undefined, should we keep the existing message as fallback in case no error message?

bc-shawnwang: https://github.com/bigcommerce/checkout-js/pull/2075/commits/faa5895329081fbc4f979efeb425ba476f267467

yes, and also I checked if it's empty string, because if error message is empty string and we do not change to use fallback value, it will be like this:
Screenshot 2024-10-31 at 11 44 21 AM

very confusing. 

after change it is:
Screenshot 2024-10-31 at 11 42 09 AM

packages/core/src/app/cart/Redeemable.tsx
```diff
@@ -136,7 +136,7 @@ const RedeemableForm: FunctionComponent<
                 return <TranslatedString id="redeemable.coupon_location_error" />;
 
             default:
-                return <TranslatedString id="redeemable.code_invalid_error" />;
+                return errorMessage?.trim() || <TranslatedString id="redeemable.code_invalid_error" />;
```
bc-mackxu, line 139: `errorMessage?.trim()` Do we need to trim the message?

packages/core/src/app/cart/Redeemable.spec.tsx
```diff
@@ -186,6 +210,68 @@ describe('CartSummary Component', () => {
...
+              />,
+            );
+        });
+
+        it('renders redeemable toggle link', () => {
+            expect(component.find('[data-test="redeemable-label"]')).toHaveLength(1);
+            expect(component.find('.redeemable-entry')).toHaveLength(0);
+        });
+
+        describe('when redeemable is clicked', () => {
```
bc-mackxu, line 233: I feel this is a new test group?
```suggestion
});
        describe('when redeemable is clicked', () => {
```

packages/core/src/app/cart/Redeemable.spec.tsx
```diff
@@ -86,6 +89,29 @@ describe('CartSummary Component', () => {
...
+                clearError={clearError}
+                isApplyingRedeemable={true}
+                onRemovedCoupon={onRemovedCoupon}
+                onRemovedGiftCertificate={onRemovedGiftCertificate}
+                shouldCollapseCouponCode={false}
+              />,
+            );
+        });
+
+        it('renders coupon invalid error', () => {
```
bc-peng, line 108:
```suggestion
        it('renders the generic coupon invalid error', () => {
```

packages/core/src/app/cart/Redeemable.tsx
bc-shawnwang: I think we better do it, because like I mentioned, " " will give you
image

we need to use fallback value instead (even through it's not a usual response, but for the robust)

packages/core/src/app/cart/Redeemable.spec.tsx
bc-shawnwang: I copy paste another test case: describe('when coupon code is collapsed') 's structure.

bc-mackxu (approved):

bc-peng (approved):

PR #2080, opened 2024-10-31
feat(checkout): CHECKOUT-8597 Choose Multi-shipping Shipping Method

bc-peng:
## What?
Add the ability to choose multi-shipping shipping option.

## Why?
A shopper can select a shipping method for a consignment.

## Testing / Proof

https://github.com/user-attachments/assets/eef9c9db-b093-47ea-81f8-b93030d0a8aa

@bigcommerce/team-checkout


packages/core/src/app/shipping/NewConsignment.tsx
```diff
@@ -74,8 +74,6 @@ const NewConsignment = ({
             if (error instanceof AssignItemFailedError) {
                 onUnhandledError(error);
             }
-        } finally {
-            toggleAllocateItemsModal();
```
bc-peng, line 78: This line creates a memory leak and not needed to close the modal actually.

packages/test-framework/src/react-testing-library-support/API.mock.ts
```diff
@@ -338,10 +340,110 @@ const checkoutSettingsWithCustomErrorFlashMessage = {
...
+            imageUrl: '',
+            cost: 10,
+            transitTime: '',
+            isRecommended: false,
+            additionalDescription: '',
+        },
+    ],
+};
+
+const cart: Checkout = {
```
bc-peng, line 443: Need to rename all cart as checkout in this file in another quick follow-up PR.

packages/core/src/app/shipping/MultiShippingV2.test.tsx
richapatel1510, line 1: 👏

packages/core/src/app/shipping/shippingOption/MultiShippingOptionsV2.tsx
```diff
@@ -0,0 +1,51 @@
...
+}
+
+export const MultiShippingOptionsV2 = ({
+    consignment,
+    isLoading,
+    shippingQuoteFailedMessage,
+}: MultiShippingOptionsV2Props) => {
+    const { checkoutService, checkoutState } = useCheckout();
+
+    const selectShippingOption = checkoutService.selectConsignmentShippingOption;
```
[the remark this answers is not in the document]
bc-peng: It seems that this component is quite different from the previous shipping option, primarily because the design requires using a radio button instead of a ChecklistItem. 
As a result, this component may not be able to get anything from `./shippingOptions`, and it can be considered as a multi-shipping version of `./shippingOptions`.

packages/core/src/app/shipping/shippingOption/MultiShippingOptionsListItemV2.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+    const selectThisOption = useCallback(() => {
+        handleSelect(shippingOption.id);
+    }, [handleSelect, shippingOption.id]);
+
+    useEffect(() => {
+        if (!selectedShippingOptionId && shippingOption.isRecommended) {
+            selectThisOption();
+        }
+    }, [selectedShippingOptionId, shippingOption.isRecommended, selectThisOption]);
```
bc-mackxu, line 33: If we move this logic before rendering the list, could that eliminate the need for `useEffect`?

bc-peng: If I understand correctly, regardless of where we place this block of code, except the API layer, it will always require the use of `useEffect`.
Because, after creating a new consignment with items and an address, the shipping option may initially be empty, and we need to compare the values and set the recommended one..

packages/core/src/app/shipping/NewConsignment.tsx
richapatel1510: Just identified the memory leak is due to line 74 and not this one.

bc-peng: Do we still need this line to close the modal?

richapatel1510: We don't need this line can get rid of it, just that it does not solve the issue.

bc-peng: Adding it back as it failed the test, the modal was not closed. 🤦🏻‍♂️
We can address the memory issue separately.
image

richapatel1510 (approved):

PR #2095, opened 2024-11-07
CHECKOUT-8775: Toggle from multi shipping to single shipping and vice versa

richapatel1510:
## What?
Handle actions when toggling from multi shipping to single shipping and vice versa

1. **When moving from single shipping to multi shipping in SF checkout**
- Delete the consignment so that multi shipping consignments can be freshly started.

2. **When moving from multi shipping to single shipping in SF checkout**
- Show the warning popup that all the consignments will be lost.

## Why?
To improve the multi shipping user experience

## Testing / Proof

- CI checks
- Manual testing

https://github.com/user-attachments/assets/0ac0148c-7f7c-4986-8b7b-75cc7482bbef

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/shipping/ConsignmentListItem.tsx
+++ b/packages/core/src/app/shipping/ConsignmentListItem.tsx
@@ -42,15 +42,17 @@ const ConsignmentListItem: FunctionComponent<ConsignmentListItemProps> = ({
     return (
         <div className='consignment-container'>
             <div className='consignment-header'>
-                <h3><TranslatedString data={{ consignmentNumber }} id="shipping.multishipping_consignment_index_heading" /></h3>
-                    <a
-                        className="delete-consignment"
-                        data-test="delete-consignment-button"
-                        href="#"
-                        onClick={preventDefault(handleClose)}
-                    >
-                        <IconClose size={IconSize.Small}/>
-                    </a>
+                <h3>
+                    <TranslatedString data={{ consignmentNumber }} id="shipping.multishipping_consignment_index_heading" />
+                </h3>
+                <a
+                    className="delete-consignment"
+                    data-test="delete-consignment-button"
+                    href="#"
+                    onClick={preventDefault(handleClose)}
+                >
+                    <IconClose size={IconSize.Small} />
+                </a>
             </div>
             <ConsignmentAddressSelector
                 consignment={consignment}
--- a/packages/core/src/app/shipping/Shipping.test.tsx
+++ b/packages/core/src/app/shipping/Shipping.test.tsx
@@ -140,7 +140,7 @@ describe('Shipping component', () => {
             });
         });
 
-        it.only('opens confirmation dialog on clicking the link and calls onToggleMultiShipping when confirm is clicked', async () => {
+        it('opens confirmation dialog on clicking the link and calls onToggleMultiShipping when confirm is clicked', async () => {
             render(<ComponentTest {...defaultProps} isMultiShippingMode={true} />);
 
             const shippingModeToggle = await screen.findByTestId("shipping-mode-toggle");
--- a/packages/core/src/app/shipping/Shipping.test.tsx
+++ b/packages/core/src/app/shipping/Shipping.test.tsx
@@ -1,0 +1,163 @@
+import '@testing-library/jest-dom';
+import {
+    CheckoutSelectors,
+    CheckoutService,
+    createCheckoutService,
+} from '@bigcommerce/checkout-sdk';
+import userEvent from '@testing-library/user-event';
+import React, { FunctionComponent } from 'react';
+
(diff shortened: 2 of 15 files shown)


packages/core/src/app/shipping/ShippingHeader.tsx
```diff
@@ -36,7 +48,25 @@ const ShippingHeader: FunctionComponent<ShippingHeaderProps> = ({
...
+                            headerId="shipping.ship_to_single_action"
+                            isModalOpen={isModalOpen}
+                            messageId="shipping.ship_to_single_message"
+                            onRequestClose={() => setIsModalOpen(false)}
+                        />
+                        <a
+                            data-test="shipping-mode-toggle"
+                            href="#"
+                            onClick={preventDefault(() => setIsModalOpen(true))}
+                        >
```
bc-peng, line 64: The code indentation is missing.

packages/core/src/app/shipping/Shipping.tsx
```diff
@@ -219,6 +223,18 @@ class Shipping extends Component<ShippingProps & WithCheckoutShippingProps, Ship
...
 
+        if (!isMultiShippingMode && isNewMultiShippingUIEnabled) {
+            this.setState({ isInitializing: true });
+
+            try {
+                await deleteConsignments();
+            } catch (error) {
+                onUnhandledError(error);
+            } finally {
+                this.setState({ isInitializing: false });
```
bc-peng, line 234: I see these four lines are identical to the switching to single shipping case.

Perhaps we can place the `if` condition inside the `try` block to minimize code repetition.

packages/core/src/app/shipping/Shipping.test.tsx
```diff
@@ -0,0 +1,163 @@
...
+                    hasMultiShippingEnabled: true,
+                    features: {
+                        ...getStoreConfig().checkoutSettings.features,
+                        "PROJECT-4159.improve_multi_address_shipping_ui": true,
+                    },
+                },
+            });
+        });
+
+        it.only('opens confirmation dialog on clicking the link and calls onToggleMultiShipping when confirm is clicked', async () => {
```
bc-peng, line 143: Is the `only` intentional?

packages/core/src/app/shipping/ConsignmentListItem.tsx
```diff
@@ -42,15 +42,17 @@ const ConsignmentListItem: FunctionComponent<ConsignmentListItemProps> = ({
...
-                    </a>
+                <h3>
+                    <TranslatedString data={{ consignmentNumber }} id="shipping.multishipping_consignment_index_heading" />
+                </h3>
+                <a
+                    className="delete-consignment"
+                    data-test="delete-consignment-button"
+                    href="#"
+                    onClick={preventDefault(handleClose)}
+                >
```
bc-peng, line 53: Ident missing.

richapatel1510: `<h3>` and `<a>` tags are siblings here, so should be at same ident.

packages/core/src/app/shipping/Shipping.test.tsx
richapatel1510: No, not intentional. Thanks for catching this 🙇‍♀️

packages/core/src/app/shipping/ShippingHeader.tsx
richapatel1510: `<ConfirmationModal>` and `<a>` tags are siblings here, so should be at same ident.

bc-peng: 🤦🏻‍♂️Oh mine....Yes, I missed the closing `</a>`.

packages/core/src/app/shipping/ConsignmentListItem.tsx
bc-peng: Same...🤦🏻‍♂️

bc-mackxu:
In the screen record 28 second, curiosity why the continue button went to top right screen?

image

richapatel1510:
That seems to be happening because its making multiple API calls to delete the consignment before moving back to single shipping, this doesn't happen when we only have 1 consignment in the multi shipping mode.

bc-mackxu:
@richapatel1510 It looks like a broken page to me, I think we should avoid this.

richapatel1510:
Updated the styling of LoadingSkeleton to address the reloading issue

https://github.com/user-attachments/assets/4ff7ada6-a38b-4f56-82d0-54b3d200d895

bc-peng (approved):
LGTM👏👏

bc-mackxu (approved):
👏

PR #2101, opened 2024-11-14
feat(checkout): CHECKOUT-8774 Display item split tooltip

bc-mackxu:
## What?
Display item split tooltip when items split into individual item.

## Why?
To improve user experience

## Testing / Proof
- [x] unit test
- [x] manual test
image
image
image

@bigcommerce/team-checkout


packages/core/src/app/shipping/hooks/useMultishippingConsignmentItems.ts
```diff
@@ -13,6 +14,27 @@ const calculateShippableItemsCount = (items: MultiShippingTableItemWithType[]):
...
+  
+      if (hash && processedHashes.has(hash)) {
+        return true;
+      }
+  
+      if (hash) {
+        processedHashes.add(hash);
+      }
+  
+      return false;
```
bc-peng, line 34: We may want to avoid seeing two `if(hash)` in a row.
```suggestion
    if (!hash) {
      return false;
    }

    if (processedHashes.has(hash)) {
      return true;
    }

    processedHashes.add(hash);
    return false;
```

bc-peng (approved):
LGTM.👏👏

PR #2105, opened 2024-11-20
fix(checkout): CHECKOUT-8869 Handle Errors in Extension Rendering

bc-peng:
## What?
Catches error during the rendering of a checkout extension.

## Why?
During the brief pre-rendering of the shipping step, an extension may not finish loading, causing the render extension action to hang and resulting in an error from the checkout-sdk

This PR handles those errors by logging them out in the console.
## Testing / Proof
image

### DIFF
--- a/packages/checkout-extension/src/ExtensionService.ts
+++ b/packages/checkout-extension/src/ExtensionService.ts
@@ -6,6 +6,8 @@ import {
 } from '@bigcommerce/checkout-sdk';
 import React from 'react';
 
+import { ErrorLevelType, ErrorLogger } from '@bigcommerce/checkout/error-handling-utils';
+
 import { ExtensionAction } from './ExtensionProvider';
 import * as handlerFactories from './handlers';
 import { CommandHandler } from './handlers/CommandHandler';
@@ -16,6 +18,7 @@ export class ExtensionService {
     constructor(
         private checkoutService: CheckoutService,
         private dispatch: React.Dispatch<ExtensionAction>,
+        private errorLogger: ErrorLogger,
     ) {}
 
     async loadExtensions(): Promise<void> {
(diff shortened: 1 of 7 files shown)


packages/checkout-extension/src/ExtensionService.ts
```diff
@@ -58,9 +58,18 @@ export class ExtensionService {
...
-
-        this.registerHandlers(extension);
+        try {
+            await this.checkoutService.renderExtension(container, region);
+
+            this.registerHandlers(extension);
+        } catch (error: unknown) {
+            if (error instanceof Error) {
+                // eslint-disable-next-line no-console
+                console.error(
```
animesh1987, line 68: Shall we use `ErrorLogger` here?

bc-peng: I understand `ErrorLogger` ends up at Sentry.
An extension iframe's detachment error may happen regularly, shall we just keep it in the console?
I see we have `ConsoleErrorLogger`, I can replace here with that.

bc-peng: Thanks for the feedback, I have replaced it.
This is an updated testing screenshot.
image

richapatel1510 (approved):

packages/checkout-extension/src/ExtensionService.ts
davidchin: @bc-peng Have you figured out why `Shipping` component is momentarily rendered at all? I'd expect that after customer/shipping/billing details are provided, when you refresh the page, we should only render the `Payment` component, not the `Shipping` component I suspect it is probably because for some reason `isComplete` is `false` here: https://github.com/bigcommerce/checkout-js/blob/d6d1218e0628e462d79bfc36fdd278c7936663ea/packages/core/src/app/checkout/getCheckoutStepStatuses.ts#L181

But it shouldn't be because shipping details are already provided.

Having said that, we need to have what you've done so far as a safety net - just in case the extensions are detached while loading. However, it seems something is wrong with the state machine that also needs fixing.

davidchin (approved):
My question could be tackled separately 👍

packages/checkout-extension/src/ExtensionService.ts
animesh1987: Good catch @davidchin I think might be good to check and test as part of this PR itself. It could be because of checking for additional conditions like fastlane etc. Or could it be that since we changed for extension to start loading while we fetch other shipping data if customer email already complete?

bc-peng: For the record, we have implemented this change to prevent the shipping form from rendering after a page refresh.

https://github.com/bigcommerce/checkout-js/pull/2111

PR #2107, opened 2024-11-22
CHECKOUT-8850 address design comments for new multishipping UI

richapatel1510:
## What?

1. Add loader to `Allocate` button in the allocate items modal.
2. In the address selector, the placeholder text should be 'Choose a shipping address'
3. For the 'Add New Address' modal, let's use the 'Cancel' button style from the consignment modal for consistency
4. Show cart has changed popup if the cart is changed via some external sources.

## Why?
To improve the UI/UX for checkout multishipping 
 
## Testing / Proof
**Manual testing** 

Screenshot 2024-11-22 at 2 40 05 PM
Screenshot 2024-11-22 at 2 39 51 PM
Screenshot 2024-11-22 at 2 40 20 PM
Screenshot 2024-11-22 at 2 40 41 PM

https://github.com/user-attachments/assets/345dc2d2-b470-4314-8d4a-ff190f1daf84

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/address/AddressFormModal.tsx
+++ b/packages/core/src/app/address/AddressFormModal.tsx
@@ -3,7 +3,6 @@ import { FormikProps, withFormik } from 'formik';
 import React, { FunctionComponent } from 'react';
 import { lazy } from 'yup';
 
-import { preventDefault } from '@bigcommerce/checkout/dom-utils';
 import { TranslatedString, withLanguage, WithLanguageProps } from '@bigcommerce/checkout/locale';
 
 import { Button, ButtonVariant } from '../ui/button';
@@ -59,13 +58,12 @@ const SaveAddress: FunctionComponent<
                 shouldShowSaveAddress={false}
             />
             <div className="form-actions">
-                <a
-                    className="button optimizedCheckout-buttonSecondary"
-                    href="#"
-                    onClick={preventDefault(onRequestClose)}
-                >
+                <Button
+                    onClick={onRequestClose}
+                    variant={ButtonVariant.Secondary}>
                     <TranslatedString id="common.cancel_action" />
-                </a>
+                </Button>
+
 
                 <Button
                     disabled={isLoading}
--- a/packages/core/src/app/shipping/AllocateItemsModal.tsx
+++ b/packages/core/src/app/shipping/AllocateItemsModal.tsx
@@ -31,6 +31,7 @@ interface AllocateItemsModalProps {
     assignedItems?: MultiShippingTableData;
     onAllocateItems(consignmentLineItems: ConsignmentLineItem[]): void;
     onUnassignItem?(itemToDelete: MultiShippingTableItemWithType): void;
+    isLoading: boolean;
 }
 
 const AllocateItemsModal: FunctionComponent<AllocateItemsModalProps & FormikProps<AllocateItemsModalFormValues>> = ({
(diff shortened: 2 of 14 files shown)


packages/core/src/app/shipping/AllocateItemsModal.tsx
```diff
@@ -103,8 +105,21 @@ const AllocateItemsModal: FunctionComponent<AllocateItemsModalProps & FormikProp
...
     const modalFooter = (
         <>
-            <Button onClick={onRequestClose} variant={ButtonVariant.Secondary}><TranslatedString id="shipping.multishipping_items_allocate_cancel" /></Button>
-            <Button disabled={!dirty} onClick={submitForm} type="submit" variant={ButtonVariant.Primary}>{hasItemsAssigned ? <TranslatedString id="shipping.multishipping_items_allocate_save" /> : <TranslatedString id="shipping.multishipping_items_allocate_allocate" />}</Button>
+            <Button onClick={onRequestClose} variant={ButtonVariant.Secondary}>
+                <TranslatedString id="shipping.multishipping_items_allocate_cancel" />
+            </Button>
+            <Button
+                disabled={!dirty}
+                isLoading={isLoading}
```
bc-mackxu, line 113: Should we disable the cancel button while loading?

richapatel1510: Good catch. Disabled it when loading [here](https://github.com/bigcommerce/checkout-js/pull/2107/commits/cd70595da7298cf5190c7c7930f62411d31b32fc)

packages/core/src/app/address/AddressFormModal.tsx
```diff
@@ -59,13 +58,12 @@ const SaveAddress: FunctionComponent<
                 shouldShowSaveAddress={false}
             />
             <div className="form-actions">
-                <a
-                    className="button optimizedCheckout-buttonSecondary"
-                    href="#"
-                    onClick={preventDefault(onRequestClose)}
-                >
+                <Button
+                    onClick={onRequestClose}
```
animesh1987, line 62: This seems like a change in the overall behaviour of the modal in general? Is that expected?

bc-peng: The PR looks good to me besides this question.
I did notice we have three `cancel` buttons using the `<Button /`> not `<a/>`.

animesh1987: Might be good to go in the commit history to see if there is a reason 🤔

richapatel1510: Looking into history, this has not been changed since it was first implemented ([ref](https://github.com/bigcommerce/checkout-js/commit/450c9f87c241e59e8a72ae26dc407fe1f5b08a4e#diff-7223c63a2de6c7af5e87e73bc696db9509af45606519268b5f58a0cf62ab28d4)). I guess it is implemented as `<a>` tag because we do not have button style to look like a link, no other reason that I can think of. Even in this PR, I have added [this style](https://github.com/bigcommerce/checkout-js/pull/2107/files#diff-3db8c7765aebecc0f9859a3c3f636d8227097b2c550bd3c98a01ce101bbe2a2e) to remove button border to make it look like an actionable text. We can change it back to `<a>` tag if you suggest. Moreover, this modal is only used in multi shipping (old and new both)

animesh1987: Thanks @richapatel1510 for the detailed analysis, looks good to me to change to Button in this case. We just have to be diligent around the changes.

animesh1987 (approved):

bc-peng (approved):

PR #2111, opened 2024-11-28
feat(checkout): CHECKOUT-8869 Remove Available Shipping Option Check

bc-peng:
## What?
Remove available shipping option check when determining if a consignment has selected a shipping method..

## Why?
This change will avoid always loading available shipping methods after a checkout page refresh.

## Testing / Proof
CI.

### DIFF
--- a/packages/core/src/app/shipping/hasSelectedShippingOptions.ts
+++ b/packages/core/src/app/shipping/hasSelectedShippingOptions.ts
@@ -9,13 +9,7 @@ export default function hasSelectedShippingOptions(consignments: Consignment[]):
     return every(
         consignments,
         (consignment) =>
-            (consignment.selectedShippingOption &&
-                consignment.selectedShippingOption.id &&
-                // Selected option is available
-                consignment.availableShippingOptions &&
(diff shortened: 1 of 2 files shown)


packages/core/src/app/shipping/hasSelectedShippingOptions.ts
```diff
@@ -9,13 +9,7 @@ export default function hasSelectedShippingOptions(consignments: Consignment[]):
...
         consignments,
         (consignment) =>
-            (consignment.selectedShippingOption &&
-                consignment.selectedShippingOption.id &&
-                // Selected option is available
-                consignment.availableShippingOptions &&
-                consignment.availableShippingOptions.filter(
-                    ({ id }) => id === consignment.selectedShippingOption?.id,
-                ).length) ||
+            (consignment.selectedShippingOption && consignment.selectedShippingOption.id) ||
```
davidchin, line 12: For context, this check is no longer unnecessary since the API automatically invalidates the selected shipping option if it's unavailable after updating the address.

davidchin (approved):

animesh1987 (approved):

PR #2112, opened 2024-12-02
fix(shipping): CHECKOUT-7560 add max length validation for address fields

richapatel1510:
## What?
Add validation of form field for max length for address fields. Do not allow to continue shipping if address fields are invalid.

## Why?
**To fix:** You are able to bypass the Address line character limit on Checkout if a customer already has a saved address

## Testing / Proof

- CI checks
- Manual testing 

**Experiment off**

https://github.com/user-attachments/assets/a690bf08-9fa0-46c8-ae36-dc0ccf0c50fc

**Experiment on**

https://github.com/user-attachments/assets/a23af2c0-e5c8-42d6-8cc5-650e251d98dd

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/address/SingleLineStaticAddress.tsx
+++ b/packages/core/src/app/address/SingleLineStaticAddress.tsx
@@ -1,5 +1,4 @@
 import { Address } from '@bigcommerce/checkout-sdk';
-import { isEmpty } from 'lodash';
 import React from "react";
 
 import { useCheckout } from "@bigcommerce/checkout/payment-integration-api";
@@ -7,7 +6,7 @@ import { useCheckout } from "@bigcommerce/checkout/payment-integration-api";
 import { isExperimentEnabled } from '../common/utility';
 
 import AddressType from "./AddressType";
-import isValidAddress from "./isValidAddress";
+import isValidStaticAddress from './isValidStaticAddress';
 
 export interface SingleLineStaticAddressProps {
     address: Address;
@@ -56,15 +55,7 @@ const SingleLineStaticAddress = ({ address, type }: SingleLineStaticAddressProps
                 ? getShippingAddressFields(address.countryCode)
                 : undefined;
 
-    let isValid = !isEmpty(address);
-    
-    if (!!fields && !validateAddressFields) {
-        isValid = isValidAddress(
-            address,
-            fields.filter((field) => !field.custom),
-        );
-    }
-
+    const isValid = isValidStaticAddress(address, validateAddressFields, fields);
 
     return !isValid ? null : (
         <div className="vcard checkout-address--static" data-test="static-address">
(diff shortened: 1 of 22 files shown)


packages/core/src/app/address/SingleLineStaticAddress.tsx
```diff
@@ -36,23 +38,33 @@ export const getAddressContent: (value: Address) => string = ({
...
             : type === AddressType.Shipping
                 ? getShippingAddressFields(address.countryCode)
                 : undefined;
 
-    const isValid = !fields
-        ? !isEmpty(address)
-        : isValidAddress(
+    let isValid = !isEmpty(address);
+    
+    if (!!fields && !validateAddressFields) {
```
bc-peng, line 61: I believe `!![]` is `true`, are we just expecting fields to be an array here?

packages/core/src/app/address/StaticAddress.tsx
```diff
@@ -30,18 +31,26 @@ export interface StaticAddressEditableProps extends StaticAddressProps {
...
+    }) => {
+        const address = localizeAddress(addressWithoutLocalization, countries);
+        let isValid = !isEmpty(address);
+        
+        if (!!fields && !validateAddressFields) {
+            isValid = isValidAddress(
+                address,
+                fields.filter((field) => !field.custom),
+            );
+        }
```
bc-peng, line 53: Perhpas we can build a helper function as it is being reused.

packages/core/src/app/address/isValidStaticAddress.tsx
```diff
@@ -0,0 +1,21 @@
...
+        
+    if (fields && !validateAddressFields) {
+        isValid = isValidAddress(
+            address,
+            fields.filter((field) => !field.custom),
+        );
+    }
+
+    return isValid;
+}
```
animesh1987, line 21: EOL.

packages/core/src/app/address/isValidStaticAddress.tsx
```diff
@@ -0,0 +1,21 @@
+import { Address, FormField } from "@bigcommerce/checkout-sdk";
+import { isEmpty } from "lodash";
+
+import isValidAddress from "./isValidAddress";
+
+export default function isValidStaticAddress(
```
animesh1987, line 6: If I see this correctly, this is the old logic when the experiment is off?

richapatel1510: That's right, its the old logic, which used to display the empty option in address selector. Once the experiment is rolled out we may get rid of this logic and will only need to check if address is empty, since we want to show the address in the selector even if it is invalid.

animesh1987 (approved):

bc-peng (approved):

PR #2114, opened 2024-12-05
fix(shipping): CHECKOUT-8872 Fix Missing Shipping Option

bc-peng:
### Fixed:
1.Selecting a shipping method for one consignment no longer updates the shipping method of another consignment.

2.When creating a new consignment, the selected shipping methods of existing consignments will no longer be randomly lost.

### Updated:
1.The recommended shipping method is now automatically selected at the time of consignment creation, moving the trigger from the display of shipping methods.

## Why?
Address existing API limitations and fix bugs.

### Before

https://github.com/user-attachments/assets/fb6fa579-9c4e-4ce0-9826-80524ec1ee07

#### Updating Shipping Method
https://github.com/user-attachments/assets/983e7956-c039-4936-9910-44802dcb413d
#### Creating New Consignment
https://github.com/user-attachments/assets/1f41695a-835d-424a-aa06-4462fa1a3669

### DIFF
--- a/packages/core/src/app/shipping/ConsignmentAddressSelector.tsx
+++ b/packages/core/src/app/shipping/ConsignmentAddressSelector.tsx
@@ -10,6 +10,7 @@ import { EMPTY_ARRAY, isFloatingLabelEnabled } from "../common/utility";
 
 import { AssignItemFailedError, AssignItemInvalidAddressError } from "./errors";
 import { MultiShippingConsignmentData } from "./MultishippingV2Type";
+import { setRecommendedOrMissingShippingOption } from './MultiShippingSelectShippingOption';
 
 interface ConsignmentAddressSelectorProps {
     consignment?: MultiShippingConsignmentData;
@@ -35,9 +36,19 @@ const ConsignmentAddressSelector = ({
 
     const {
         checkoutState: {
-            data: { getShippingCountries, getCustomer, getConfig, getShippingAddressFields: getFields },
+            data: {
+                getShippingCountries,
+                getCustomer,
+                getConfig,
+                getConsignments: getPreviousConsignments,
+                getShippingAddressFields: getFields,
+            },
         },
-        checkoutService: { assignItemsToAddress: assignItem, createCustomerAddress },
+        checkoutService: {
+            assignItemsToAddress: assignItem,
+            createCustomerAddress,
+            selectConsignmentShippingOption,
+        },
     } = useCheckout();
 
     const countries = getShippingCountries() || EMPTY_ARRAY;
@@ -49,7 +60,7 @@ const ConsignmentAddressSelector = ({
     }
 
     const isFloatingLabelEnabledFlag = isFloatingLabelEnabled(config.checkoutSettings);
-    // TODO: add filter for addresses 
+    // TODO: add filter for addresses
     const addresses = customer.addresses || EMPTY_ARRAY;
     const {
         checkoutSettings: {
@@ -73,11 +84,22 @@ const ConsignmentAddressSelector = ({
         }
 
         try {
-            await assignItem({
+            const {
+                data: { getConsignments },
+            } = await assignItem({
                 address,
                 lineItems: consignment.lineItems.map(({ id, quantity }) => ({ itemId: id, quantity })),
             });
 
+            const currentConsignments = getConsignments();
+
+            if (currentConsignments) {
+                await setRecommendedOrMissingShippingOption(
+                    getPreviousConsignments() ?? [],
+                    currentConsignments,
+                    selectConsignmentShippingOption,
+                );
+            }
         } catch (error) {
             if (error instanceof Error) {
                 onUnhandledError(new AssignItemFailedError(error));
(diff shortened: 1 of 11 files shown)


packages/core/src/app/shipping/shippingOption/MultiShippingOptionsListItemV2.tsx
```diff
@@ -22,29 +22,22 @@ export const MultiShippingOptionsListItemV2: FunctionComponent<
         </>
     );
 
-    const selectThisOption = useCallback(() => {
```
bc-peng, line 25: > Selecting a shipping method for one consignment no longer updates the shipping method of another consignment.

This FE bug has two casues:
 1. `useCallback` here may return another consignment's `handleSelect`, due to `consignment` is not on the dependency list.
 2. `<RadioInput />` was in `<button />`. As a result, clicking `<RadioInput />` may trigger `<button />`'c `onClick` event twice.

packages/core/src/app/shipping/MultiShippingSelectShippingOption.ts
richapatel1510, line 1: Looks like its a component by the file name, but it works as a util. Shall we move this to utils folder, or maybe just rename it?

richapatel1510 (commented):
Apart from one comment, rest LGTM 👍 . Can you also attach the screen recording for adding new consignment with a similar item present in existing consignment.

packages/core/src/app/shipping/MultiShippingSelectShippingOption.ts
animesh1987: Good catch, this should be named as a regular file. Also probably named to `setRecommendedOrMissingShippingOption` to match with the export.

packages/core/src/app/shipping/ConsignmentAddressSelector.tsx
```diff
@@ -73,11 +84,22 @@ const ConsignmentAddressSelector = ({
...
+            const {
+                data: { getConsignments },
+            } = await assignItem({
                 address,
                 lineItems: consignment.lineItems.map(({ id, quantity }) => ({ itemId: id, quantity })),
             });
 
+            const currentConsignments = getConsignments();
+
+            if (currentConsignments) {
```
animesh1987, line 96: Do we need to check the length of consignments? Since `[]` is truthy as well?

packages/core/src/app/shipping/shippingOption/MultiShippingOptionsListItemV2.tsx
animesh1987: Good catch 👏

packages/core/src/app/shipping/NewConsignment.tsx
```diff
@@ -61,22 +64,37 @@ const NewConsignment = ({
...
         } catch (error) {
             if (error instanceof AssignItemFailedError) {
                 onUnhandledError(error);
             }
         } finally {
             toggleAllocateItemsModal();
             setIsAddShippingDestination(false);
+            resetErrorConsignmentNumber();
+
+            if (currentConsignments) {
```
animesh1987, line 91: Same as above.

packages/core/src/app/shipping/MultiShippingSelectShippingOption.ts
```diff
@@ -0,0 +1,39 @@
...
+            const previousShippingOptionId = previousShippingOptions.get(consignment.id);
+
+            if (previousShippingOptionId) {
+                // eslint-disable-next-line no-await-in-loop
+                await selectConsignmentShippingOption(consignment.id, previousShippingOptionId);
+            } else {
+                const recommendedOption = consignment.availableShippingOptions?.find(
+                    (option) => option.isRecommended,
+                );
+
```
animesh1987, line 31: Can we refactor to this, this way feels more reading in a sequence

```suggestion
       if (previousShippingOptionId) {
            await selectConsignmentShippingOption(consignment.id, previousShippingOptionId);
            continue;
       }
       
       const recommendedOption = consignment.availableShippingOptions?.find(
            (option) => option.isRecommended
        );

```

bc-peng: Updated.

packages/core/src/app/shipping/NewConsignment.tsx
bc-peng: Same.

packages/core/src/app/shipping/ConsignmentAddressSelector.tsx
bc-peng: The condition is to make Typescript happy as `getConsignments` can return `undefined` by definition, not strictly required.
Maybe when creating a consignment fails, checking length here can skip `setRecommendedOrMissingShippingOption` and save some time. I have added the length check in this case.

packages/core/src/app/shipping/MultiShippingSelectShippingOption.ts
bc-peng: Thanks for the feedback.
Adding `MultiShipping` ahead of the name was to mark the helper method for multi-shipping use only.
I have moved it to `utils` and renamed it as `setRecommendedOrMissingShippingOption.ts`.

bc-peng:
> Apart from one comment, rest LGTM 👍 . If you can also attach the screen recording for adding new consignment with a similar item present in existing consignment would be nice.

@richapatel1510 I have added a new video to the description.

richapatel1510 (approved):

animesh1987 (approved):

PR #2115, opened 2024-12-05
chore(shipping): CHECKOUT-8871 Remove bundled items from being displayed in items allocation modal

animesh1987:
## What?
Remove bundled items from being displayed in items allocation modal

## Why?
Bundled items are shipped with the parent item, therefore should not be available to be allocated as a separate item in the modal as that causes confusion because bundled item should be shipped with the parent element.

## Testing / Proof
- CI
- Screencast

https://github.com/user-attachments/assets/9d22775a-bdb7-4481-b387-8abaf08f9ea6

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/shipping/hooks/useMultishippingConsignmentItems.ts
+++ b/packages/core/src/app/shipping/hooks/useMultishippingConsignmentItems.ts
@@ -10,6 +10,15 @@ interface MultiShippingConsignmentItemsHook {
     consignmentList: MultiShippingConsignmentData[];
 }
 
+// TODO: consolidate this from /app/order/removeBundledItems
+function removeBundledItems(lineItems: LineItemMap): LineItemMap {
+    return {
+        ...lineItems,
+        physicalItems: lineItems.physicalItems.filter((item) => typeof item.parentId !== 'string'),
+        digitalItems: lineItems.digitalItems.filter((item) => typeof item.parentId !== 'string'),
+    };
+}
+
 const calculateShippableItemsCount = (items: MultiShippingTableItemWithType[]): number => {
     return items.reduce((total, item) => total + item.quantity, 0);
 };
@@ -52,8 +61,7 @@ function mapConsignmentsItems(
     lineItems.physicalItems.forEach((item) => {
         unassignedItemsMap.set(item.id.toString(), { ...item, type: LineItemType.Physical });
         itemHashMap.set(item.id.toString(), generateItemHash(item));
-    }
-    );
+    });
     lineItems.customItems?.forEach((item) =>
         unassignedItemsMap.set(item.id, { ...item, type: LineItemType.Custom }),
     );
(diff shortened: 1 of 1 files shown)


packages/core/src/app/shipping/hooks/useMultishippingConsignmentItems.ts
```diff
@@ -73,6 +81,8 @@ function mapConsignmentsItems(
             }
         });
 
+        console.log(consignmentLineItems);
+
```
richapatel1510, line 85:
```suggestion
```

TomA-R (approved):

packages/core/src/app/shipping/hooks/useMultishippingConsignmentItems.ts
```diff
@@ -52,8 +61,7 @@ function mapConsignmentsItems(
     lineItems.physicalItems.forEach((item) => {
         unassignedItemsMap.set(item.id.toString(), { ...item, type: LineItemType.Physical });
         itemHashMap.set(item.id.toString(), generateItemHash(item));
-    }
-    );
+    });
```
richapatel1510, line 64: Shall we filter the items here (or before this) so that the `itemHash` and all does not need to be generated for picklist items.

packages/core/src/app/shipping/hooks/useMultishippingConsignmentItems.ts
```diff
@@ -10,6 +10,15 @@ interface MultiShippingConsignmentItemsHook {
     consignmentList: MultiShippingConsignmentData[];
 }
 
+// TODO: consolidate this from /app/order/removeBundledItems
+function removeBundledItems(lineItems: LineItemMap): LineItemMap {
+    return {
+        ...lineItems,
+        physicalItems: lineItems.physicalItems.filter((item) => typeof item.parentId !== 'string'),
+        digitalItems: lineItems.digitalItems.filter((item) => typeof item.parentId !== 'string'),
```
richapatel1510, line 18: Can we not directly check `!item.parentId`, like https://github.com/bigcommerce/manual-order-ui/blob/main/src/Consignment/hooks/useMultiShippingConsignmentItems.tsx#L60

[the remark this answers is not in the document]
animesh1987: Yes, that is correct. Parent Id is either string or null https://github.com/bigcommerce/checkout-sdk-js/blob/master/packages/core/src/cart/line-item.ts#L68.

animesh1987: We can do that too, but this feels a better check in my opinion. Plus after this I am going to consolidate this with the existing function that has the bundled item logic.

bc-peng: Manual order does not filter out digital child item, perhaps we should add it asap. 🤦🏻‍♂️

richapatel1510 (approved):

PR #2116, opened 2024-12-06
feat(payment): STRIPE-546 allow billing address editing for all Google Pay providers

bc-dronov:
## What?
Allow billing address editing for all Google Pay providers.

Related PR:
https://github.com/bigcommerce/checkout-sdk-js/pull/2747

## Why?
Because sometimes the locality field from GP is empty, we will add stateOrProvinceCode instead, and also allow the user to edit the billing address if needed.

## Testing / Proof
Manual testing.
Before:
before
After:
after

Stripe Google Pay

https://github.com/user-attachments/assets/9ed50aac-ea91-4dad-9c0c-1e4d5d4dd655

PPCP Google Pay

https://github.com/user-attachments/assets/a9f35bad-d32f-4f24-9ace-13e9e5b99882

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/checkout/getCheckoutStepStatuses.ts
+++ b/packages/core/src/app/checkout/getCheckoutStepStatuses.ts
@@ -89,7 +89,8 @@ const getBillingStepStatus = createSelector(
             ? data.getBillingAddressFields(billingAddress.countryCode)
             : EMPTY_ARRAY;
     },
-    (checkout, billingAddress, billingAddressFields) => {
+    ({ data }: CheckoutSelectors) => data.getConfig(),
+    (checkout, billingAddress, billingAddressFields, config) => {
         const hasAddress = billingAddress
             ? isValidAddress(billingAddress, billingAddressFields)
             : false;
@@ -120,6 +121,25 @@ const getBillingStepStatus = createSelector(
                 isActive: false,
                 isComplete: isAmazonPayBillingStepComplete,
                 isEditable: isAmazonPayBillingStepComplete && hasCustomFields,
+                isRequired: true,
+            };
+        }
+
+        const isGooglePayBillingAddressEditingEnabled = isExperimentEnabled(
+            config?.checkoutSettings,
+            'STRIPE-546.allow_billing_address_editing_for_all_Google_Pay_providers',
+        );
+        const isUsingGooglePay =
+            isGooglePayBillingAddressEditingEnabled && (checkout && checkout.payments
+                ? checkout.payments.some((payment) => (payment?.providerId || '').startsWith('googlepay'))
+                : false);
+
+        if (isUsingGooglePay) {
+            return {
+                type: CheckoutStepType.Billing,
+                isActive: false,
+                isComplete: hasAddress,
+                isEditable: hasAddress,
                 isRequired: true,
             };
         }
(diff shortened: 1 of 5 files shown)


packages/core/src/app/checkout/getCheckoutStepStatuses.ts
```diff
@@ -124,6 +125,15 @@ const getBillingStepStatus = createSelector(
...
         }
 
+        const isGooglePayBillingAddressEditingEnabled = isExperimentEnabled(
+            config?.checkoutSettings,
+            'experiment name will be added here after adding it in LD',
+        );
+        const isUsingGooglePay =
+            (checkout && checkout.payments
+                ? checkout.payments.some((payment) => (payment?.providerId || '').startsWith('googlepay'))
+                : false) && isGooglePayBillingAddressEditingEnabled;
```
bc-nick, line 135: Would be better to check `isGooglePayBillingAddressEditingEnabled` first in this statement

`const isUsingGooglePay = isGooglePayBillingAddressEditingEnabled && ...`

bc-dronov: Done

bc-nick (approved):

packages/core/src/app/checkout/getCheckoutStepStatuses.ts
```diff
@@ -124,6 +125,15 @@ const getBillingStepStatus = createSelector(
             };
         }
 
+        const isGooglePayBillingAddressEditingEnabled = isExperimentEnabled(
+            config?.checkoutSettings,
+            'STRIPE-546.allow_billing_address_editing_for_all_Google_Pay_providers',
+        );
+        const isUsingGooglePay =
+            isGooglePayBillingAddressEditingEnabled && (checkout && checkout.payments
+                ? checkout.payments.some((payment) => (payment?.providerId || '').startsWith('googlepay'))
```
PavlenkoM, line 134: Are you sure that `.startsWith('googlepay')` is enough and we don't have providers with `providerId` for GPay that starts from some other prefix?

PavlenkoM (commented):
I think will be good to cover this change with unit test.

packages/core/src/app/checkout/getCheckoutStepStatuses.ts
bc-dronov: I've checked our Google pay provider names, and it looks like all this providers starts with `googlepay`.
Screenshot 2024-12-11 at 16 11 45

PavlenkoM (approved):

packages/core/src/app/checkout/getCheckoutStepStatuses.ts
```diff
@@ -139,7 +149,7 @@ const getBillingStepStatus = createSelector(
                             .includes(payment.providerId))
                 : false;
 
-        if (isUsingPaypal) {
+        if (isUsingPaypal || isUsingGooglePay) {
```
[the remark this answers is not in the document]
bc-peng: Just to simplify the logic, can we do early return like L109 for Gpay?
```
if (isUsingAmazonPay) {
```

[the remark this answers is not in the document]
bc-dronov: @bc-peng updated

richapatel1510 (approved):

bc-peng (approved):

PR #2118, opened 2024-12-09
feat(shipping): CHECKOUT-8861 Handle promotional item in multishipping

richapatel1510:
## What?

- Do not allow multiple shipping if the cart contains promotional items. Show the multi shipping unavailable modal.
- Switch to single shipping if customer edits cart after creating multi shipping consignments and adds a promotional item to the cart.

## Why?
This is because all free items are automatically assigned to the first address. This inflexibility in the underlying consignments api leads to a sub-optimal UX which results to a perception of poor quality.

## Testing / Proof

- CI checks
- Manual testing

https://github.com/user-attachments/assets/f639071a-aae2-418a-b08d-4b58f1629463

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/shipping/Shipping.tsx
+++ b/packages/core/src/app/shipping/Shipping.tsx
@@ -31,6 +31,7 @@ import { PaymentMethodId } from '../payment/paymentMethod';
 import { UnassignItemError } from './errors';
 import getShippableItemsCount from './getShippableItemsCount';
 import getShippingMethodId from './getShippingMethodId';
+import hasPromotionalItems from './hasPromotionalItems';
 import { MultiShippingFormValues } from './MultiShippingForm';
 import ShippingForm from './ShippingForm';
 import ShippingHeader from './ShippingHeader';
@@ -53,6 +54,7 @@ export interface ShippingProps {
 export interface WithCheckoutShippingProps {
     billingAddress?: Address;
     cart: Cart;
+    cartHasPromotionalItems: boolean;
     consignments: Consignment[];
     countries: Country[];
     countriesWithAutocomplete: string[];
@@ -110,10 +112,17 @@ class Shipping extends Component<ShippingProps & WithCheckoutShippingProps, Ship
             loadShippingOptions,
             onReady = noop,
             onUnhandledError = noop,
+            cartHasPromotionalItems,
+            isMultiShippingMode,
+            isNewMultiShippingUIEnabled,
         } = this.props;
 
         try {
             await Promise.all([loadShippingAddressFields(), loadShippingOptions(), loadBillingAddressFields()]);
+
+            if(cartHasPromotionalItems && isMultiShippingMode && isNewMultiShippingUIEnabled) {
+                await this.handleMultiShippingModeSwitch();
+            }
 
             onReady();
         } catch (error) {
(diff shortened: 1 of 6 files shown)


bc-mackxu:
@bc-charlesho I understand the reasoning behind blocking shoppers from creating multi-shipping when their cart has a free item. However, the old UI allows this, while the new UI restricts it. This inconsistency could confuse shoppers and lead to a poor user experience. Do we know how many carts with free items are currently creating multi-shipping?

bc-charlesho:
Mack, less than 3% of manual orders have multi ship. I don't know how many have free items but I'd assume its be a very small amount.

This is one of the reasons contributing to the confusing UX on the checkout's multi-ship feature. Therefore, we have decided to take it out due to low usage.

animesh1987:
@bc-charlesho this PR is for storefront checkout and not manual orders.

packages/core/src/app/shipping/Shipping.tsx
```diff
@@ -109,11 +111,18 @@ class Shipping extends Component<ShippingProps & WithCheckoutShippingProps, Ship
...
+            isMultiShippingMode,
+            isNewMultiShippingUIEnabled,
         } = this.props;
 
         try {
             await Promise.all([loadShippingAddressFields(), loadShippingOptions(), loadBillingAddressFields()]);
 
+            if(cartHasPromotionalItems && isMultiShippingMode && isNewMultiShippingUIEnabled) {
+                await this.handleMultiShippingModeSwitch();
+            }
```
bc-peng, line 124: Moving forward, we should avoid putting heavy API calls function inside a `useEffect` hook or similar places like this for better testability.
Instead, it would be more effective to put this function call on the button that triggers it.
This will simplify our testing, as we can focus on testing the button itself rather than preparing the entire component for testing an API-related functionality.
cc @animesh1987

richapatel1510: The button that triggers this is going to be on Cart page, hence that cannot be controlled.

bc-peng (approved):

animesh1987 (approved):

animesh1987:
PR looks good, but I have a question for @bc-charlesho around product experience. Shall we just clear all consignments with a warning dialog, if the shopper has made progress in multi shipping and click on the `Edit cart` link. 

This way we can avoid the concern raised [here](https://github.com/bigcommerce/checkout-js/pull/2118#discussion_r1877058903) by @bc-peng.

bc-charlesho:
Clearing all consignments when clicking on edit. The problem with that is the shopper may not add an item to the cart that is part of a promotion with a free item.

bc-mackxu:
The PR looks good, but I have a concern about the user experience. In the new manual order UI, users can currently switch back to the old UI if they encounter issues, such as unavailable features in the new version. But shopper cannot switch back the old one in the SF checkout page once we rollout the new UI.

Maybe we could consider to implement a more cautious rollout plan. WDYT @bc-charlesho @animesh1987

bc-charlesho:
We have multi-shipping as an option in Checkout Settings that merchants can toggle off if they encounter issues. Another thing we can do is a slow rollout. However, we don't have an opt-in on the checkout UI itself.

richapatel1510:
Added information modal on checkout load when user was already in multi shipping and created few consignments, then goes back to edit cart and adds an item with promotion


https://github.com/user-attachments/assets/f70a1931-9681-4152-8320-cdb08f9c3290

bc-mackxu (approved):

animesh1987 (approved):

PR #2120, opened 2024-12-09
feat(payment): PAYPAL-4692 fixed credit card loading glitch

andriiVitvitskyi1990:
## What?
Fixed credit card loading glitch

## Why?
To fix bug

## Testing / Proof
**Before**

https://github.com/user-attachments/assets/25a03563-93a7-4bd9-8109-639b1d459c8e

**After**

https://github.com/user-attachments/assets/da30da82-e4e3-42ec-b3bc-fbfbf73ec09f

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/payment/paymentMethod/CreditCardPaymentMethod.spec.tsx
+++ b/packages/core/src/app/payment/paymentMethod/CreditCardPaymentMethod.spec.tsx
@@ -196,16 +196,31 @@ describe('CreditCardPaymentMethod', () => {
         expect(defaultProps.deinitializePayment).toHaveBeenCalled();
     });
 
-    it('renders loading overlay while waiting for method to initialize', () => {
+    it('renders loading overlay while waiting for method to initialize', async () => {
         let component: ReactWrapper;
+        const mockInitializePayment = jest.fn().mockImplementation(() => {
+            return new Promise((resolve) => resolve(true));
+        });
 
         component = mount(<CreditCardPaymentMethodTest {...defaultProps} isInitializing />);
 
         expect(component.find(LoadingOverlay).prop('isLoading')).toBe(true);
 
-        component = mount(<CreditCardPaymentMethodTest {...defaultProps} />);
+        component = mount(
+            <CreditCardPaymentMethodTest
+                {...defaultProps}
+                initializePayment={mockInitializePayment}
+                isInitializing={false}
+            />
+        );
 
-        expect(component.find(LoadingOverlay).prop('isLoading')).toBe(false);
+        await new Promise((resolve) => process.nextTick(resolve));
+
+        component.update();
+
+        const loadingOverlay = component.find(LoadingOverlay);
+        expect(loadingOverlay.prop('isLoading')).toBe(false);
+
     });
 
     it('renders credit card fieldset', () => {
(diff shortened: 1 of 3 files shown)


packages/core/src/app/payment/paymentMethod/CreditCardPaymentMethod.spec.tsx
```diff
@@ -196,16 +196,29 @@ describe('CreditCardPaymentMethod', () => {
...
+        component = mount(
+            <CreditCardPaymentMethodTest
+                {...defaultProps}
+                initializePayment={mockInitializePayment}
+                isInitializing={false}
+            />
+        );
+
+        const loadingOverlay = component.find(LoadingOverlay);
+        setTimeout(() => {
```
bc-nick, line 218: Did you try how it will be working with `process.nextTick()`?

andriiVitvitskyi1990: It did not work, the problem is we need to wait until initializePayment finish and then check loading overlay, but in test we have updated isLoading prop after we get LoadingOverlay component because of initializePayment async. So i've used setTimeout to rearrange call stack

bc-nick, line 218: Thanks for explanation. Than you can try to do something like this. It should be work because after `nextTick` we update component and try checking the loadingOverlay's `isLoading` prop with a new updated value

```suggestion
        await new Promise((resolve) => process.nextTick(resolve));

        component.update();

        const loadingOverlay = component.find(LoadingOverlay);
        
        expect(loadingOverlay.prop('isLoading')).toBe(false);
```

andriiVitvitskyi1990: fixed

bc-nick (approved):

serhii-tkachenko (approved):

packages/core/src/app/payment/paymentMethod/CreditCardPaymentMethod.tsx
```diff
@@ -89,6 +90,7 @@ class CreditCardPaymentMethod extends Component<
 > {
     state: CreditCardPaymentMethodState = {
         isAddingNewCard: false,
+        isPreloaderOn: true,
```
animesh1987, line 93: This component is used by a quite a few providers, have you tested if the other providers work fine?

andriiVitvitskyi1990: This state sets to false whenever any of providers finish initialization, so should work fine for any provider that uses this component. I've checked PaypalCommerce and Braintree

animesh1987: Can you please test QP access and Hosted credit card as well. Thanks.

andriiVitvitskyi1990: What do you mean by WP access? Could you explain please

andriiVitvitskyi1990: @animesh1987 I've tested PaypalCommerce, Braintree and regular Credit Card with no provider onboarded

animesh1987 (approved):

PR #2123, opened 2024-12-11
fix(checkout): CHECKOUT-8824 Fix Error Message for First Consignment

bc-peng:
## What?
Fix the issue that when the first multi-shipping consignment is not done, the error message is missing.

## Why?
Improve UX.

## Testing / Proof
https://github.com/user-attachments/assets/1b734d38-86cb-49cf-b37a-f07424bcc647

### DIFF
--- a/packages/core/src/app/shipping/MultiShippingFormV2.test.tsx
+++ b/packages/core/src/app/shipping/MultiShippingFormV2.test.tsx
@@ -79,6 +79,38 @@ describe('MultiShippingFormV2 Component', () => {
         });
     });
 
+    it('shows error when creating a new destination if destination 1 is incomplete.', async () => {
+        jest.spyOn(checkoutState.data, 'getConsignments').mockReturnValue([]);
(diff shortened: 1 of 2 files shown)


packages/core/src/app/shipping/MultiShippingFormV2.test.tsx
```diff
@@ -79,6 +79,38 @@ describe('MultiShippingFormV2 Component', () => {
...
+            <CheckoutProvider checkoutService={checkoutService}>
+                <LocaleContext.Provider value={localeContext}>
+                    <ExtensionProvider checkoutService={checkoutService} errorLogger={errorLogger} >
+                        <MultiShippingFormV2 {...defaultProps} />
+                    </ExtensionProvider>
+                </LocaleContext.Provider>
+            </CheckoutProvider>,
+        );
+
+        await userEvent.click(screen.getByRole('button', { name: 'Add new destination' }));
```
richapatel1510, line 99: Shall we also get this text (`Add new destination`) from `localeContext.language`

richapatel1510 (approved):

PR #2144, opened 2025-01-16
fix(payment): PAYPAL-4987 fixed credit card duplication when initialize BT LPM

andriiVitvitskyi1990:
## What?
Fixed credit card duplication when initialize BT LPM

## Why?
To avoid duplication of credit card 

## Testing / Proof

https://github.com/user-attachments/assets/a105c222-f67e-47ef-8be5-3e1f0a691bb1

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/payment/Payment.spec.tsx
+++ b/packages/core/src/app/payment/Payment.spec.tsx
@@ -206,6 +206,29 @@ describe('Payment', () => {
         );
     });
 
+    it('does not render methods with braintreelocalmethods id', async () => {
+        const expectedPaymentMethods = paymentMethods.filter(
+            (method) => method.id !== PaymentMethodId.BraintreeLocalPaymentMethod,
+        );
+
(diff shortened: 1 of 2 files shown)


serhii-tkachenko (approved):
LGTM 👍

packages/core/src/app/payment/Payment.spec.tsx
```diff
@@ -206,6 +206,29 @@ describe('Payment', () => {
         );
     });
 
+    it('does not render methods with braintreelocalmethods id', async () => {
+        const expectedPaymentMethods = paymentMethods.filter(
+            (method) => method.id !== PaymentMethodId.BraintreeLocalPaymentMethod,
+        );
+
+        paymentMethods[3] = {
```
bc-nick, line 214: I do not think that is the right mock

bc-nick (approved):

PR #2148, opened 2025-01-21
feat(extension): CHECKOUT-8964 Introduce GetConsignment Command

bc-peng:
## What?
Handle a new checkout extension command, `GetConsignments`

## Depends on
https://github.com/bigcommerce/checkout-sdk-js/pull/2766
* Will include the SDK bump in this PR.

## Why?
To pass consignment data to a checkout extension.

## Testing / Proof
Screenshot 2025-01-21 at 11 17 42 AM

### DIFF
--- a/packages/checkout-extension/src/ExtensionService.test.tsx
+++ b/packages/checkout-extension/src/ExtensionService.test.tsx
@@ -131,7 +131,7 @@ describe('ExtensionService', () => {
 
         extensionService.removeListeners(ExtensionRegion.ShippingShippingAddressFormBefore);
 
-        expect(remover).toBeCalledTimes(3);
+        expect(remover).toBeCalledTimes(4);
         expect(checkoutService.clearExtensionCache).toHaveBeenCalled();
     });
 
--- a/packages/checkout-extension/src/handlers/createGetConsignmentHandler.ts
+++ b/packages/checkout-extension/src/handlers/createGetConsignmentHandler.ts
@@ -1,0 +1,22 @@
+import { ExtensionCommandType, ExtensionMessageType } from '@bigcommerce/checkout-sdk';
+
+import { CommandHandler, HandlerProps } from './CommandHandler';
+
+export function createGetConsignmentHandler({
+    checkoutService,
+    extension,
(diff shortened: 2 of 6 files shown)


packages/checkout-extension/src/handlers/createGetConsignmentHandler.ts
```diff
@@ -0,0 +1,22 @@
+import { ExtensionCommandType, ExtensionMessageType } from '@bigcommerce/checkout-sdk';
+
+import { CommandHandler, HandlerProps } from './CommandHandler';
+
+export function createGetConsignmentHandler({
```
animesh1987, line 5: Do we need this handler? Since I envision this to be called from extension lib to SDK?

bc-peng: If it's a command based communication, we need a command handler on the `checkout-js` side.

packages/checkout-extension/src/handlers/handlers.test.ts
```diff
@@ -34,9 +38,6 @@ describe('Handlers', () => {
 
             handler.handler({
                 type: ExtensionCommandType.ReloadCheckout,
-                payload: {
-                    extensionId: '123',
-                },
```
bc-peng, line 39: This is incorrect. `extensionId` optionally appears in `context` of every message, not in `payload`.
```
context: {
                    extensionId: '123',
                },
```

packages/checkout-extension/src/ExtensionService.test.tsx
```diff
@@ -131,7 +131,7 @@ describe('ExtensionService', () => {
 
         extensionService.removeListeners(ExtensionRegion.ShippingShippingAddressFormBefore);
 
-        expect(remover).toBeCalledTimes(3);
+        expect(remover).toBeCalledTimes(4);
```
bc-peng, line 134: One more handler leads to one more remover when unmounting an extension.

richapatel1510 (approved):

animesh1987 (approved):

PR #2149, opened 2025-01-22
feat(checkout): CHECKOUT-8959 Enable Shipping Form Reinitialization

bc-peng:
## What?
Enable the shipping address form to reinitialize itself to reflect the current checkout data state.

## Formik Note
- `initialValues` not available to the higher-order component, use `mapPropsToValues` instead. ([doc](https://formik.org/docs/api/formik#initialvalues-values))
- Setting `enableReinitialize` to `true` allows Formik to reset the form to the values defined in `mapPropsToValues` whenever any prop value of the nested component changes. (not documented in the Formik doc)

## Why?
To support use cases where a checkout extension updates data externally, the checkout UI can automatically display any changes after reloading the checkout data.

## Testing / Proof
Changing the shipping option on the right and then clicking the emoji on the left triggers a checkout reload, causing the shipping option extension to re-render.

https://github.com/user-attachments/assets/33ff651a-aeec-4046-aed6-9ce78dab0c43

### DIFF
--- a/packages/test-framework/src/fixture/pageObject/Checkout.ts
+++ b/packages/test-framework/src/fixture/pageObject/Checkout.ts
@@ -170,10 +170,6 @@ export class Checkout {
             addressAppendText,
         });
 
-        if (!isBillingAddressSame) await this.page.locator('label[for="sameAsBilling"]').click();
-        if (!shouldSaveAddress)
-            await this.page.locator('label[for="shippingAddress.shouldSaveAddress"]').click();
-
         await this.page.locator('#checkout-shipping-options').waitFor({ state: 'visible' });
 
         const firstShippingMethod = this.page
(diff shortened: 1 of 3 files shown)


packages/test-framework/src/fixture/pageObject/Checkout.ts
```diff
@@ -170,17 +170,22 @@ export class Checkout {
...
         await firstShippingMethod.click();
+
+        if (!shouldSaveAddress) {
+            await this.page.getByText('Save this address in my address book.').click();
+        }
+
+        if (!isBillingAddressSame) {
+            await this.page.locator('label[for="sameAsBilling"]').click();
+        }
+
```
bc-peng, line 188: This may be a side effect related to the changes.
The previous method of selecting checkboxes under the address form is no longer working.
However, simply moving the actions to the end resolves the problem.

bc-peng:
Closing the PR as we have decided to utilize the extension context to trigger a rerendering of the shipping form.

PR #2150, opened 2025-01-22
test(checkout): CHECKOUT-8926 add RTL tests for guest customer step use cases

richapatel1510:
## What?
Write unit tests to cover essential use cases of the customer step in the checkout process for guest shopper using RTL.

## Why?
In order to be able to move all the enzyme tests to RTL which enables us to move to React 18

## Testing / Proof

- CI checks

@bigcommerce/team-checkout


packages/core/src/app/customer/Customer.test.tsx
```diff
@@ -0,0 +1,128 @@
...
+    });
+
+    beforeEach(() => {
+        window.scrollTo = jest.fn();
+
+        checkoutService = createCheckoutService();
+        embeddedMessengerMock = createEmbeddedCheckoutMessenger({
+            parentOrigin: 'https://store.url',
+        });
+        analyticsTracker = {
```
bc-peng, line 46: We definitely want to avoid writing this and `defaultProps` in most files.
Maybe we can consider putting in the test framework.

richapatel1510: Makes sense. Can create a separate PR for this refactor.

packages/core/src/app/customer/CustomerGuest.test.tsx
```diff
@@ -0,0 +1,215 @@
...
+    let localeContext: LocaleContextType;
+    let checkout: Checkout;
+    let cart: Cart;
+    let config: StoreConfig;
+    let customer: CustomerData;
+    const defaultProps = {
+        isSubscribed: false,
+        isWalletButtonsOnTop: false,
+        onSubscribeToNewsletter: jest.fn(),
+        step: {} as CheckoutStepStatus,
```
animesh1987, line 41: Can we avoid using `as` here? Maybe better to create mocks for step?

animesh1987 (approved):

bc-peng (approved):

PR #2151, opened 2025-01-24
feat(payment): PI-3099 [AdyenV3] Remove iDEAL bank selection dropdown

bc-ania:
## What?
Remove iDEAL bank selection dropdown for Adyen under the experiment `PI-2456.adyen_enable_ideal_hosted_page`

## Why?
iDEAL have updated their central technical infrastructure.
It requires from us to:

remove the list of banks from their checkout pages.
redirect shoppers to the central iDEAL payment page.
The change is in the experimental phase, as Adyen plans to do so on April 1, [according to official documentation](https://docs.adyen.com/payment-methods/ideal/).

## Testing / Proof
before:

https://github.com/user-attachments/assets/a824a2ef-b43b-41c4-a158-d124d8fa4b7e

after:
https://github.com/user-attachments/assets/30b901f6-dd8c-4cef-aa16-9cbb8b2ebd63


vitalii-koshovyi (changes requested):
I believe we should update Adyen-web package version, they have changed implementation of the Ideal on their side.
Why do we need an experiment for it?

PR #2154, opened 2025-01-29
feat(payment): PAYPAL-4324 Added experiment for Apple Pay

bc-nick:
## What?

Added experiment for switching Apple Pay implementation
Fixed test related to apple pay on Customer Step
Updated checkout-sdk-js

## Why?

For supporting third party browsers

## Testing / Proof

Local testing, take a look here https://github.com/bigcommerce/checkout-js/pull/2154

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/checkout/Checkout.tsx
+++ b/packages/core/src/app/checkout/Checkout.tsx
@@ -205,9 +205,9 @@ class Checkout extends Component<
             }), extensionService.loadExtensions()]);
 
             const providers = data.getConfig()?.checkoutSettings?.remoteCheckoutProviders || [];
-            const features = data.getConfig()?.checkoutSettings?.features || {};
+            const checkoutSettings = data.getConfig()?.checkoutSettings;
 
-            const supportedProviders = getSupportedMethodIds(providers, features);
+            const supportedProviders = getSupportedMethodIds(providers, checkoutSettings);
 
             if (providers.length > 0) {
                 const configs = await loadPaymentMethodByIds(supportedProviders);
--- a/packages/core/src/app/checkout/Checkout.tsx
+++ b/packages/core/src/app/checkout/Checkout.tsx
@@ -205,7 +205,9 @@ class Checkout extends Component<
             }), extensionService.loadExtensions()]);
 
             const providers = data.getConfig()?.checkoutSettings?.remoteCheckoutProviders || [];
-            const supportedProviders = getSupportedMethodIds(providers);
+            const features = data.getConfig()?.checkoutSettings?.features || {};
+
+            const supportedProviders = getSupportedMethodIds(providers, features);
 
             if (providers.length > 0) {
                 const configs = await loadPaymentMethodByIds(supportedProviders);
--- a/packages/core/src/app/common/utility/should-filter-apple-pay.ts
+++ b/packages/core/src/app/common/utility/should-filter-apple-pay.ts
@@ -1,5 +1,9 @@
 import { isApplePayWindow } from './';
 
-export default function shouldFilterApplePay(methodId: string, isBrowserSupport: boolean): boolean {
-    return Boolean(methodId === 'applepay' && !isApplePayWindow(window) && !isBrowserSupport);
+export default function shouldFilterApplePay(methodId: string, isBrowserSupported: boolean): boolean {
+    if (methodId !== 'applepay') {
+        return false;
+    }
+
+    return !isApplePayWindow(window) && !isBrowserSupported;
 }
--- a/packages/core/src/app/common/utility/should-filter-apple-pay.ts
+++ b/packages/core/src/app/common/utility/should-filter-apple-pay.ts
@@ -1,0 +1,5 @@
+import { isApplePayWindow } from './';
+
+export default function shouldFilterApplePay(methodId: string, isBrowserSupport: boolean): boolean {
+    return Boolean(methodId === 'applepay' && !isApplePayWindow(window) && !isBrowserSupport);
+}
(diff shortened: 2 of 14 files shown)


andriiVitvitskyi1990 (approved):

serhii-tkachenko (approved):

vincent89291 (approved):

packages/core/src/app/checkout/Checkout.tsx
```diff
@@ -205,7 +205,9 @@ class Checkout extends Component<
             }), extensionService.loadExtensions()]);
 
             const providers = data.getConfig()?.checkoutSettings?.remoteCheckoutProviders || [];
-            const supportedProviders = getSupportedMethodIds(providers);
+            const features = data.getConfig()?.checkoutSettings?.features || {};
+
+            const supportedProviders = getSupportedMethodIds(providers, features);
```
animesh1987, line 210: Can we use this here https://github.com/bigcommerce/checkout-js/blob/master/packages/core/src/app/common/utility/isExperimentEnabled.ts#L3

packages/core/src/app/customer/getSupportedMethods.ts
```diff
@@ -28,9 +31,10 @@ export const SUPPORTED_METHODS: string[] = [
     'googlepaytdonlinemart',
 ];
 
-export const getSupportedMethodIds = (methodIds: string[]): string[] => {
+export const getSupportedMethodIds = (methodIds: string[], features: CheckoutSettings['features'] = {}): string[] => {
     return methodIds.filter((methodId) => {
-        if (methodId === APPLE_PAY && !isApplePayWindow(window)) {
+        // TODO: this check have to be deleted after implementation of Apple Pay for third party browsers will be tested and released
+        if (methodId === APPLE_PAY && !isApplePayWindow(window) && !features['PAYPAL-4324.applepay_web_browser_support']) {
```
animesh1987, line 37: My suggestion would be to move the logic of `'PAYPAL-4324.applepay_web_browser_support'` in `isApplePayWindow` method so eventually when this experiment is rolled out we can remove the whole method?

bc-nick: This method will be removed in any case when the exp is rolled out, because we use it only for filtering apple pay method. I do not see a valid reason to move this check to `isApplePayWindow` in this case

packages/core/src/app/checkout/Checkout.tsx
bc-nick: This function return `true` if an exp is `undefined` by default. It is not what we are expected.

packages/core/src/app/customer/getSupportedMethods.ts
animesh1987: It will be a cleaner and more encapsulated approach where all logic lives in one place.

packages/core/src/app/checkout/Checkout.tsx
animesh1987: This is because we have had issues when an experiment is rolled out completely, say in a few weeks and a merchant forks checkout-js to create custom checkout in the meantime it causes us to never be able to remove the experiment. 
Hence this is the new approach to roll out experiments so we can successfully remove experiments.

packages/core/src/app/customer/getSupportedMethods.ts
bc-nick: Are you suggesting to pass the features to `isApplePayWindow` and handle this logic there?

packages/core/src/app/checkout/Checkout.tsx
bc-nick: Thanks, that is a good point. I will keep it in mind and add changes related to it.

packages/core/src/app/customer/getSupportedMethods.ts
bc-nick: Why am i asking it is because i do not understand what is difference between checking this exp in current place and checking inside isApplePayWindow if eventually this function and this `getSupportedMethods` method will be deleted, except we're gonna do another nesting?

packages/core/src/app/customer/getSupportedMethods.ts
```diff
@@ -34,7 +35,7 @@ export const SUPPORTED_METHODS: string[] = [
 export const getSupportedMethodIds = (methodIds: string[], features: CheckoutSettings['features'] = {}): string[] => {
     return methodIds.filter((methodId) => {
         // TODO: this check have to be deleted after implementation of Apple Pay for third party browsers will be tested and released
-        if (methodId === APPLE_PAY && !isApplePayWindow(window) && !(features['PAYPAL-4324.applepay_web_browser_support'] ?? true)) {
+        if (shouldFilterApplePay(methodId, features['PAYPAL-4324.applepay_web_browser_support'] ?? true)) {
```
animesh1987, line 38: Can we use the `isExperimentEnabled` utility here please.

packages/core/src/app/common/utility/should-filter-apple-pay.ts
```diff
@@ -0,0 +1,5 @@
+import { isApplePayWindow } from './';
+
+export default function shouldFilterApplePay(methodId: string, isBrowserSupport: boolean): boolean {
+    return Boolean(methodId === 'applepay' && !isApplePayWindow(window) && !isBrowserSupport);
```
animesh1987, line 4: This already returns a Boolean so wrapping in boolean is unnecessary. Can be cleaner if we can do 

```
if (methodId !== 'applepay') {
        return false;
    }
    
 if (isApplePayWindow(window) && isBrowserSupported) {
     return false;
 }
 
 return true;
```

animesh1987 (approved):

PR #2157, opened 2025-01-31
feat(extension): CHECKOUT-8974 Introduce Query Handler

bc-peng:
## What?
Introduce a new `query` concept that allows an extension to fetch any data from the checkout object by using a simple query key. 

## Depends on
https://github.com/bigcommerce/checkout-sdk-js/pull/2769

## Why?
An extension can fetch checkout data immediately.

## Testing / Proof
After the extension service initialization, clicking the emoji will run:
```
console.log(await service.getConsignments(useCache));
```
`useCache` is set to `true` at first. On the second time of clicking the emoji, `useCache` will be `false`, therefore, a `GET` checkout call is made.

https://github.com/user-attachments/assets/8e87cdfb-103c-4338-bf1b-73bf2f1d2e28

@bigcommerce/team-checkout @bigcommerce/team-payments

### DIFF
--- a/packages/checkout-extension/src/ExtensionService.ts
+++ b/packages/checkout-extension/src/ExtensionService.ts
@@ -2,6 +2,7 @@ import {
     CheckoutService,
     Extension,
     ExtensionCommandMap,
+    ExtensionQueryMap,
     ExtensionRegion,
 } from '@bigcommerce/checkout-sdk';
 import React from 'react';
@@ -9,8 +10,9 @@ import React from 'react';
 import { ErrorLevelType, ErrorLogger } from '@bigcommerce/checkout/error-handling-utils';
 
 import { ExtensionAction } from './ExtensionProvider';
-import * as handlerFactories from './handlers';
-import { CommandHandler } from './handlers/CommandHandler';
+import { CommandHandler, QueryHandler } from './handler';
+import * as commandHandlerFactories from './handler/commandHandlers';
+import * as queryHandlerFactories from './handler/queryHandlers';
 
 export class ExtensionService {
     private handlers: { [extensionId: string]: Array<() => void> } = {};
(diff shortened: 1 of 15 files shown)


packages/checkout-extension/src/ExtensionService.ts
```diff
@@ -129,6 +131,21 @@ export class ExtensionService {
...
+
+        Object.values(queryHandlerFactories).forEach((createQueryHandlerFactory) => {
+            const handlerFactory = createQueryHandlerFactory(handlerProps);
+
+            if (this.isQueryHandler(handlerFactory.queryType, handlerFactory)) {
+                this.handlers[extension.id].push(
+                    this.checkoutService.handleExtensionQuery(
+                        extension.id,
+                        handlerFactory.queryType,
+                        // eslint-disable-next-line @typescript-eslint/no-misused-promises
```
bc-peng, line 143: Here should be updated as returning `Promise<void>`.
https://github.com/bigcommerce/checkout-sdk-js/blob/9c4d9b738484cc4eca9506a3dbac720b6ca75e22/packages/core/src/checkout/checkout-service.ts#L1488

packages/checkout-extension/src/ExtensionService.ts
```diff
@@ -129,6 +131,21 @@ export class ExtensionService {
...
+                        handlerFactory.handler,
+                    ),
```
bc-peng, line 145: We currently define query handler the same way as the command handler.
```
handler: (command: ExtensionQueryMap[T]) => void,
```
However, due to the nature of making API calls for a query, we may consider change it to below.
```
handler: (command: ExtensionQueryMap[T]) => Promise<void>,
```
```

File: packages/checkout-extension/src/handler/queryHandlers/createGetConsignmentHandler.ts [hunk 3]
```diff
@@ -1,14 +1,18 @@
...
     extension,
-}: HandlerProps): CommandHandler<ExtensionCommandType.GetConsignments> {
+}: QueryHandlerProps): QueryHandler<ExtensionQueryType.GetConsignments> {
     return {
-        commandType: ExtensionCommandType.GetConsignments,
-        handler: () => {
+        queryType: ExtensionQueryType.GetConsignments,
+        handler: async (data) => {
+            if (!data.payload?.useCache) {
+                await checkoutService.loadCheckout();
```
animesh1987, line 13: Could this cause side effects that might change the display of values on checkout page?

bc-peng: Yes, it will refresh the checkout state, hence updating the checkout page UI accordingly.
I feel it is in line with our existing logic, checkout page should be in sync with the remote data after an action.

animesh1987 (approved):

PR #2160, opened 2025-02-04
feat(extension): CHECKOUT-8959 Introduce ReRenderShippingForm Command

bc-peng:
## What?
Introduce a new checkout extension command.

## Why?
To support the force re-rendering of the shipping form.

## Out of the scope
Displaying an external change to the Google AutoComplete address field (address line one when enabled) is not supported, as its value is managed by the Downshift library.

## Testing / Proof

*Clicking on the emoji triggers the `ReRenderShippingForm` command.

### Update Shipping Form Fields
The update to the delivery date form field on another page is applied once the command is triggered.

https://github.com/user-attachments/assets/3113abbf-a662-4a22-bae5-c7f779266b75

### Update Selected Shipping Option

https://github.com/user-attachments/assets/063640fc-0a7d-4c8f-bbbe-a492476df4c0

### Multi-shipping
The update to the shipping option on another page is applied once the command is triggered.

https://github.com/user-attachments/assets/39d4f152-f549-41f3-854b-c53b8bfab41a

### DIFF
--- a/packages/checkout-extension/src/ExtensionProvider.tsx
+++ b/packages/checkout-extension/src/ExtensionProvider.tsx
@@ -9,15 +9,17 @@ import { ExtensionService } from './ExtensionService';
 
 export interface ExtensionState {
     isShowingLoadingIndicator: boolean;
+    shippingFormRenderTimestamp: undefined | number;
 }
 
 export interface ExtensionAction {
     type: ExtensionActionType;
-    payload: boolean;
+    payload: boolean | number;
 }
 
 export enum ExtensionActionType {
     SHOW_LOADING_INDICATOR,
+    RE_RENDER_SHIPPING_FORM,
 }
 
 export interface ExtensionProviderProps {
@@ -33,6 +35,7 @@ export const ExtensionProvider = ({
 }: ExtensionProviderProps) => {
     const [extensionState, dispatch] = useReducer(extensionReducer, {
         isShowingLoadingIndicator: false,
+        shippingFormRenderTimestamp: undefined,
     });
     const extensionService = new ExtensionService(checkoutService, dispatch, errorLogger);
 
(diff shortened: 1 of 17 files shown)


packages/core/src/app/shipping/shippingOption/ShippingOptionsForm.tsx
```diff
@@ -30,16 +44,22 @@ class ShippingOptionsForm extends PureComponent<
...
+            setValues,
         } = this.props;
 
         if (consignments?.length && shouldShowShippingOptions) {
             analyticsTracker.showShippingMethods();
         }
+
+        if (newShippingFormRenderTimestamp !== shippingFormRenderTimestamp) {
+            setValues(getShippingOptionIds(this.props));
+        }
```
bc-peng, line 62: I was unable to cover this with a unit test because I couldn't trigger `componentDidUpdate` in Enzyme.

packages/checkout-extension/src/ExtensionProvider.tsx
```diff
@@ -9,15 +9,17 @@ import { ExtensionService } from './ExtensionService';
 
 export interface ExtensionState {
     isShowingLoadingIndicator: boolean;
+    shippingFormRenderTimestamp: undefined | number;
```
animesh1987, line 12: Why do we need a timestamp?

bc-peng: We need a way to trigger Formik's `setValues` method to update Formik's internal context.
In my testing, the component does re-render when the checkout state is updated, however, Formik's context stays the same.
Therefore, using a timestamp to manually trigger two Formik's contexts update(one for address, one for shipping option).

animesh1987: Thanks for the explanation, must be due to the way we heavily optimise formik and caching. 
Also doesn't help this lib outdated as well 😢

animesh1987 (approved):

PR #2178, opened 2025-02-20
feat(shipping): CHECKOUT-8999 Bump checkout sdk

animesh1987:
## What?
Bump checkout sdk

## Why?
Release https://github.com/bigcommerce/checkout-sdk-js/pull/2794

## Testing / Proof
- CI

@bigcommerce/team-checkout

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "build": "nx run core:build",
     "build:server": "http-server dist",
     "dev": "nx run core:dev",
+    "dev:vm": "WEBPACK_DONE='rsync -rvc --progress --delete --compress-level=9 build store.bcdev:/opt/bigcommerce_app/vagrant_code/vendor/bower_components/checkout/' npm run dev",
     "dev:server": "http-server build --cors",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -14,6 +14,7 @@
     "build": "nx run core:build",
     "build:server": "http-server dist",
     "dev": "nx run core:dev",
+    "dev:local": "WEBPACK_DONE='rsync -rvc --progress --delete --compress-level=9 build store.bcdev:/opt/bigcommerce_app/vagrant_code/vendor/bower_components/checkout/' npm run dev",
```
animesh1987, line 17: Added this as a convenience method to run in local environment.

richapatel1510 (approved):

package.json
bc-peng: Perhaps we should avoid calling it `dev:local` because `local` typically indicates a local machine for most developers.
Maybe we could use `dev:vm` or something more relevant to our specific context.

animesh1987: I like `dev:vm`, will update it. Thanks.

PR #2180, opened 2025-02-21
test(checkout): CHECKOUT-9007 Guest Checkout Shipping Step Happy Path

bc-peng:
## What?
Add a test for the happy path of the Guest Checkout Shipping step.

Test path:
- Entering the shipping step as a guest.
- Filling in essential address details.
- Triggering two API calls: a POST call to create a consignment and a PUT call to select the default shipping option.
- Asserting that “billing address is the same as shipping” is checked.
- Clicking “Entering” goes to the payment step.

## Why?
Enable us to upgrade to react 18.

## Testing / Proof
CI.

### DIFF
--- a/packages/core/src/app/shipping/Shipping.test.tsx
+++ b/packages/core/src/app/shipping/Shipping.test.tsx
@@ -1,258 +1,181 @@
-import '@testing-library/jest-dom';
 import {
-    CheckoutSelectors,
     CheckoutService,
     createCheckoutService,
+    createEmbeddedCheckoutMessenger,
+    EmbeddedCheckoutMessenger,
 } from '@bigcommerce/checkout-sdk';
 import userEvent from '@testing-library/user-event';
+import { noop } from 'lodash';
 import React, { FunctionComponent } from 'react';
 
-import { ExtensionProvider } from '@bigcommerce/checkout/checkout-extension';
 import {
-    createLocaleContext,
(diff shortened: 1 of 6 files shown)


packages/core/src/app/shipping/Shipping.test.tsx
bc-peng, line 1: Renamed `Shipping.test.tsx` to `ShippingComponent.test.tsx` because the original file does not test e2e flow, while the testing files for major steps, such as `Customer.test.tsx`, should do.

packages/core/src/app/shipping/ShippingComponent.test.tsx
bc-peng, line 1: Old `Shipping.test.tsx`, nothing changed inside.

packages/test-framework/src/react-testing-library-support/CheckoutPageNodeObject.ts
```diff
@@ -206,4 +209,19 @@ export class CheckoutPageNodeObject {
     async waitForPaymentStep(): Promise<void> {
         await waitFor(() => screen.getByText(/place order/i));
     }
+
+    async fillShippingAddress(): Promise<void> {
```
bc-peng, line 213: Create a new method as it is needed for the following tests.

richapatel1510 (approved):

PR #2184, opened 2025-02-25
test(checkout): CHECKOUT-9030 Single Shipping for customer with no saved addresses

richapatel1510:
## What?
Add RTL test for single shipping for logged in customer with no saved addresses. 

**Test path:**

- Enter the shipping step as a logged in customer.
- Fill in essential address details in the form.
- Verify the `Save this address in my address book` checkbox is present and selected.
- Click “Continue” to proceed to the payment step.

## Why?
In order to be able to move all the enzyme tests to RTL which enables us to move to React 18

## Testing / Proof

- CI checks

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/shipping/Shipping.test.tsx
+++ b/packages/core/src/app/shipping/Shipping.test.tsx
@@ -157,7 +157,7 @@ describe('Shipping step', () => {
             );
             // eslint-disable-next-line jest-dom/prefer-to-have-attribute
             expect(
-                screen.getByTestId('billingSameAsShipping').hasAttribute('checked'),
+                screen.getByLabelText('My billing address is the same as my shipping address.').hasAttribute('checked'),
             ).toBeTruthy();
 
             checkout.updateCheckout(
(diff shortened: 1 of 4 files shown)


packages/core/src/app/shipping/Shipping.test.tsx
```diff
@@ -207,7 +207,7 @@ describe('Shipping step', () => {
             );
 
             await checkout.fillShippingAddress();
-            await userEvent.click(screen.getByTestId('billingSameAsShipping'));
+            await userEvent.click(screen.getByLabelText('My billing address is the same as my shipping address.'));
```
bc-peng, line 210: Once we finish covering the shipping step, is it possible to remove the testId from the component?

packages/test-framework/src/react-testing-library-support/mocks/checkout.mock.ts
```diff
@@ -158,6 +158,18 @@ const customer: Customer = {
...
+    id: 1,
+    isGuest: false,
+    email: 'user@example.com',
+    firstName: 'John',
+    lastName: 'Doe',
+    fullName: 'John Doe',
+    addresses: [],
+    storeCredit: 0,
+    shouldEncourageSignIn: true,
+};
```
bc-peng, line 171:
```suggestion
const customerWithoutSavedAddresses: Customer = {
    ...customer,
    addresses: [],
};
```

bc-peng (approved):

packages/core/src/app/shipping/Shipping.test.tsx
richapatel1510: Yes, I think we should be able to.

PR #2189, opened 2025-02-28
test(checkout): CHECKOUT-8932 Add tests for single shipping form for customer with saved addresses

richapatel1510:
## What?
Create Tests for Customer's Single Shipping Use Cases with Saved Addresses.

## Why?
In order to be able to move all the enzyme tests to RTL which enables us to move to React 18.

## Testing / Proof

- CI checks

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/shipping/Shipping.test.tsx
+++ b/packages/core/src/app/shipping/Shipping.test.tsx
@@ -416,7 +416,7 @@ describe('Shipping step', () => {
             expect(screen.getByText(payments[0].config.displayName)).toBeInTheDocument();
         });
 
-        it('selects the invalid customer address and completes the shipping step', async () => {
+        it('selects the invalid customer address, fills the address form and finally completes the shipping step', async () => {
             checkout.use(CheckoutPreset.CheckoutWithCustomerHavingInvalidAddress);
 
             const { container } = render(<CheckoutTest {...defaultProps} />);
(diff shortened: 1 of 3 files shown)


packages/core/src/app/shipping/Shipping.test.tsx
```diff
@@ -284,6 +286,221 @@ describe('Shipping step', () => {
...
+            );
+
+            await userEvent.type(
+                screen.getByRole('textbox', { name: /address/i }),
+                shippingAddress.address1,
+            );
+
+            expect(checkoutService.updateShippingAddress).toHaveBeenCalled();
+
+            // screen.debug(undefined, Infinity);
```
animesh1987, line 473: Is this intentional?

packages/core/src/app/shipping/Shipping.test.tsx
```diff
@@ -284,6 +286,221 @@ describe('Shipping step', () => {
...
+
+            await userEvent.click(screen.getByRole('button', { name: 'Continue' }));
+
+            await checkout.waitForPaymentStep();
+
+            expect(checkoutService.updateBillingAddress).toHaveBeenCalled();
+            expect(screen.getByText(payments[0].config.displayName)).toBeInTheDocument();
+        });
+
+        it('selects the invalid customer address and completes the shipping step', async () => {
```
animesh1987, line 419: Not sure if we need a better description here, as this feels like `user is able to complete shipping step with an invalid address` 🤔

animesh1987 (approved):

PR #2213, opened 2025-03-19
fix(checkout): CHECKOUT-9086 add available shipping option check

richapatel1510:
## What?
Fix: If checkout has selected shipping method, but the shipping option id is not in available shipping methods list, on the UI the there is no shipping method selected, but shopper can just ignore and continue to the next step.

## Why?
We should block shopper continue to the next step if selected shipping option is not present in the list of available shipping options.

## Testing / Proof

- CI checks
- Manual testing proof [(here)](https://github.com/bigcommerce/checkout-js/pull/2213#issuecomment-2742011710)

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/shipping/isSelectedShippingOptionValid.ts
+++ b/packages/core/src/app/shipping/isSelectedShippingOptionValid.ts
@@ -1,0 +1,17 @@
+import { Consignment } from '@bigcommerce/checkout-sdk';
+import { every } from 'lodash';
+
+export default function isSelectedShippingOptionValid(consignments: Consignment[]): boolean {
+    if (!consignments.length) {
+        return false;
+    }
+
+    return every(
+        consignments,
+        (consignment) =>
+        (consignment.availableShippingOptions &&
+            consignment.availableShippingOptions.filter(
+                ({ id }) => id === consignment.selectedShippingOption?.id,
+            ).length)
+    );
+}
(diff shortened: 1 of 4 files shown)


bc-shawnwang:
@richapatel1510 I tested entire flow, it works as expected:
- if selected shipping options (id) is not in available shipping options, then go to the shipping section, there is no shipping option selected, and cannot continue, until select a valid one. And it's no matter in the middle of the checkout, or submitting order failed and shopper is guided to the shipping section.
- reload checkout page, there is no error that could affect the shipping section.

https://github.com/user-attachments/assets/ae4b13d3-e3bd-43b1-9a4a-9fa1fc90591a

So the change looks good to me

packages/core/src/app/shipping/isSelectedShippingOptionValid.ts
```diff
@@ -0,0 +1,17 @@
...
+        return false;
+    }
+
+    return every(
+        consignments,
+        (consignment) =>
+        (consignment.availableShippingOptions &&
+            consignment.availableShippingOptions.filter(
+                ({ id }) => id === consignment.selectedShippingOption?.id,
+            ).length)
```
bc-peng, line 15: PR looks good.

Also, I can see this won’t be an issue for the custom shipping method where `availableShippingOptions` is always undefined, but the shipping step is not editable.

bc-peng (approved):

packages/core/src/app/shipping/isSelectedShippingOptionValid.ts
bc-mackxu, line 15: If selectedShippingOption is null, maybe don't need to loop the filter?
```suggestion
        (consignment.availableShippingOptions &&
         consignment.selectedShippingOption &&
            consignment.availableShippingOptions.filter(
                ({ id }) => id === consignment.selectedShippingOption.id,
            ).length)
```

richapatel1510: It will not come to this function if there is no `selectedShippingOption`, since that is already checked via `hasSelectedShippingOptions`

bc-mackxu (approved):

PR #2218, opened 2025-03-20
feat(checkout): CHECKOUT-8517 Display shipping discount on checkout page

bc-mackxu:
## What?  
Display auto shipping discounts in the shipping summary and order summary on the checkout page.  

## Why?  
To ensure users can see the auto shipping discount applied to their cart.

## CI passed after update SDK version
Note - Since the SDK [PR](https://github.com/bigcommerce/checkout-sdk-js/pull/2811) hasn't been merged yet, the CI is failing, but they are passed locally.
image

## Testing / Proof
- [x] unit test
- [x] manual test

Single consignment:
image

Multi consignment:
image

### DIFF
--- a/packages/core/src/app/cart/CartSummary.spec.tsx
+++ b/packages/core/src/app/cart/CartSummary.spec.tsx
@@ -6,7 +6,7 @@ import { ExtensionProvider } from '@bigcommerce/checkout/checkout-extension';
 import { createLocaleContext, LocaleContext, LocaleContextType } from '@bigcommerce/checkout/locale';
 import { CheckoutProvider } from '@bigcommerce/checkout/payment-integration-api';
 
-import { getCheckout } from '../checkout/checkouts.mock';
+import { getCheckout, getCheckoutWithShippingDiscounts } from '../checkout/checkouts.mock';
 import { getStoreConfig } from '../config/config.mock';
 import { getCustomer } from '../customer/customers.mock';
 import OrderSummary from '../order/OrderSummary';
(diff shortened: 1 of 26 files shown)


packages/core/src/app/cart/CartSummary.spec.tsx
bc-peng, line 1: We are going to delete this test file, can you make a test in `checkout.test.tsx`?

packages/core/src/app/shipping/shippingOption/StaticShippingOption.tsx
```diff
@@ -9,14 +9,35 @@ import './StaticShippingOption.scss';
...
+    const renderShippingPrice = () => {
+        if (shippingCostAfterDiscount && shippingCostAfterDiscount !== method.cost) {
+            return (
+                <>
+                    <span className="shippingOption-price-before-discount">
+                        <ShopperCurrency amount={method.cost} />
+                    </span>
+                    <ShopperCurrency amount={shippingCostAfterDiscount} />
+
+                </>
```
bc-peng, line 29:
```suggestion
                    <ShopperCurrency amount={shippingCostAfterDiscount} />
                </>
```

packages/core/src/app/cart/mapToOrderSummarySubtotalsProps.ts
```diff
@@ -14,15 +14,39 @@ export default function mapToOrderSummarySubtotalsProps({
     coupons,
     taxes,
     fees,
-}: Checkout): OrderSummarySubtotalsProps {
+}: Checkout,
+isShippingDiscountDisplayEnabled: boolean,
+): OrderSummarySubtotalsProps {
+    const allConsignmentsHaveSelectedShippingOption = hasSelectedShippingOptions(consignments);
+
+    if (!isShippingDiscountDisplayEnabled) {
```
bc-peng, line 22: Can we use control flow to determine the impacted values instead of using the condition to early return an almost same object?

packages/core/src/app/cart/CartSummary.spec.tsx
bc-mackxu: I remove the test here since I created a test in checkout.test.tsx already.

packages/core/src/app/cart/mapToOrderSummarySubtotalsProps.ts
```diff
@@ -14,15 +14,27 @@ export default function mapToOrderSummarySubtotalsProps({
...
+isShippingDiscountDisplayEnabled: boolean,
+): OrderSummarySubtotalsProps {
+    const allConsignmentsHaveSelectedShippingOption = hasSelectedShippingOptions(consignments);
+    const shippingAmount = isShippingDiscountDisplayEnabled
+        ? allConsignmentsHaveSelectedShippingOption
+            ? getShippingCostAfterAutomaticDiscount(shippingCostBeforeDiscount, consignments)
+            : undefined
+        : allConsignmentsHaveSelectedShippingOption
+            ? shippingCostBeforeDiscount
+            : undefined;
```
bc-peng, line 27: Can you create a function for this calculation?

bc-mackxu: Yes, the code may look duplicated, but I prefer to keep it separate. This makes it easier to remove the old code when cleaning up the experiment, reducing the risk of mistakes. WDYT?

bc-peng (approved):

richapatel1510 (approved):

PR #2221, opened 2025-03-21
fix(payment): STRIPE-696 Remove browser alert before Stripe UPE iDEAL redirect

PavlenkoM:
## What?
Remove browser confirmation alert before redirect to Stripe UPE iDEAL confirmation page.

## Why?
Browser alert is causing confusion and customer churn when using redirect-based Stripe UPE iDEAL APM.

## Testing / Proof
Before:

https://github.com/user-attachments/assets/95c55e57-8c80-4bc6-9c59-3752c7961a3b

After:

https://github.com/user-attachments/assets/872bf0be-7002-417e-8335-3758ebef6650

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/payment/Payment.tsx
+++ b/packages/core/src/app/payment/Payment.tsx
@@ -321,7 +321,7 @@ class Payment extends Component<
         const { defaultMethod, isSubmittingOrder, language } = this.props;
         const { selectedMethod = defaultMethod } = this.state;
 
-        // TODO: Perhaps there is a better way to handle `adyen`, `afterpay`, `amazonpay`,
+        // TODO: [PI-3551] Perhaps there is a better way to handle `adyen`, `afterpay`, `amazonpay`,
         // `checkout.com`, `converge`, `sagepay`, `stripev3` and `sezzle`. They require
         //  a redirection to another website during the payment flow but are not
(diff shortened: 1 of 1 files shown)


vitalii-koshovyi (approved):

serhii-tkachenko (approved):

packages/core/src/app/payment/Payment.tsx
```diff
@@ -352,7 +352,8 @@ class Payment extends Component<
             selectedMethod.gateway === PaymentMethodId.Clearpay ||
             selectedMethod.gateway === PaymentMethodId.Checkoutcom ||
             selectedMethod.gateway === PaymentMethodId.Mollie ||
-            selectedMethod.gateway === PaymentMethodId.StripeV3
+            selectedMethod.gateway === PaymentMethodId.StripeV3 ||
```
animesh1987, line 355: There is a TODO here https://github.com/bigcommerce/checkout-js/pull/2221/files#diff-01ca856011dd0cc1233e0450e9ece93e798fe8d07af017fa78240e27559f5dd7R324 which seems like will never get resolved. Is there a plan to address this at some point?

animesh1987 (approved):

PR #2225, opened 2025-03-25
chore(checkout): CHECKOUT-9080 Refactor test for gift certificate

animesh1987:
## What?
Refactor test for gift certificate

## Why?
As part of migration to RTL tests refactor tests.

## Testing / Proof
- CI

Screenshot 2025-03-25 at 4 58 49 pm

@bigcommerce/team-checkout


packages/core/src/app/giftCertificate/AppliedGiftCertificate.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+    const giftCertificate = getGiftCertificate();
+
+    it('renders correctly if gift certificate is applied', () => {
+        render(<AppliedGiftCertificate giftCertificate={giftCertificate} />);
+
+        const usedAmount = roundTo(giftCertificate.used * config.shopperCurrency.exchangeRate, parseInt(config.shopperCurrency.decimalPlaces, 10));
+        const remainingAmount = roundTo(giftCertificate.remaining * config.shopperCurrency.exchangeRate, parseInt(config.shopperCurrency.decimalPlaces, 10));
+
+        const usedAmountElement = screen.getByTestId('giftCertificate-amount');
+        const remainingAmountElement = screen.getByTestId('giftCertificate-remaining');
```
animesh1987, line 26: Have to use this since currency formatter is an internal implementation in SDK, so it's ok to check for text content which is the amount minus the currency symbol.

packages/core/src/app/giftCertificate/AppliedGiftCertificate.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+describe('AppliedGiftCertificate', () => {
+    const config = getStoreConfig();
+    const giftCertificate = getGiftCertificate();
+
+    it('renders correctly if gift certificate is applied', () => {
+        render(<AppliedGiftCertificate giftCertificate={giftCertificate} />);
+
+        const usedAmount = roundTo(giftCertificate.used * config.shopperCurrency.exchangeRate, parseInt(config.shopperCurrency.decimalPlaces, 10));
+        const remainingAmount = roundTo(giftCertificate.remaining * config.shopperCurrency.exchangeRate, parseInt(config.shopperCurrency.decimalPlaces, 10));
```
bc-peng, line 23: Could you elaborate on why we should not hardcode static values here?

packages/core/src/app/giftCertificate/AppliedGiftCertificate.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+        const usedAmountElement = screen.getByTestId('giftCertificate-amount');
+        const remainingAmountElement = screen.getByTestId('giftCertificate-remaining');
+
+        expect(usedAmountElement).toHaveTextContent(usedAmount);
+        expect(remainingAmountElement).toHaveTextContent(remainingAmount);
```
richapatel1510, line 29:
```suggestion

        expect(screen.getByTestId('giftCertificate-amount')).toHaveTextContent(usedAmount);
        expect(screen.getByTestId('giftCertificate-remaining')).toHaveTextContent(remainingAmount);
```

packages/core/src/app/giftCertificate/AppliedGiftCertificate.test.tsx
animesh1987: The mock has amount like this https://github.com/bigcommerce/checkout-js/blob/master/packages/core/src/app/giftCertificate/giftCertificate.mock.ts#L6 which after going through currencyFormatter becomes `22.40` since exchange rate is `1.12` with 2 decimal places as part of config.

bc-peng: In other words, can we just write something like `expect(screen.getByTestId('giftCertificate-amount')).toHaveTextContent('22.40');` instead?

animesh1987: That’s ok as well, but feel its better to have the config and amount. Or else it can get confusing since mock value of 20 becomes 22.40. And if someone in future changes the mock, the test will fail.

bc-peng (approved):

richapatel1510 (approved):

PR #2228, opened 2025-03-26
test(checkout): CHECKOUT-8937 Clean up Payment Tests Part 2

bc-peng:
## What?
- Remove last remaining enzyme tests from `packages/core/src/app/payment/`.

Coverage Before
Screenshot 2025-03-26 at 17 29 27

Coverage After
Screenshot 2025-03-26 at 17 20 09

## Why?
Enable us to upgrade to React 19.

## Testing / Proof
CI.

@bigcommerce/team-checkout


packages/core/src/app/payment/Payment.test.tsx
```diff
@@ -0,0 +1,323 @@
...
+
+        const location = window.location;
+
+        Object.defineProperty(window, 'location', {
+            value: {
+                ...location,
+                replace: jest.fn(),
+            },
+            writable: true,
+        });
```
bc-peng, line 134: `...location,` is needed to keep `msw` working.
😂 Learned this in a hard way.

packages/test-framework/src/react-testing-library-support/CheckoutPageNodeObject.ts
```diff
@@ -164,14 +169,22 @@ export class CheckoutPageNodeObject {
                 );
                 break;
 
-            case CheckoutPreset.CheckoutWithMultiShipping:
+            case CheckoutPreset.CheckoutWithMultiShippingCart:
```
bc-peng, line 172: The mock does not have `consignments` value, so it is more accurate to be called `CheckoutWithMultiShippingCart`.

animesh1987 (approved):

richapatel1510 (approved):

PR #2231, opened 2025-03-31
test(checkout): CHECKOUT-9121 Convert Enzyme Tests in Cart

bc-peng:
## What?
- Remove enzyme tests from `packages/core/src/app/cart/`.

Coverage Before
Screenshot 2025-03-31 at 14 56 17

Coverage After
Screenshot 2025-03-31 at 14 30 09

## Why?
Enable us to upgrade to React 19.

## Testing / Proof
CI.


packages/core/src/app/cart/AppliedRedeemable.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+                foo
+            </AppliedRedeemable>
+        </LocaleContext.Provider>
+    );
+
+    it('renders children and button', async () => {
+        render(<AppliedRedeembleContainer isRemoving={false} />);
+
+        // eslint-disable-next-line testing-library/no-node-access
+        expect(document.querySelector('.is-loading')).not.toBeInTheDocument();
```
richapatel1510, line 26: Curious to know if there is any difference in using `document` over `container` that is returned by render?

packages/core/src/app/cart/CartSummary.test.tsx
```diff
@@ -1,26 +1,28 @@
 import { CheckoutService, createCheckoutService } from '@bigcommerce/checkout-sdk';
-import { mount, ReactWrapper } from 'enzyme';
 import React from 'react';
 
 import { ExtensionProvider } from '@bigcommerce/checkout/checkout-extension';
+import {ErrorLogger} from '@bigcommerce/checkout/error-handling-utils';
```
richapatel1510, line 5:
```suggestion
import { ErrorLogger } from '@bigcommerce/checkout/error-handling-utils';
```

packages/core/src/app/cart/CartSummary.test.tsx
```diff
@@ -1,26 +1,28 @@
...
 import { createLocaleContext, LocaleContext, LocaleContextType } from '@bigcommerce/checkout/locale';
 import { CheckoutProvider } from '@bigcommerce/checkout/payment-integration-api';
+import { render, screen } from '@bigcommerce/checkout/test-utils';
 
 import { getCheckout } from '../checkout/checkouts.mock';
+import {createErrorLogger} from '../common/error';
```
richapatel1510, line 11:
```suggestion
import { createErrorLogger } from '../common/error';
```

packages/core/src/app/cart/Redeemable.tsx
```diff
@@ -103,7 +103,7 @@ const RedeemableForm: FunctionComponent<
 
             // note: to prevent submitting main form, we manually intercept
             // the enter key event and submit the "subform".
-            if (event.keyCode === 13) {
+            if (event.keyCode === 13 || event.key === 'Enter') {
```
animesh1987, line 106: How come we added this change here?

richapatel1510, line 106: Is this new addition only for tests?

packages/core/src/app/cart/AppliedRedeemable.test.tsx
animesh1987: Good catch, probably better to use `container` here.

packages/core/src/app/cart/Redeemable.tsx
bc-peng: `keyCode` is described as deprecated [here](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode). In the testing envirnment, its value is `null`.
I didn't remove it in order to avoid unknown side effects.

packages/core/src/app/cart/AppliedRedeemable.test.tsx
bc-peng: Thought using `document` could avoid making a `const` container.

animesh1987 (approved):

richapatel1510 (approved):

PR #2235, opened 2025-04-02
feat(checkout): CHECKOUT-8521 Displays shipping discount on order confirmation page

bc-mackxu:
## What?  
Show the final shipping amount after applying automatic shipping discounts on the order confirmation page. The original shipping amount will be strikethrough.

## Why?  
To enhance transparency, allowing shoppers to easily see the applied shipping discounts.

## Testing / Proof
- [x] unit test
- [x] manual test

Desktop view
image

Mobile view
image

@bigcommerce/team-checkout


packages/core/src/app/order/OrderSummary.test.tsx
```diff
@@ -31,7 +33,7 @@ describe('OrderSummary', () => {
                     <ExtensionProvider checkoutService={checkoutService}>
                         <LocaleProvider checkoutService={checkoutService}>
                             <OrderSummary
-                                {...mapToOrderSummarySubtotalsProps(order)}
```
richapatel1510, line 34: Should we also add some tests with the flag off?

packages/core/src/app/order/OrderSummaryModal.test.tsx
```diff
@@ -28,7 +28,7 @@ describe('OrderSummaryModal', () => {
         render(
             <OrderSummaryModal
                 isOpen={true}
-                {...mapToOrderSummarySubtotalsProps(order)}
+                {...mapToOrderSummarySubtotalsProps(order, true)}
```
richapatel1510, line 31: same, some tests with the flag off?

packages/core/src/app/order/OrderSummary.test.tsx
bc-mackxu: I don't think we need to add a test with the flag turned off in this case, since the change is newly introduced and the existing tests should cover the behaviour when the experiment is on/off. I've added a new test to cover the scenario when the experiment is enabled.

richapatel1510: The above tests that I am referring to are the existing tests, aren't they? They will run with flag true?

bc-mackxu: I mean, if I copy the existing test and create another one for when the experiment is off, it would be exactly the same—except for passing the experiment value as `false`. Is it necessary to do that in this case? 🤔 
FYI, the flag is LD experiment flag.

richapatel1510: I mean in some tests can we assert that the strikethrough price is visible when experiment is ON and not visible when experiment is OFF?

bc-mackxu: Thanks for the feedback! Since the strikethrough logic is newly introduced and only active when the experiment is ON, I felt testing just the ON case was sufficient. That said, I’ve added a test for the OFF case as suggested—happy to keep it consistent 🙂.
https://github.com/bigcommerce/checkout-js/pull/2235/files#diff-f7ad407279a870fd248832f70ff4db4cae5ffe1a0f1bcf839e4154b5ce295b13R138

richapatel1510 (approved):

bc-peng (approved):

animesh1987 (approved):

PR #2240, opened 2025-04-03
feat(checkout): CHECKOUT-9107 disable continue button when checkout is reloading

richapatel1510:
## What?
Disable Continue Button During Checkout Reload to Prevent Data Discrepancy

## Why?
To keep checkout data in sync and do not allow other checkout update calls when some are already in process.

## Testing / Proof

- Manual testing

**Without extension local testing**

https://github.com/user-attachments/assets/aab87547-6720-41d2-b46c-fd1e5c3fe85c

**With extension**

https://github.com/user-attachments/assets/0d9bafee-84d9-40cd-a601-259026997ee6

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/shipping/Shipping.tsx
+++ b/packages/core/src/app/shipping/Shipping.tsx
@@ -361,6 +361,7 @@ export function mapToShippingProps({
             isUpdatingBillingAddress,
             isUpdatingCheckout,
             isDeletingConsignment,
+            isLoadingCheckout,
         },
     } = checkoutState;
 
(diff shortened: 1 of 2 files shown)


packages/core/src/app/shipping/Shipping.tsx
```diff
@@ -392,7 +393,8 @@ export function mapToShippingProps({
         isUpdatingBillingAddress() ||
         isUpdatingCheckout() ||
         isCreatingCustomerAddress() ||
-        isDeletingConsignment();
+        isDeletingConsignment() ||
+        isLoadingCheckout();
```
bc-peng, line 397: If possible, can we add a unit test?

richapatel1510: Added https://github.com/bigcommerce/checkout-js/pull/2240/commits/4fed495aafaca8a135c323ad99679dc7be154e9b

bc-peng (approved):
Nice work.👏👏

animesh1987 (approved):

PR #2268, opened 2025-04-15
fix(payment): STRIPE-725 Fix styling for custom hosted payment element

PavlenkoM:
## What?
Fix styling for custom hosted payment element.
Original PR: [https://github.com/bigcommerce/checkout-js/pull/2036](https://github.com/bigcommerce/checkout-js/pull/2036)

## Why?
To keep custom check list container clear, remove unnecessary paddings.

## Testing / Proof
Before:
Screenshot 2025-04-15 at 16 14 52

After:
Screenshot 2025-04-15 at 16 28 25

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/scss/components/checkout/checklist/_checklist.scss
+++ b/packages/core/src/scss/components/checkout/checklist/_checklist.scss
@@ -33,6 +33,13 @@
     background-color: container("fill", "dark");
 }
 
+.custom-checklist-item {
+    .paymentMethod--hosted,
+    .widget {
+        padding: 0;
+    }
+}
+
 //
 // CHECKLIST HEADER
 // -----------------------------------------------------------------------------


serhii-tkachenko (approved):

bc-sebastianszafraniec (approved):

vitalii-koshovyi (approved):

bc-dronov (approved):

packages/core/src/scss/components/checkout/checklist/_checklist.scss
```diff
@@ -33,6 +33,16 @@
     background-color: container("fill", "dark");
 }
 
+.custom-checklist-item {
+    margin: 0;
+    overflow: hidden;
```
bc-peng, line 38: I noticed that this class is being used as `form-checklist-item optimizedCheckout-form-checklist-item custom-checklist-item` in checkout-js.

I also saw that `overflow: hidden;` is defined at L20. Is this redundant?

Could you explain how `margin: 0;` won’t cause any unwanted side effects for the other payment methods?

PavlenkoM: Hi @bc-peng 
You are right, we don't need `overflow: hidden;` - removed.

`margin: 0;` won't case any effects to  the other payment methods because for now we are using this style with margin 0 for our list items. So this option removed as well, because we already have it.

bc-peng (approved):

PR #2279, opened 2025-04-22
feature(payment): create blank package bigcommerce-payments-integration

serhiiFilkovskyi:
## What?
Create blank package for bigcommerce-payments-integration with scope of integration

## Why?
To create separate package for BCP (bigcommerce payments) features

## Testing / Proof
Checkout build command passed successfully


.circleci/config.yml
```diff
@@ -74,6 +74,7 @@ jobs:
 
   build:
     <<: *node_executor
+    resource_class: medium+
```
serhii-tkachenko, line 77: Why did you add this change?

serhiiFilkovskyi: because build has failed from time to time and advice was to add this parameter to job

animesh1987 (approved):

bc-peng (approved):

PR #2286, opened 2025-05-01
chore(checkout): Upgrade Formik 2

bc-peng:
## What?
Upgrade Formik from `v1.5.8` to `v2`.

I’ve made some comments in the PR to explain the breaking changes.

## Why?
Formik `v1.5.8` has dependency conflicts with React 18. Therefore, Formik `v2` is necessary.

## Testing / Proof
CI.

### DIFF
--- a/jest-setup.ts
+++ b/jest-setup.ts
@@ -1,3 +1,4 @@
+/* eslint-disable no-console,jest/no-standalone-expect */
 import '@testing-library/jest-dom';
 import '@testing-library/jest-dom/extend-expect';
 import { configure as configureRTL } from '@testing-library/react';
@@ -36,6 +37,25 @@ Object.defineProperty(
 
 (global as any).__webpack_public_path__ = undefined;
 
+const originalConsoleError = console.error;
+const originalConsoleWarn = console.warn;
+
 beforeAll(() => {
     expect.hasAssertions();
+
+    console.error = (...args: unknown[]) => {
+        if (args.map(String).join().includes('Formik')) {
+            return;
+        }
+
+        originalConsoleError(...args);
+    };
+
+    console.warn = (...args: unknown[]) => {
+        if (args.map(String).join().includes('Formik')) {
+            return;
+        }
+
+        originalConsoleWarn(...args);
+    };
 });
--- a/packages/core/src/app/common/form/withFormikExtended.test.tsx
+++ b/packages/core/src/app/common/form/withFormikExtended.test.tsx
@@ -1,5 +1,7 @@
 import userEvent from '@testing-library/user-event';
+import { ErrorMessage } from 'formik';
 import React, { ComponentType } from 'react';
+import { act } from 'react-dom/test-utils';
 import { object, string } from 'yup';
 
 import { render, screen } from '@bigcommerce/checkout/test-utils';
@@ -12,7 +14,6 @@ import withFormikExtended from './withFormikExtended';
 describe('withFormikExtended', () => {
     interface TestComponentProps {
         defaultTitle?: string;
-        isInitialValueLoaded?: boolean;
         submit(values: TestComponentValues): void;
     }
 
@@ -24,6 +25,7 @@ describe('withFormikExtended', () => {
         <Form>
             <Label htmlFor="title">Title</Label>
             <Input id="title" name="title" type="text" />
+            <ErrorMessage component="div" name="title" />
             <Button type="submit">Submit</Button>
         </Form>
     );
@@ -35,9 +37,7 @@ describe('withFormikExtended', () => {
         mapPropsToValues({ defaultTitle }) {
             return { title: defaultTitle };
         },
-        isInitialValid({ defaultTitle }) {
-            return !!defaultTitle;
-        },
+        enableReinitialize: true,
         validationSchema: object({
             title: string().required(),
         }),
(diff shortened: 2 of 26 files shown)


jest-setup.ts
```diff
@@ -36,6 +37,25 @@ Object.defineProperty(
...
+        originalConsoleError(...args);
+    };
+
+    console.warn = (...args: unknown[]) => {
+        if (args.map(String).join().includes('Formik')) {
+            return;
+        }
+
+        originalConsoleWarn(...args);
+    };
```
bc-peng, line 60: Hide the warning.
```
console.warn
    Warning: isInitialValid has been deprecated and will be removed in future versions of Formik. Please use initialErrors or validateOnMount instead.
```

jest-setup.ts
```diff
@@ -36,6 +37,25 @@ Object.defineProperty(
...
 beforeAll(() => {
     expect.hasAssertions();
+
+    console.error = (...args: unknown[]) => {
+        if (args.map(String).join().includes('Formik')) {
+            return;
+        }
+
+        originalConsoleError(...args);
+    };
```
bc-peng, line 52: Hide the error:
```
    Warning: An update to Formik inside a test was not wrapped in act(...).

    When testing, code that causes React state updates should be wrapped into act(...):

    act(() => {
      /* fire events that update state */
    });
    /* assert on the output */
```

packages/core/src/app/payment/Payment.test.tsx
```diff
@@ -139,8 +140,13 @@ describe('Payment step', () => {
 
         await checkout.waitForPaymentStep();
 
-        await userEvent.click(screen.getByRole('radio', { name: 'Cash on Delivery' }));
-        await userEvent.click(screen.getByText('Place Order'));
+        expect(screen.getByRole('radio', { name: 'Pay in Store', checked: true })).toBeInTheDocument();
+
+        await act(async () => userEvent.click(screen.getByRole('radio', { name: 'Cash on Delivery' })));
+
+        expect(await screen.findByRole('radio', { name: 'Cash on Delivery', checked: true })).toBeInTheDocument();
```
bc-peng, line 147: Add checks to ensure a payment method is selected.

packages/core/src/app/shipping/SingleShippingForm.test.tsx
```diff
@@ -229,31 +229,6 @@ describe('SingleShippingForm', () => {
         );
     });
 
-    it('calls updateAddress when address is updated but form is rendered before its initial value is loaded', async () => {
```
bc-peng, line 232: This test is not valid any more as only two fields have values, it cannot pass validation to submit the form.

packages/core/src/app/shipping/SingleShippingForm.tsx
```diff
@@ -145,8 +145,22 @@ class SingleShippingForm extends PureComponent<
...
+        );
+
+        if (
+            stateOrProvinceCodeFormField &&
+            shippingAddress?.stateOrProvinceCode &&
+            values.shippingAddress?.stateOrProvinceCode !== shippingAddress.stateOrProvinceCode
+        ) {
+            setFieldValue('shippingAddress.stateOrProvinceCode', shippingAddress.stateOrProvinceCode);
+        }
+
```
bc-peng, line 163: An interesting formik v2 bug during manual testing.  It wasn’t an issue previously.

Initially, the `stateOrProvinceCode` field is not present when the shipping step is first shown. However, after some time, it appears in the form with an empty value if the countries does have `stateOrProvinceCode` options.

To address this, manually setting `stateOrProvinceCode` in formik v2.

packages/core/src/app/common/form/withFormikExtended.test.tsx
```diff
@@ -35,28 +37,32 @@ describe('withFormikExtended', () => {
         mapPropsToValues({ defaultTitle }) {
             return { title: defaultTitle };
         },
-        isInitialValid({ defaultTitle }) {
-            return !!defaultTitle;
-        },
+        enableReinitialize: true,
```
bc-peng, line 40: Didn't change `enableReinitialize` in production code, only here for testing HOC.

jest-setup.ts
bc-peng: We may need to replace `formik` or update `isInitialValid` before it is gone.

jest-setup.ts
bc-peng: Meanwhile, `eslint` would pick up the fix as `no-unnecessary-act`. 😂

packages/core/src/app/shipping/stripeUPE/StripeShippingForm.test.tsx
```diff
@@ -83,7 +83,7 @@ describe('StripeShippingForm', () => {
         const { container } = renderContainer({ isLoading: false });
 
         expect(defaultProps.initialize).toHaveBeenCalled();
-        expect(defaultProps.getFields).toHaveBeenCalledTimes(3);
+        expect(defaultProps.getFields).toHaveBeenCalledTimes(2);
```
bc-peng, line 86: The reduced call to `defaultProps.getFields`  is the one in `validationSchema` as it is not called at the time of rendering form any more.

bc-peng, line 86: The removed 1 call to `defaultProps.getFields`  is the one in `validationSchema` as it is not called at the time of rendering form any more.

jest-setup.ts
animesh1987: My ideal goal will be to replace `formik` with `react-hook-form` but that will be a project in itself. What does `isInitialValid` do?

packages/core/src/app/payment/createPaymentFormService.ts
```diff
@@ -9,11 +9,20 @@ import { FormContextType } from '@bigcommerce/checkout/ui';
...
-    const { setFieldTouched, setFieldValue, submitForm, validateForm, values } = formikContext;
+    const {
+        setFieldTouched: formikSetFieldTouched,
+        setFieldValue: formikSetFieldValue,
+        submitForm,
+        validateForm,
+        values,
+    } = formikContext;
+
+    const setFieldTouched = formikSetFieldTouched as PaymentFormService['setFieldTouched'];
```
animesh1987, line 24: Why do we need to map `setFieldTouched` to `formikSetFieldTouched` and then do this here?

packages/core/src/app/shipping/SingleShippingForm.tsx
animesh1987: Maybe add some comment on this as to why we need to have this logic in place.

jest-setup.ts
bc-peng: >What does `isInitialValid` do?

I believe we wanted to bypass form validation with this.
However, in Formik 2, validation always occurs before submission. 
I have removed all `isInitialValid` and ran `e2e` and `test`, and there was no impact.

packages/core/src/app/payment/createPaymentFormService.ts
animesh1987: It's ok, just wanted to check why. I am ok with this but maybe we move it to L40 so it becomes less confusing?

animesh1987 (approved):

PR #2290, opened 2025-05-07
fix(checkout): Update formik logic

bc-peng:
## What?
Refactor Formik logic to ensure proper handling of the province form field.

## Why?
To resolve the issue where shoppers are unable to select a province value when returning to the shipping step.

## Testing / Proof
CI.

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/shipping/Shipping.test.tsx
+++ b/packages/core/src/app/shipping/Shipping.test.tsx
@@ -4,6 +4,7 @@ import {
     createEmbeddedCheckoutMessenger,
     EmbeddedCheckoutMessenger,
 } from '@bigcommerce/checkout-sdk';
+import faker from '@faker-js/faker';
 import userEvent from '@testing-library/user-event';
 import { noop } from 'lodash';
 import React, { FunctionComponent } from 'react';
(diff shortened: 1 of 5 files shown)


packages/core/src/app/shipping/Shipping.test.tsx
```diff
@@ -500,6 +501,56 @@ describe('Shipping step', () => {
...
+                countryCode: 'AU',
+                stateOrProvinceCode: faker.helpers.arrayElement(['NSW', 'VIC', 'QLD', 'TAS']),
+                postalCode: faker.address.zipCode(),
+            }));
+
+            await checkout.fillAddressForm(randomAddress2);
+
+            expect((checkoutService.updateShippingAddress as any).mock.calls.slice(-1)[0][0]).toEqual(
+                expect.objectContaining(randomAddress2)
+            );
```
bc-peng, line 552: We have not previously tested the state/province dropdown menu.
This test will ensure that both the dropdown menu and the text box state/province scenarios are covered.

packages/core/src/app/shipping/SingleShippingForm.tsx
bc-peng, line 1: Now we want to only set the value if the form field is empty but the shipping address has a state/province value.

packages/test-framework/src/react-testing-library-support/CheckoutPageNodeObject.ts
```diff
@@ -262,18 +262,45 @@ export class CheckoutPageNodeObject {
         await waitFor(() => screen.getByText(/place order/i));
     }
 
-    async fillAddressForm(): Promise<void> {
```
bc-peng, line 265: You can see we didn't test the dropdown menu before. 😂🤦🏻‍♂️

animesh1987:
Spoke offline @bc-peng will test custom form field dropdown as well.

animesh1987 (approved):

richapatel1510 (approved):

PR #2297, opened 2025-05-09
fix(payments) fix consistent re-renders on payment step

bc-dronov:
## What?
 Fix consistent re-renders on payment step
 
## Why?
Because it was a blocker for payment UI

## Testing / Proof
Before changes:

https://github.com/user-attachments/assets/aaccb0a7-e389-4cad-bfa3-80a4b9746868

After changes Stripe UPE behaviour:

https://github.com/user-attachments/assets/38086683-1c5b-416b-a26a-12f86c034694

After changes Adyen behaviour:

https://github.com/user-attachments/assets/f72639ca-ae3b-4b5b-9f6d-5548e41192b3

@bigcommerce/team-checkout


bc-peng (approved):
LGTM. Thanks for making the fix.

PR #2301, opened 2025-05-13
feat(checkout): Add Checkout Extension Using Web Worker

bc-peng:
## What?
This is a draft PR for discussing building checkout extension using web worker.

## Why?
Support building checkout extension using web worker.

## Testing / Proof
N/A.

### DIFF
--- a/packages/core/src/app/checkout/Checkout.tsx
+++ b/packages/core/src/app/checkout/Checkout.tsx
@@ -524,6 +524,7 @@ class Checkout extends Component<
                     if (matched) {
                         return (
                             <LazyContainer>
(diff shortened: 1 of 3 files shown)


packages/core/src/app/checkout/Checkout.tsx
```diff
@@ -524,6 +524,7 @@ class Checkout extends Component<
                     if (matched) {
                         return (
                             <LazyContainer>
+                                <Extension region={ExtensionRegion.GlobalWebWorker} />
```
bc-peng, line 527: ## Topic 5
Should we render a web worker extension in this manner, considering that we still apply the region concept to it?

PR #2306, opened 2025-05-19
feat(checkout): CHECKOUT-9212 Add data-translatable tag

bc-mackxu:
## What?
Add data-translatable tag

## Why?
To support multilang in checkout

## Testing / Proof
- CI

@bigcommerce/team-checkout


bc-mackxu:
Closed since we need to decide whether we will allow translating all the content on the page or just the dynamic untranslated content.

PR #2313, opened 2025-05-22
fix(shipping): Update autocomplete value for Australian google maps addresses

animesh1987:
## What?
Update autocomplete value for Australian google maps addresses

## Why?
Australian addresses with unit/apartment for now duplicate unit number information in address1 and address2 fields which causes confusion with shipping providers.

## Testing / Proof
- CI
- Australian address

https://github.com/user-attachments/assets/387ecfd1-0e9d-46e0-bed2-b440cf169377

- US address

https://github.com/user-attachments/assets/902af906-3d2c-4bae-9465-b13c749d1c65

@bigcommerce/team-checkout


packages/core/src/app/ui/autocomplete/Autocomplete.test.tsx
```diff
@@ -38,40 +38,41 @@ describe('Autocomplete Component', () => {
     });
 
     it('trigger onSelect function when an item is clicked', async () => {
-        const onSelect = jest.fn();
-        const { container } = render(<Autocomplete items={items} onSelect={onSelect} />);
+        let value = '';
+        const onSelect = (item: typeof items[number]) => {
+            value = item.value as string;
+            rerender(Component());
```
richapatel1510, line 44: How come we are able to use this `rerender` function before defining it?

packages/core/src/app/address/googleAutocomplete/mapToAddress.ts
```diff
@@ -16,7 +16,11 @@ export default function mapToAddress(
     const country = countries && countries.find((c) => countryCode === c.code);
     const street2 = accessor.getStreet2();
 
+    // TODO: Apply this fix for US, UK and CA addresses too.
+    const steet1 = countryCode === 'AU' ? accessor.getStreet() : undefined;
+
     return {
+        address1: steet1,
```
richapatel1510, line 23: Will this set `address1` as undefined for all other countries? Will that not cause issues?

packages/core/src/app/address/googleAutocomplete/AddressSelectorFactory.ts
```diff
@@ -14,6 +15,9 @@ export default class AddressSelectorFactory {
 
             case 'CA':
                 return new AddressSelectorCA(autocompleteData);
+            
+            case 'AU':
+                return new AddressSelectorAU(autocompleteData);
```
richapatel1510, line 20: Shall we add any tests for this?

packages/core/src/app/ui/autocomplete/Autocomplete.test.tsx
animesh1987: I suppose it's due to JS hoisting, not a good practice but that's the only way I found so far 😞

animesh1987: Changed the test.

packages/core/src/app/address/AddressForm.tsx
```diff
@@ -202,12 +202,18 @@ class AddressForm extends Component<AddressFormProps & WithLanguageProps> {
         const address = mapToAddress(place, countries);
 
         forIn(address, (value, fieldName) => {
+            if (fieldName === AUTOCOMPLETE_FIELD_NAME && value === undefined) {
+                return;
+            }
```
bc-peng, line 207: Can you explain what the problem would be if `setFieldValue(AUTOCOMPLETE_FIELD_NAME, undefined);` runs?

richapatel1510 (approved):

packages/core/src/app/address/AddressForm.tsx
animesh1987: I am not sure when that will run, since if address1 is undefined we fallback to the existing implementation.

bc-peng: I'm asking because I see it's an early return to avoid `setFieldValue(AUTOCOMPLETE_FIELD_NAME, undefined);`.

Perhaps we just don't want to run `onChange(AUTOCOMPLETE_FIELD_NAME, undefined);`?

bc-peng (approved):

PR #2314, opened 2025-05-22
feat(payment): created utility package to contain utility methods that can be used across whole application

serhii-tkachenko:
## What?
Created utility package to contain utility methods that can be used across whole application.

Also transferred `navigateToOrderConfirmation` and `isBuyNowCard` function from core to `utilities` package.

## Why?
Some payment provider related files still available in core packages, so to move them to related integration package some utility functions should be available to use outside of core package. To resolve migration issue, this package was created and several method were moved to it.

## Testing / Proof
Unit tests
Manual tests:
- Checked if PayPal Commerce button works as expected on checkout page;
- Checked if there any issue with placing orders with BuyNow cart;


packages/utility/jest.config.js
```diff
@@ -0,0 +1,13 @@
+module.exports = {
+    displayName: 'utility',
+    preset: '../../jest.preset.js',
+    transform: {
+        '^.+\\.[tj]sx?$': ['ts-jest', {
+            tsconfig: '<rootDir>/tsconfig.spec.json',
+            diagnostics: false,
+        }],
+    },
+    collectCoverageFrom: ['src/**/*.{ts,tsx}', '!src/**/*/index.ts', '!src/icon/*.tsx'],
```
bc-peng, line 10:
```suggestion
    collectCoverageFrom: ['src/**/*.{ts,tsx}', '!src/**/*/index.ts'],
```

bc-peng (approved):
LGTM👏👏

richapatel1510 (approved):

PR #2317, opened 2025-05-27
fix(shipping): reset value in formik on checkbox unmount

animesh1987:
## What?
Reset value in formik on checkbox unmount

## Why?
Formik forms tend to save value in formik contexts across form renders. Clear form value when checklist item is unmounted.

## Testing / Proof
- CI
- TBD

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/ui/form/Checklist.tsx
+++ b/packages/core/src/app/ui/form/Checklist.tsx
@@ -1,3 +1,4 @@
+import { Accordion } from '@bigcommerce/checkout/ui';
 import { noop } from 'lodash';
 import React, {
     createContext,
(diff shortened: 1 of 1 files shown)


bc-peng (approved):

packages/core/src/app/ui/form/Checklist.tsx
```diff
@@ -29,6 +30,12 @@ export const ChecklistContext = createContext<ChecklistContextProps | undefined>
 const Checklist: FunctionComponent<
     ChecklistProps & ConnectFormikProps<{ [key: string]: string }>
 > = ({ formik: { setFieldValue }, name, onSelect = noop, ...props }) => {
+    useEffect(() => {
```
davidchin, line 33: It looks like `PaymentMethodList` also uses `Checklist` component. Can you double check nothing breaks?

PR #2331, opened 2025-05-30
chore(common): testing if e2e suit will fail on master

serhii-tkachenko:
## What?
testing if e2e suit will fail on master

## Why?
testing if e2e suit will fail on master

## Testing / Proof
testing if e2e suit will fail on master


AndrewBarber:
@serhii-tkachenko did they fail or pass? 👀

PR #2334, opened 2025-05-30
fix(core): CHECKOUT-000 Fix redundant alt text on checkout sidebar product images

### LINKED ISSUE #2332: Redundant alt text on checkout sidebar product images - EAA Accessibility Warning
## Description

The current implementation of product images in the checkout sidebar has redundant alt text which can cause accessibility issues for screen reader users. This is a **warning-level** European Accessibility Act (EAA) issue, not a critical violation.

## Current Implementation

In `packages/core/src/app/order/getOrderSummaryItemImage.tsx`, the alt text for product images is set to the product name:

```tsx
return <img alt={item.name} data-test="cart-item-image" src={item.imageUrl} />;
```

This is redundant because the product name is already displayed as text next to the image in the checkout sidebar. Screen readers would read both the image alt text and the visible product name, creating a redundant experience.

## Proposed Solution

Update the alt text to be more descriptive and avoid redundancy, or make the image decorative with an empty alt attribute if the image is purely decorative.

## Accessibility Impact

This change would improve the experience for screen reader users by eliminating redundant information. While this is a warning-level issue rather than a critical violation, addressing it would improve compliance with the European Accessibility Act (EAA) requirements for clear and non-redundant content presentation.

AndrewBarber:
## What?
Fixed redundant alt text on product images in the checkout sidebar to comply with EAA (European Accessibility Act) standards.

Resolves #2332

## Why?
The current implementation uses the product name as alt text, which is redundant since the product name is already displayed next to the image in the UI. This creates a repetitive experience for screen reader users who will hear the same information twice.

According to accessibility best practices, when an image is purely decorative and its content is already conveyed through nearby text, an empty alt attribute (`alt=""`) should be used to prevent screen readers from announcing redundant information.

## Testing / Proof
- Unit tests have been updated to reflect the new implementation
- Manually verified with screen readers that the product information is properly conveyed without redundancy

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/order/getOrderSummaryItemImage.test.tsx
+++ b/packages/core/src/app/order/getOrderSummaryItemImage.test.tsx
@@ -21,7 +21,7 @@ describe('getOrderSummaryItemImage()', () => {
 
         it('returns image', () => {
             expect(getOrderSummaryItemImage(lineLineItem)).toEqual(
-                <img alt={lineLineItem.name} data-test="cart-item-image" src={lineLineItem.imageUrl} />
+                <img alt="" data-test="cart-item-image" src={lineLineItem.imageUrl} />
             );
         });
     });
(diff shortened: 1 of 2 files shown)


animesh1987 (approved):

TomA-R:
Many thanks, Andrew!

PR #2338, opened 2025-05-30
fix(core): CHECKOUT-000 Fix broken ARIA references in shipping inputs

### LINKED ISSUE #2337: fix(core): Broken ARIA References in Shipping Inputs - EAA Accessibility Issue
# Broken ARIA References in Shipping Inputs - EAA Accessibility Issue

## Description
There are broken ARIA references in the shipping input fields. Each shipping input has an `aria-labelledby` attribute that references two IDs, but the second ID (for error messages) points to an element that only exists when an error is present. This fails WCAG A standards and is an EAA (European Accessibility Act) compliance issue.

## Current Implementation
In `DynamicFormField.tsx`, each input has an `aria-labelledby` attribute that references two IDs:
```tsx
aria-labelledby={`${fieldInputId}-label ${fieldInputId}-field-error-message`}
```

The first ID (`${fieldInputId}-label`) refers to the label element, which is always present.
The second ID (`${fieldInputId}-field-error-message`) refers to the error message element.

In `FormFieldError.tsx`, the error message is only rendered when the form is submitted and there's an error:
```tsx
{isSubmitted && <ErrorMessage name={name} render={renderMessage} />}
```

This means that when there's no error, the `aria-labelledby` attribute references a non-existent element, which breaks accessibility.

## Expected Behavior
The error element should always be present in the DOM, even when there's no error. It should be visually hidden when not needed, but still accessible to screen readers when referenced by `aria-labelledby`.

## Impact
This issue affects accessibility for users with screen readers, as they may receive confusing or incomplete information about form fields. This violates WCAG 2.1 Success Criterion 4.1.2 (Name, Role, Value), which is a requirement under the European Accessibility Act.

## Steps to Reproduce
1. Open the checkout page
2. Navigate to the shipping section
3. Inspect any input field
4. Notice that the `aria-labelledby` attribute references an error message element that doesn't exist until an error occurs

## Suggested Fix
Modify `FormFieldError.tsx` to always render the error container, but only show the error message when there's an error. Use CSS to visually hide the error container when there's no error message to display.

AndrewBarber:
## What?
Fixed broken ARIA references in shipping input fields by ensuring that error message elements are always present in the DOM, even when there are no errors.

Resolves #2337

## Why?
Each shipping input had an `aria-labelledby` attribute that referenced two IDs:
1. The label element's ID (always present)
2. The error message element's ID (only present when there was an error)

This caused broken ARIA references when there were no errors, which fails WCAG A standards and is an EAA (European Accessibility Act) compliance issue.

The fix ensures that the error message elements are always present in the DOM, even when there are no errors, by rendering a hidden span with the correct ID when no error is present.

## Testing / Proof
- Manually verified that all aria-labelledby references now point to valid elements in the DOM
- Confirmed that error message elements are present even when there are no errors
- Verified that error elements are properly hidden when there are no errors
- Tested with browser console queries to ensure all referenced IDs exist

Console test code:
```javascript
// Get all shipping input fields
const shippingInputs = document.querySelectorAll('input[name^="shippingAddress."]');

// Check their aria-labelledby references
shippingInputs.forEach(input => {
  const ariaLabelledby = input.getAttribute('aria-labelledby');
  if (ariaLabelledby) {
    const ids = ariaLabelledby.split(' ');
    ids.forEach(id => {
      const element = document.getElementById(id);
      console.log(`Input ${input.name}, ID ${id} exists:`, !!element);
    });
  }
});
```

Console test results:
```
Input shippingAddress.firstName, ID firstNameInput-label exists: true
Input shippingAddress.firstName, ID firstNameInput-field-error-message exists: true
Input shippingAddress.lastName, ID lastNameInput-label exists: true
Input shippingAddress.lastName, ID lastNameInput-field-error-message exists: true
Input shippingAddress.company, ID companyInput-label exists: true
Input shippingAddress.company, ID companyInput-field-error-message exists: true
Input shippingAddress.phone, ID phoneInput-label exists: true
Input shippingAddress.phone, ID phoneInput-field-error-message exists: true
Input shippingAddress.address2, ID addressLine2Input-label exists: true
Input shippingAddress.address2, ID addressLine2Input-field-error-message exists: true
Input shippingAddress.city, ID cityInput-label exists: true
Input shippingAddress.city, ID cityInput-field-error-message exists: true
Input shippingAddress.stateOrProvince, ID provinceInput-label exists: true
Input shippingAddress.stateOrProvince, ID provinceInput-field-error-message exists: true
Input shippingAddress.postalCode, ID postCodeInput-label exists: true
Input shippingAddress.postalCode, ID postCodeInput-field-error-message exists: true
```

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/app/ui/form/FormFieldError.tsx
+++ b/packages/core/src/app/ui/form/FormFieldError.tsx
@@ -1,4 +1,4 @@
-import { ErrorMessage } from 'formik';
+import { ErrorMessage, useFormikContext } from 'formik';
 import React, { FunctionComponent, memo, useCallback } from 'react';
 
 import { FormContext } from '@bigcommerce/checkout/ui';
@@ -10,30 +10,44 @@ export interface FormFieldErrorProps {
 }
 
 const FormFieldError: FunctionComponent<FormFieldErrorProps> = ({ name, testId, errorId }) => {
+    // Get form context to check for errors
+    const formikContext = useFormikContext<{ [key: string]: any }>();
+    const hasError = formikContext?.errors?.[name] && formikContext?.touched?.[name];
+    
     const renderMessage = useCallback(
         (message: string) => (
-            <ul className="form-field-errors" data-test={testId}>
-                <li className="form-field-error">
-                    <label
-                        aria-live="polite"
-                        className="form-inlineMessage"
-                        htmlFor={name}
-                        id={errorId}
-                        role="alert"
-                    >
-                        {message}
-                    </label>
-                </li>
-            </ul>
+            <label
+                aria-live="polite"
+                className="form-inlineMessage"
+                htmlFor={name}
+                id={errorId}
+                role="alert"
+            >
+                {message}
+            </label>
         ),
-        [errorId, name, testId],
+        [errorId, name],
     );
 
     return (
         <FormContext.Consumer>
-            {({ isSubmitted }) =>
-                isSubmitted && <ErrorMessage name={name} render={renderMessage} />
-            }
+            {({ isSubmitted }) => (
+                <ul className="form-field-errors" data-test={testId}>
+                    <li className="form-field-error">
+                        {/* Show error message when form is submitted and there's an error */}
+                        {isSubmitted && <ErrorMessage name={name} render={renderMessage} />}
+                        
+                        {/* When there's no error or form isn't submitted, render a hidden element with the same ID */}
+                        {(!isSubmitted || !hasError) && (
+                            <span 
+                                aria-hidden="true"
+                                className="is-srOnly" 
+                                id={errorId} 
+                            />
+                        )}
+                    </li>
+                </ul>
+            )}
         </FormContext.Consumer>
     );
 };


packages/core/src/app/ui/form/FormFieldError.tsx
```diff
@@ -10,30 +10,44 @@ export interface FormFieldErrorProps {
...
+                        {isSubmitted && <ErrorMessage name={name} render={renderMessage} />}
+                        
+                        {/* When there's no error or form isn't submitted, render a hidden element with the same ID */}
+                        {(!isSubmitted || !hasError) && (
+                            <span 
+                                aria-hidden="true"
+                                className="is-srOnly" 
+                                id={errorId} 
+                            />
+                        )}
```
bc-peng: In my manual testing, I saw two elements with the same id.


I'm working on a solution for the raised issue.

cc @animesh1987 

```
<li class="form-field-error">
    <label 
        aria-live="polite" 
        class="form-inlineMessage" 
        for="shippingAddress.lastName" 
        id="lastNameInput-field-error-message" 
        role="alert">
        Last Name is required
    </label>
    <span 
        aria-hidden="true" 
        class="is-srOnly" 
        id="lastNameInput-field-error-message">
    </span>
</li>
```
image

packages/core/src/app/ui/form/FormFieldError.tsx
```diff
@@ -1,39 +1,51 @@
...
 export interface FormFieldErrorProps {
     name: string;
     testId?: string;
     errorId: string;
 }
 
 const FormFieldError: FunctionComponent<FormFieldErrorProps> = ({ name, testId, errorId }) => {
+    const formikContext = useFormikContext<{ [key: string]: any }>();
+
+    const hasError = getNestedValue(formikContext?.errors, name) && getNestedValue(formikContext?.touched, name);
```
bc-peng, line 17: The actual value of `formikContext.errors` looks like this:
```
{
  shippingAddress: {
    postalCode: "Postal Code is required",
    stateOrProvinceCode: "State/Province is required",
    city: "City is required",
    address1: "Address is required",
    lastName: "Last Name is required",
    firstName: "First Name is required"
  }
}
```
Therefore, to check for an error on a field such as `firstName`, we need to access it via `formikContext?.errors.shippingAddress?.firstName`.

packages/core/src/app/ui/form/FormFieldError.tsx
```diff
@@ -1,39 +1,51 @@
...
-                        htmlFor={name}
-                        id={errorId}
-                        role="alert"
-                    >
-                        {message}
-                    </label>
-                </li>
-            </ul>
+            <label
+                aria-live="polite"
```
animesh1987, line 22: Should this be assertive?

bc-peng: `assertive` may overwhelm a shopper by reading aloud all error messages at once.
>[assertive](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-live#assertive)
Indicates that updates to the region have the highest priority and should be presented to the user immediately.

> [polite](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-live#polite)
Indicates that updates to the region should be presented at the next graceful opportunity, such as at the end of speaking the current sentence or when the user pauses typing.

bc-launchbay:
success [20250618021421](https://launchbay.bigcommerce.net/projects/535/releases/257847) by Launchbay

bc-launchbay:
success [20250618021421](https://launchbay.bigcommerce.net/projects/535/releases/257847) by David Chin

PR #2349, opened 2025-06-04
chore(shipping): update padding in shipping step

animesh1987:
## What?
Update padding in shipping step and billing

## Why?
Update padding in shipping step as per updated designs.

## Testing / Proof
- CI
- screenshot 
#### Before

https://github.com/user-attachments/assets/94651388-0c73-4dbd-a6d1-a45915012e9e

#### After

https://github.com/user-attachments/assets/5d869c97-5447-431a-b2d5-c0ca63e7e1f7

@bigcommerce/team-checkout

### DIFF
--- a/packages/core/src/scss/components/foundation/forms/_forms.scss
+++ b/packages/core/src/scss/components/foundation/forms/_forms.scss
@@ -21,7 +21,7 @@
     }
 
     .form-fieldset {
-        margin-bottom: spacing("single");
+        margin-bottom: #{spacing("half") + spacing("third")};
 
         &:last-child {
             margin-bottom: 0;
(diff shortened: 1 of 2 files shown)


packages/core/src/scss/components/foundation/forms/_forms.scss
```diff
@@ -186,7 +186,8 @@
         background-color: $legend-bg;
         color: $header-font-color;
         font-weight: $legend-font-weight;
-        padding: spacing("half") 0;
+        padding-top: 0;
```
richapatel1510: Is this needed, if we provide all 4 paddings below?

bc-peng: This rule seems to be overridden by the line below.

bc-peng (approved):

richapatel1510 (approved):

PR #2365, opened 2025-06-16
Guest multishipping address UI

richapatel1510:
## What?
Add UI for guest multishipping address entry

## Why?
So that guest customer is able to add/edit new address for multishipping its items.

## Testing / Proof

- CI checks
- Manual testing

https://github.com/user-attachments/assets/9767ee85-91aa-4ecc-b4cc-68e8498166c2


packages/core/src/app/address/AddressFormModal.test.tsx
```diff
@@ -84,7 +85,12 @@ describe('AddressFormModal Component', () => {
 
         expect(screen.getByText('First Name is required')).toBeInTheDocument();
         expect(screen.getByText('Last Name is required')).toBeInTheDocument();
-        expect(screen.getByText('Address is required')).toBeInTheDocument();
+
+        await userEvent.clear(screen.getByTestId('addressLine1Input-text'));
+        await userEvent.click(screen.getByText('Save Address'));
+
+        await new Promise((resolve) => process.nextTick(resolve));
```
bc-peng: Is it possible that the test can pass without this line? However, you may need to replace `getByText` with `findByText`.

packages/ui/src/icon/IconEdit.tsx
```diff
@@ -0,0 +1,11 @@
+import React, { FunctionComponent } from 'react';
+
+import withIconContainer from './withIconContainer';
```
animesh1987, line 3: I will suggest we only add this one in ui package and use it. And not duplicate the component in core.

packages/core/src/app/address/AddressFormModal.tsx
```diff
@@ -83,22 +84,12 @@ const SaveAddressForm = withLanguage(
...
-            customFields: {},
-            country: '',
-            countryCode: defaultCountryCode,
-            stateOrProvince: '',
-            stateOrProvinceCode: '',
-            postalCode: '',
-            phone: '',
-            city: '',
-            company: '',
-            shouldSaveAddress: false,
```
bc-peng, line 100: I see in `mapAddressToFormValues`, `shouldSaveAddress` sometimes can be `true`.
Is removing this safe?

richapatel1510: AFAIK for logged in customer we always save the address in case of multishipping regardless of this flag so should not be a problem there. Also I think for guest customers even if this passed as `true`, it won't be saving any addresses since there is no registered customer. This modal is only used in multishipping mode.

bc-launchbay:
success [20250618021421](https://launchbay.bigcommerce.net/projects/535/releases/257847) by David Chin

PR #2383, opened 2025-06-20
feat(extension): introduce new ReRenderShippingStep command

richapatel1510:
## What?
Introduce a new checkout extension command.

## Why?
To support the force re-rendering of the shipping step.

## Testing / Proof
TBA


packages/core/src/app/shipping/ShippingForm.tsx
```diff
@@ -87,11 +93,36 @@ const ShippingForm = ({
...
+        if (previousShippingFormRenderTimestamp !== shippingFormRenderTimestamp) {
+            const isMultiShippingMode =
+                !!cart &&
+                !!consignments &&
+                hasMultiShippingEnabled &&
+                isUsingMultiShipping(consignments, cart.lineItems);
+
+            setIsMultishippingMode(isMultiShippingMode);
+        }
+    }, [shippingFormRenderTimestamp]);
```
bc-peng, line 117: I just noticed that the dependency of `shippingFormRenderTimestamp` seems to be equivalent to checking `previousShippingFormRenderTimestamp !== shippingFormRenderTimestamp`. Could you check if we still need `useRef` after adding this dependency? 🙇🏻‍♂️

richapatel1510: Oh yes, that's correct. Updated it. Thanks 🙇

packages/core/src/app/shipping/ShippingForm.tsx
```diff
@@ -87,11 +93,29 @@ const ShippingForm = ({
...
-        extensionState: { shippingFormRenderTimestamp },
-    } = useExtensions();
+        checkoutState: {
+            data: { getConfig },
+        },
+    } = useCheckout();
+    const config = getConfig();
+
+    useEffect(() => {
+
```
bc-peng:
```suggestion
```

bc-peng (approved):
LGTM👏👏

bc-launchbay:
success [20250625015513](https://launchbay.bigcommerce.net/projects/535/releases/258241) by Launchbay

PR #2385, opened 2025-06-23
fix(checkout): CHECKOUT-9311 Fix Billing Address Aria Label

bc-peng:
## What?
Fix billing address's aria label.

## Why?
Improve checkout page's accessiblity.

## Testing / Proof
CI.

### DIFF
--- a/packages/core/src/app/address/AddressSelectButton.tsx
+++ b/packages/core/src/app/address/AddressSelectButton.tsx
@@ -36,8 +36,8 @@ const AddressSelectButton: FunctionComponent<AddressSelectButtonProps & WithLang
     return (
         <a
             aria-controls="addressDropdown"
-            aria-describedby={language.translate('address.enter_or_select_address_action')}
             aria-expanded={ariaExpanded}
+            aria-label={language.translate('address.enter_or_select_address_action')}
             className="button dropdown-button dropdown-toggle--select"
             data-test="address-select-button"
             href="#"


packages/core/src/app/address/AddressSelectButton.tsx
```diff
@@ -36,8 +36,8 @@ const AddressSelectButton: FunctionComponent<AddressSelectButtonProps & WithLang
     return (
         <a
             aria-controls="addressDropdown"
-            aria-describedby={language.translate('address.enter_or_select_address_action')}
```
bc-peng, line 39: > The aria-describedby attribute lists the ids of the elements that describe the object. It is used to establish a relationship between widgets or groups and the text that describes them. [[ref](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-describedby)]


The `aria-describedby` should reference an `id` rather than providing text content directly.

bc-launchbay:
success [20250624014002](https://launchbay.bigcommerce.net/projects/535/releases/258152) by Launchbay

bc-launchbay:
success [20250624014002](https://launchbay.bigcommerce.net/projects/535/releases/258152) by Launchbay

bc-launchbay:
success [20250624014002](https://launchbay.bigcommerce.net/projects/535/releases/258152) by Launchbay

PR #2389, opened 2025-06-24
CHECKOUT-9307: Fix secrets scan filter so PR workflow does not trigger on merge

davidchin:
## What/Why?
Currently, the "pull request" workflow is triggered when a PR is merged into master. This happens because the "Gitleaks secrets scan" job is missing the correct filter, unlike the other jobs in the same workflow.

Screenshot 2025-06-24 at 2 50 24 pm

## Testing / Proof
CircleCI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -234,8 +234,7 @@ workflows:
       - security/scan:
           name: "Gitleaks secrets scan"
           filters:
-            branches:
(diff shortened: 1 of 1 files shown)


bc-launchbay:
success [20250624064447](https://launchbay.bigcommerce.net/projects/535/releases/258173) by Launchbay

PR #2404, opened 2025-06-30
fix(checkout): Fix a11y issue on payment section

richapatel1510:
## What?
Fix A11Y issue of Payment section radio buttons not announced properly on selection by adding a hidden element with its aria-live to be assertive so that it is picked up by screen readers and announced properly. 

## Why?
So that the selected payment method should be announced in the screen reader

## Testing / Proof

- CI checks
- Recording
 
**Before**

https://github.com/user-attachments/assets/3edb8d34-d7a4-4595-86bd-e3e5ce437467

**After**

https://github.com/user-attachments/assets/274c9dc4-c7aa-4d1d-8a3b-6ba7cdb1b0b5

### DIFF
--- a/packages/core/src/app/payment/PaymentForm.test.tsx
+++ b/packages/core/src/app/payment/PaymentForm.test.tsx
@@ -77,8 +77,7 @@ describe('PaymentForm', () => {
         expect(screen.getAllByRole('listitem')).toHaveLength(8);
         expect(screen.getAllByRole('group')).toHaveLength(2);
         expect(screen.getAllByRole('radio')).toHaveLength(2);
-        expect(screen.getAllByRole('heading')).toHaveLength(2);
-        expect(screen.getAllByText('Authorizenet')).toHaveLength(2);
+        expect(screen.getAllByText('Authorizenet')).toHaveLength(3);  // 2 radio buttons + 1 title for a11y (hidden) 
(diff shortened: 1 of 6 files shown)


packages/core/src/app/payment/PaymentForm.test.tsx
```diff
@@ -77,8 +77,7 @@ describe('PaymentForm', () => {
         expect(screen.getAllByRole('listitem')).toHaveLength(8);
         expect(screen.getAllByRole('group')).toHaveLength(2);
         expect(screen.getAllByRole('radio')).toHaveLength(2);
-        expect(screen.getAllByRole('heading')).toHaveLength(2);
```
bc-peng, line 80: Is removing this line intentional?

richapatel1510: Yes, because we removed the role="heading" from the payment method title [here](https://github.com/bigcommerce/checkout-js/pull/2404/files#diff-2e242bb8919d43d0b2d5da6fb4fa41c9f6cc8deacd549b997ab5327bb5e0f988L336)

bc-launchbay:
success [20250701054350](https://launchbay.bigcommerce.net/projects/535/releases/258648) by Launchbay

PR #2419, opened 2025-07-09
fix(payment): Make payment method title icon more readable for the sc…

vitalii-koshovyi:
…reen readers

## What?
Make payment method title icon more readable for the screen readers

## Why?
All input fields under the payment section should only be announced once by screen reader.
However, they are announced twice for the Adyen. 
The reason of it is a icon alt attribute, that sets to the methodName. I've added icon word to the alt attribute for more understandable reading.
image

## Testing / Proof
image


bc-launchbay:
success [20250709084331](https://launchbay.bigcommerce.net/projects/535/releases/259227) by Launchbay

PR #2422, opened 2025-07-10
CHECKOUT-9204: Hydrate initial state to reduce latency caused by request waterfalls

davidchin:


packages/core/src/app/checkout/hooks/useLoadCheckout.ts
```diff
@@ -43,12 +44,24 @@ export const useLoadCheckout = (checkoutId: string): {isLoadingCheckout: boolean
...
+
+        if (!initialState) {
+            // If the initial data has not been preloaded from the server, we need to make API calls to fetch it.
+            fetchDataWithRetry()
+                .then(() => setIsLoadingCheckout(false))
+                .catch((error) => {
+                    throw error;
+                });
+        } else {
+            requestAnimationFrame(async () => {
```
animesh1987: Is there a reason why we need to wrap the initial render and hydration calls in `requestAnimationFrame`?

animesh1987 (approved):

packages/core/src/app/checkout/hooks/useLoadCheckout.ts
davidchin: The idea is to spread the work across multiple tasks so we don't do too much in a single one, as this could negatively impact Total Blocking Time. Having said that, I notice I'm using two, so I'll look into whether I need both.

davidchin: Ok I checked. We do need do break up the task in both places. But I found that `requestAnimationFrame` is not the most ideal. `scheduler.yield` (https://developer.mozilla.org/en-US/docs/Web/API/Scheduler/yield) is a better option.

PR #2434, opened 2025-07-16
fix(payment): updated billing form to show fastlane addresses only for guest users

serhii-tkachenko:
## What?
Updated billing form to show fastlane addresses only for guest users

## Why?
Fastlane is not supported for BC logged in customers, so logged in customers should be able to use addresses from their address book in billing step

## Testing / Proof
Manual tests
CI

Before:
Screenshot 2025-07-16 at 17 32 15

After:
Screenshot 2025-07-16 at 17 32 37


bc-peng (approved):

bc-launchbay:
success [20250717072311](https://launchbay.bigcommerce.net/projects/535/releases/259849) by Launchbay

PR #2463, opened 2025-07-30
feat(checkout): CHECKOUT-9395 Rename StyleContext to ThemeContext

bc-peng:
## What?
Rename the existing `StyleContext` to `ThemeContext`.
Update tests accordingly.

## Why?
Improve code quality and readability.

## Testing / Proof
CI.


packages/ui/.eslintrc.json
```diff
@@ -4,6 +4,7 @@
     "jsx-a11y/anchor-is-valid": "off",
     "@typescript-eslint/naming-convention": "off",
     "@typescript-eslint/consistent-type-assertions": "off",
-    "@typescript-eslint/no-unsafe-call": "off"
+    "@typescript-eslint/no-unsafe-call": "off",
+    "@nx/enforce-module-boundaries": "off"
```
bc-peng, line 8: Circular dependency is evitable because `test-utils` needs the theme context from `ui`, `ui` needs `test-utils`. 😂

animesh1987 (approved):

bc-launchbay:
success [20250731003715](https://launchbay.bigcommerce.net/projects/535/releases/260665) by Launchbay

bc-launchbay:
success [20250731003715](https://launchbay.bigcommerce.net/projects/535/releases/260665) by Launchbay

packages/ui/.eslintrc.json
richapatel1510: Yep, that was the reason to keep that context in the integration package in first place 😃

animesh1987: Maybe we need to rethink how the code is setup in that case. So we don't have to turn this rule off. A new package perhaps?

bc-peng: Thanks all! 

For next steps, I’ll finish up the `newFontStyle` renaming and restructure work first.

After that, I’ll create a new ticket to make a `contexts` package and start moving all the contexts from `payment-integration-api` and `ui` into it. This should help avoid the circular dependency and keep things cleaner going forward.

PR #2469, opened 2025-08-01
fix(payment): Unable to place an order when PayLater is enabled

bc-nick:
## What?

Fix of infinite BCP Pay Later Banner initialisation

## Why?

To fix an issue with placing an order via BCP

## Testing / Proof

Before

https://github.com/user-attachments/assets/cc8cf669-d00c-4ded-8307-014d099012e2

After

https://github.com/user-attachments/assets/9e550ac6-56e1-4a59-99c8-21d9ca384348


bc-launchbay:
success [20250804073322](https://launchbay.bigcommerce.net/projects/535/releases/260819) by Launchbay

PR #2480, opened 2025-08-06
CHECKOUT-9406: Replace Sentry npm installation with loader script to reduce bundle size

davidchin:
## What/Why?
This PR replaces the Sentry NPM installation with the [loader script](https://docs.sentry.io/platforms/javascript/install/loader/) to reduce the application’s bundle size.

The current setup includes unnecessary modules such as tracing, session replay, and Node-specific integrations (e.g., MongoDB), which are not relevant in the browser environment. These unused features significantly increase the bundle size and can negatively impact performance.

By switching to the loader script, we gain the following benefits:
* **Lazy-loaded**: Only a small bootstrap script is included initially. The full Sentry SDK is loaded only when an exception is raised, reducing the upfront payload.
* **Automatic version updates**: Minor and patch version updates are automatically applied, reducing maintenance overhead.

### Before

The screenshots below show that the Sentry bundle accounts for 13% of the initial JavaScript payload and includes numerous unnecessary or unused modules.

Screenshot 2025-08-06 at 11 37 08 am

Screenshot 2025-08-06 at 11 38 19 am

### After

This screenshot shows that the full Sentry SDK is now excluded from the initial JavaScript bundle.

Screenshot 2025-08-06 at 11 35 40 am

This screenshot shows that errors are still being captured and sent to the Sentry server.

Screenshot 2025-08-06 at 11 58 43 am

### DIFF
--- a/packages/core/src/app/common/error/SentryErrorLogger.ts
+++ b/packages/core/src/app/common/error/SentryErrorLogger.ts
@@ -19,12 +19,6 @@ import computeErrorCode from './computeErrorCode';
 import ConsoleErrorLogger from './ConsoleErrorLogger';
 import NoopErrorLogger from './NoopErrorLogger';
 
-declare global {
-    interface Window {
-        sentryOnLoad?: () => void;
-    }
-}
-
 const FILENAME_PREFIX = 'app://';
 
 export interface SentryErrorLoggerOptions {
--- a/packages/core/types/global.d.ts
+++ b/packages/core/types/global.d.ts
@@ -1,0 +1,7 @@
+declare global {
+    interface Window {
+        sentryOnLoad?: () => void;
+    }
+}
+
+export {};


bc-peng (approved):
👏👏 Nice job, solid delivery!

packages/core/src/app/common/error/SentryErrorLogger.ts
```diff
@@ -22,6 +19,12 @@ import computeErrorCode from './computeErrorCode';
 import ConsoleErrorLogger from './ConsoleErrorLogger';
 import NoopErrorLogger from './NoopErrorLogger';
 
+declare global {
```
animesh1987: Should this be in a separate file, something like `global.d.ts` perhaps?

bc-launchbay:
success [20250812010645](https://launchbay.bigcommerce.net/projects/535/releases/261344) promoted by https://launchbay.bigcommerce.net/projects/535/releases/261352 being deployed. Deployed by Launchbay

PR #2489, opened 2025-08-07
CHECKOUT-9416: Reduce bundle size by removing ES5 and unstable polyfills

davidchin:
## What/Why?

Currently, the app targets [ES6](https://github.com/bigcommerce/checkout-js/blob/151df8c24876e83e3a14e05d9ab1ba1bacb35265/webpack.config.js#L197) but includes ES5 polyfills like `regenerator-runtime`, which are unnecessary for browsers that support ES6. It also loads all ECMAScript polyfills, including early-stage proposals, despite only using stable features.

This PR includes the following changes:
* Removed `regenerator-runtime` and other redundant ES5 polyfills
* Included only required polyfills for stable ECMAScript features

### Before
Screenshot 2025-08-07 at 4 11 41 pm

### After
Screenshot 2025-08-12 at 11 18 19 am


packages/core/src/app/polyfill.ts
```diff
@@ -1,2 +1 @@
-import 'core-js';
-import 'regenerator-runtime/runtime';
+import 'core-js/stable';
```
animesh1987: Shall we drop this too, given we are updating docs to remove ie11 support as well?

davidchin: Right now we're still targeting [ES2015](https://github.com/bigcommerce/checkout-js/blob/151df8c24876e83e3a14e05d9ab1ba1bacb35265/webpack.config.js#L197) (ES6) with our esbuild config. So I assume we're still targeting older browsers beside IE11. Therefore I don't think we can drop it completely. Having said that, I can import specific polyfills (e.g.: `import "core-js/modules/es.array.at.js";`) if I know the minimum versions that we need to target besides IE. This can further reduce the bundle size.

animesh1987 (approved):

packages/core/src/app/polyfill.ts
davidchin: I did some further research and realised that we used to use `babel-env` to target the last two versions of browsers, as well as IE11. When we switched over to `esbuild`, we no longer had that configuration.

With that in mind, I think a better approach would be to include only the polyfills necessary for the targeted browsers, rather than importing `core-js/stable` in its entirety. Unfortunately, `esbuild` doesn’t have a feature equivalent to `babel-env`, but I could work around this by creating a simple generator script that achieves a similar result.

packages/workspace-tools/src/generators/polyfill-generator/generator.ts
```diff
@@ -0,0 +1,43 @@
...
+export interface PolyfillGeneratorOptions {
+    projectName: string;
+    browserslistQuery?: string;
+    outputPath?: string;
+}
+
+export default function polyfillGenerator(tree: Tree, options: PolyfillGeneratorOptions) {
+    const {
+        projectName,
+        browserslistQuery = 'defaults and fully supports es6',
```
davidchin, line 16: `defaults` is a shorthand for `0.5%, last 2 versions, Firefox ESR, not dead`

animesh1987 (approved):

tsconfig.base.json
```diff
@@ -114,6 +114,6 @@
     "strict": true,
     "strictNullChecks": true,
     "stripInternal": true,
-    "target": "es5"
+    "target": "es2015"
```
davidchin, line 117: For context, this is to align with ESBuild setting: https://github.com/bigcommerce/checkout-js/blob/151df8c24876e83e3a14e05d9ab1ba1bacb35265/webpack.config.js#L197

bc-launchbay:
success [20250813044637](https://launchbay.bigcommerce.net/projects/535/releases/261476) by Launchbay

PR #2501, opened 2025-08-14
CHECKOUT-9432: Lazy load payment method & checkout button components to reduce bundle size

davidchin:
## What/Why?

Optimized bundle size by implementing lazy loading for payment method and checkout button components. Previously, `checkout-js` bundled all components for every payment provider regardless of whether they were enabled by the merchant, leading to unnecessary bloat and slower performance.

To enable lazy loading, a custom build script was added that extracts the payment method each component resolves to at build time, allowing them to be dynamically loaded only when needed at runtime.

To reduce risk, I’ve wrapped the change in an experiment. As a result, there won’t be any performance benefits until the experiment is removed, since the old code path is still being bundled. I plan to remove the experiment one week after the change is released.

## Rollout/Rollback

Revert or turn off `CHECKOUT-9432.lazy_load_payment_components` experiment.

### Before

Screenshot 2025-08-14 at 4 06 34 pm

### After

Screenshot 2025-08-15 at 11 36 00 am

Screenshot 2025-08-14 at 4 05 48 pm

### DIFF
--- a/packages/core/src/app/customer/CheckoutButtonContainer.tsx
+++ b/packages/core/src/app/customer/CheckoutButtonContainer.tsx
@@ -1,12 +1,13 @@
 import { CheckoutSelectors, CheckoutService } from '@bigcommerce/checkout-sdk';
 import classNames from 'classnames';
-import React, { FunctionComponent, memo } from 'react';
+import React, { FunctionComponent, memo, Suspense } from 'react';
 
 import { TranslatedString, useLocale } from '@bigcommerce/checkout/locale';
 import { CheckoutContextProps } from '@bigcommerce/checkout/payment-integration-api';
 import { useThemeContext, WalletButtonsContainerSkeleton } from '@bigcommerce/checkout/ui';
 
 import { withCheckout } from '../checkout';
+import { isExperimentEnabled } from '../common/utility';
 
 import { getSupportedMethodIds } from './getSupportedMethods';
 import resolveCheckoutButton from './resolveCheckoutButton';
@@ -54,12 +55,14 @@ const CheckoutButtonContainer: FunctionComponent<CheckoutButtonContainerProps & 
         return null;
     }
 
+    const { getConfig } = checkoutState.data;
+
     const renderButtons = () => availableMethodIds.map((methodId) => {
         if (isPaymentStepActive && isPayPalCommerce(methodId)) {
             return null;
         }
 
-        const ResolvedCheckoutButton = resolveCheckoutButton({ id: methodId });
+        const ResolvedCheckoutButton = resolveCheckoutButton({ id: methodId }, isExperimentEnabled(getConfig()?.checkoutSettings, 'CHECKOUT-9432.lazy_load_payment_components'));
 
         if (!ResolvedCheckoutButton) {
             return <CheckoutButtonV1Resolver
@@ -73,16 +76,17 @@ const CheckoutButtonContainer: FunctionComponent<CheckoutButtonContainerProps & 
             />
         }
 
-        return <ResolvedCheckoutButton
-                    checkoutService={checkoutService}
-                    checkoutState={checkoutState}
-                    containerId={`${methodId}CheckoutButton`}
-                    key={methodId}
-                    language={language}
-                    methodId={methodId}
-                    onUnhandledError={onUnhandledError}
-                    onWalletButtonClick={onWalletButtonClick}
-                />;
+        return <Suspense key={methodId}> 
+            <ResolvedCheckoutButton
+                checkoutService={checkoutService}
+                checkoutState={checkoutState}
+                containerId={`${methodId}CheckoutButton`}
+                language={language}
+                methodId={methodId}
+                onUnhandledError={onUnhandledError}
+                onWalletButtonClick={onWalletButtonClick}
+            />
+        </Suspense>;
     });
 
     return (
--- a/packages/core/src/app/customer/CheckoutButtonContainer.tsx
+++ b/packages/core/src/app/customer/CheckoutButtonContainer.tsx
@@ -62,7 +62,10 @@ const CheckoutButtonContainer: FunctionComponent<CheckoutButtonContainerProps & 
             return null;
         }
 
-        const ResolvedCheckoutButton = resolveCheckoutButton({ id: methodId }, isExperimentEnabled(getConfig()?.checkoutSettings, 'CHECKOUT-9432.lazy_load_payment_components'));
+        const ResolvedCheckoutButton = resolveCheckoutButton(
+            { id: methodId },
+            isExperimentEnabled(getConfig()?.checkoutSettings, 'CHECKOUT-9432.lazy_load_payment_components', false)
+        );
 
         if (!ResolvedCheckoutButton) {
             return <CheckoutButtonV1Resolver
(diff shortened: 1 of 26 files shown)


packages/core/src/app/customer/CheckoutButtonContainer.tsx
```diff
@@ -54,12 +55,17 @@ const CheckoutButtonContainer: FunctionComponent<CheckoutButtonContainerProps &
...
+
     const renderButtons = () => availableMethodIds.map((methodId) => {
         if (isPaymentStepActive && isPayPalCommerce(methodId)) {
             return null;
         }
 
-        const ResolvedCheckoutButton = resolveCheckoutButton({ id: methodId });
+        const ResolvedCheckoutButton = resolveCheckoutButton(
+            { id: methodId },
+            isExperimentEnabled(getConfig()?.checkoutSettings, 'CHECKOUT-9432.lazy_load_payment_components', false)
```
animesh1987, line 67: Should the fallback value be true here?

packages/core/src/app/customer/CheckoutButtonContainer.tsx
```diff
@@ -73,16 +79,17 @@ const CheckoutButtonContainer: FunctionComponent<CheckoutButtonContainerProps &
...
-                    checkoutService={checkoutService}
-                    checkoutState={checkoutState}
-                    containerId={`${methodId}CheckoutButton`}
-                    key={methodId}
-                    language={language}
-                    methodId={methodId}
-                    onUnhandledError={onUnhandledError}
-                    onWalletButtonClick={onWalletButtonClick}
-                />;
+        return <Suspense key={methodId}>
```
animesh1987, line 82: Do we need to have a fallback loading state here or is it fine?

packages/core/src/app/customer/CheckoutButtonContainer.tsx
davidchin: The fallback value is set to `false` for the following reasons:
* Checkout microapp deployments are now decoupled from BCApp, but experiment flags are still released through BCApp. This means we can’t merge this PR until the relevant experiment flag has been released to all tiers.
* I plan to remove the experiment and the old code path in a week, so the impact on custom checkout after the experiment is removed will be minimal.

We should have a separate discussion on how to propagate experiment values going forward. Coupling them with BCApp causes the first problem mentioned above. Ideally, we’d be able to query the LD experiment directly on the client side. We need to investigate whether this is possible and what the performance impact would be.

packages/core/src/app/customer/CheckoutButtonContainer.tsx
davidchin: I considered adding a loading state, but decided not to overcomplicate things at this stage since the lazy-loaded component is relatively small. We can always revisit if it ends up impacting UX.

packages/core/src/app/customer/CheckoutButtonContainer.tsx
animesh1987: Yep since we will need to have a follow up PR to `true`, once the bcapp PR reaches T3 in this case I suppose.  Or for experiment wrapped changes we have to wait bcapp code to reach T3?

davidchin: Yes, once BCApp PR reaches T3 and experiment is 100%, we can verify that the change is safe and we can proceed to remove the experiment and the old code paths.

packages/workspace-tools/src/generators/auto-export/auto-export.ts
```diff
@@ -31,6 +32,13 @@ export default async function autoExport({
         });
     }
 
+    if (useLazyLoad) {
+        const componentRegistry = await createComponentRegistryExport(filePaths, memberPattern);
+        const lazyExports = await createLazyLoadingExports(filePaths, tsConfigPath, memberPattern);
+
+        return [lazyExports, componentRegistry].filter(Boolean).join('\n\n');
```
davidchin, line 39: The output looks something like below:

```ts
import { lazy } from 'react';
import { PaymentMethodId } from '@bigcommerce/checkout/payment-integration-api';

const AdyenV2PaymentMethod = lazy(() => import(/* webpackChunkName: "adyen-v2-payment-method" */'@bigcommerce/checkout/adyen-integration').then(module => ({ default: module.AdyenV2PaymentMethod })));
const GooglePayPaymentMethod = lazy(() => import(/* webpackChunkName: "google-pay-payment-method" */'@bigcommerce/checkout/google-pay-integration').then(module => ({ default: module.GooglePayPaymentMethod })));
// ...

export {
  AdyenV2PaymentMethod,
  GooglePayPaymentMethod,
  // ...
};

export const ComponentRegistry = {
  'AdyenV2PaymentMethod': [
    { "gateway": "adyenv2" }
  ],
  'GooglePayPaymentMethod': [
    { "id": PaymentMethodId.AdyenV2GooglePay },
    { "id": PaymentMethodId.AdyenV3GooglePay },
    { "id": PaymentMethodId.AuthorizeNetGooglePay },
    { "id": PaymentMethodId.BNZGooglePay },
    { "id": PaymentMethodId.BraintreeGooglePay },
    { "id": PaymentMethodId.PayPalCommerceGooglePay },
    { "id": PaymentMethodId.BigCommercePaymentsGooglePay },
    { "id": PaymentMethodId.CheckoutcomGooglePay },
    { "id": PaymentMethodId.CybersourceV2GooglePay },
    { "id": PaymentMethodId.OrbitalGooglePay },
    { "id": PaymentMethodId.StripeGooglePay },
    { "id": PaymentMethodId.StripeUPEGooglePay },
    { "id": PaymentMethodId.WorldpayAccessGooglePay },
    { "id": PaymentMethodId.TdOnlineMartGooglePay },
    { "id": PaymentMethodId.StripeOCSGooglePay }
  ],
  // ...
} as const;


```

packages/core/src/app/payment/paymentMethod/PaymentMethod.tsx
```diff
@@ -53,22 +54,22 @@ const PaymentMethodComponent: FunctionComponent<
     const { method } = props;
 
     if (method.type === PaymentMethodProviderType.PPSDK) {
-        return <PPSDKPaymentMethod {...props} />;
+        return <Suspense><PPSDKPaymentMethod {...props} /></Suspense>;
```
bc-peng, line 57: I'm seeing `<Suspense />` is being used to wrap a single component without any prop in this PR.
Could you please explain the expected benefits of this?

packages/core/src/app/customer/CheckoutButtonContainer.tsx
bc-peng: `isExperimentEnabled` is designed to handle situations where the experiment value is `undefined`. Here, we want different behaviours when the experiment is `undefined`, so we are not seeing the usual `true`.

> Ideally, we’d be able to query the LD experiment directly on the client side. 

As DC mentioned above, this solution is ideal, it will allow us to read newly set experiment values without touching BCapp code.

packages/core/src/app/payment/paymentMethod/PaymentMethod.tsx
davidchin: These are lazy components so they need to be wrapped inside `Suspense` otherwise React throws an error.

bc-peng (approved):

bc-launchbay:
success [20250820042424](https://launchbay.bigcommerce.net/projects/535/releases/261916) by Launchbay

bc-launchbay:
success [20250820042424](https://launchbay.bigcommerce.net/projects/535/releases/261916) by Launchbay

bc-launchbay:
success [20250820042424](https://launchbay.bigcommerce.net/projects/535/releases/261916) by Launchbay

PR #2503, opened 2025-08-15
refactor(checkout): CHECKOUT-9384 Collect ErrorBoundary Usage Data

bc-peng:
## What/Why?
Add tags value to collect `ErrorBoundary` usage data.

We are collecting this to determine whether we can safely remove this class component, because we want to avoid handling an error at the top level with `ErrorBoundary`. 

## Rollout/Rollback
Revert.

## Testing
CI.


bc-launchbay:
success [20250815021209](https://launchbay.bigcommerce.net/projects/535/releases/261662) by Launchbay

PR #2506, opened 2025-08-15
refactor(checkout): CHECKOUT-9384 Convert ShippingAddressForm

bc-peng:
## What/Why?

Convert class component `ShippingAddressForm` into function component.

## Rollout/Rollback

Revert.

## Testing
CI.

### DIFF
--- a/packages/core/src/app/shipping/ShippingAddressForm.tsx
+++ b/packages/core/src/app/shipping/ShippingAddressForm.tsx
@@ -1,30 +1,27 @@
 import {
     type Address,
     type Consignment,
-    type Country,
-    type CustomerAddress,
     type FormField,
 } from '@bigcommerce/checkout-sdk';
-import React, { Component, type ReactNode } from 'react';
+import React, { type ReactElement } from 'react';
 
+import { useCheckout } from '@bigcommerce/checkout/payment-integration-api';
 import { LoadingOverlay } from '@bigcommerce/checkout/ui';
 
 import { AddressForm, AddressSelect, AddressType, isValidCustomerAddress } from '../address';
 import { connectFormik, type ConnectFormikProps } from '../common/form';
+import { EMPTY_ARRAY } from '../common/utility';
 import { Fieldset } from '../ui/form';
 
 import { type SingleShippingFormValues } from './SingleShippingForm';
 
 export interface ShippingAddressFormProps {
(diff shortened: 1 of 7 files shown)


packages/core/src/app/shipping/ShippingAddressForm.tsx
```diff
@@ -117,25 +62,69 @@ class ShippingAddressForm extends Component<
...
+                        setFieldValue={setFieldValue}
+                        shouldShowSaveAddress={shouldShowSaveAddress}
+                    />
+                </LoadingOverlay>
+            )}
+        </Fieldset>
+    );
+};
 
 export default connectFormik(ShippingAddressForm);
```
bc-peng, line 138: As discussed offline, we decided to keep Formik HOC.

animesh1987: Is there an opportunity to refactor the components with props to avoid prop drilling once we have converted throughout. I suppose we can get a lot of information from useCheckout hook's slices.

bc-peng: I have removed 3 props.

Not reducing prop drilling completely because:
1. Some props are expensive. For example, `address`, we may consider calculate it once and put it in a context:

   ```ts
   const shippableItemsCount = getShippableItemsCount(cart);
   const methodId = getShippingMethodId(checkout, config);
   const shouldShowMultiShipping =
       hasMultiShippingEnabled && !methodId && shippableItemsCount > 1;

   const shippingAddress =
       !shouldShowMultiShipping && consignments.length > 1 ? undefined : getShippingAddress();
    ```
2.	Lower-level component isn’t a function component yet, so props are still needed.
For example, `googleMapsApiKey` should be derived in `<AddressForm />`, not here.

animesh1987: Thanks maybe that's a future opportunity as you pointed out that we need to convert the leaf components first.

bc-launchbay:
success [20250826042055](https://launchbay.bigcommerce.net/projects/535/releases/262274) by Launchbay

bc-launchbay:
success [20250826042055](https://launchbay.bigcommerce.net/projects/535/releases/262274) by Launchbay

PR #2514, opened 2025-08-19
feat(checkout): CHECKOUT-7730 Apply styles to checkout page layout

richapatel1510:
## What/Why?

Apply stylistic treatment to make each step standout more for themeV2

- Checkout background colour - #F5F5F5
- Step and order summary background colour - #FFFFFF
- No border for the step and order summary
- Radio button style changes

## Rollout/Rollback

Revert this PR

## Testing

- CI checks
- Manual testing

**BEFORE**

https://github.com/user-attachments/assets/21ab3daf-6015-4cf4-879c-9c2c844f553d

**AFTER**

https://github.com/user-attachments/assets/d96fa453-3297-4fa2-9f66-a611907da037


packages/core/src/scss/themeV2/form/_form.scss
```diff
@@ -30,4 +30,32 @@
     //    color: #3c4043;
     //    font-weight: bold;
     //}
```
animesh1987: Not in this PR, but can this be removed?

bc-peng: L3 can be removed along with the commented code as well.

bc-peng (approved):

PR #2525, opened 2025-08-21
refactor(payment): created HostedCreditCardComponent to be able to use HostedCreditCardPaymentMethod implementation in other packages

serhii-tkachenko:
## What/Why?
Created HostedCreditCardComponent to be able to use HostedCreditCardPaymentMethod implementation in paypal-payments-pro-integration package / and other packages in the future

## Rollout/Rollback
Merge / revert

## Testing
Unit tests
Manual tests
Functional tests run
CI

Screenshot 2025-08-21 at 16 19 36


packages/hosted-credit-card-integration/src/components/HostedCreditCardComponent.test.tsx
```diff
@@ -0,0 +1,355 @@
+import { fireEvent, render, screen, waitFor } from '@testing-library/react';
```
bc-peng: We avoid using `render` from `@testing-library/react` directly.
Please use `renderWithoutWrapper` from `@bigcommerce/checkout/test-utils` for inheirting settings like `legacyRoot`.

packages/hosted-credit-card-integration/src/components/HostedCreditCardComponent.test.tsx
```diff
@@ -0,0 +1,355 @@
...
+        translate: (key: string) => key,
+    },
+    onUnhandledError: jest.fn(),
+    ...overrides,
+});
+
+describe('HostedCreditCardComponent', () => {
+    it('renders HostedCreditCardFieldset with correct ids', () => {
+        render(<HostedCreditCardComponent {...getDefaultProps()} />);
+        expect(screen.getByTestId('fieldset')).toBeInTheDocument();
```
bc-peng:
```suggestion
        render(<HostedCreditCardComponent {...getDefaultProps()} />);
        
        expect(screen.getByTestId('fieldset')).toBeInTheDocument();
```
Kindly add an empty line before `expect` in this file.. Not sure why eslint didn't pick this up.

packages/hosted-credit-card-integration/src/components/HostedCreditCardComponent.tsx
```diff
@@ -0,0 +1,278 @@
...
+    const isInstrumentCardCodeRequiredProp = isInstrumentCardCodeRequiredSelector(checkoutState);
+    const isInstrumentCardNumberRequiredProp =
+        isInstrumentCardNumberRequiredSelector(checkoutState);
+    const {
+        config: { cardCode, showCardHolderName },
+    } = method;
+    const isCardCodeRequired = cardCode || cardCode === null;
+    const isCardHolderNameRequired = showCardHolderName ?? true;
+
+    const getHostedFieldId: (name: string) => string = useCallback(
```
bc-peng, line 43: I see `useCallback` is being used a lot in this component.
Could you please explain are they really improving the performance here?

serhii-tkachenko: I just moved code from parent component to a new one. It was written by you (@bc-peng) previously: 
https://github.com/bigcommerce/checkout-js/commit/37faa664baab546b9ca76d7f59b87d602086ac8b#diff-58199e10437f9f66a68e806fca3e412733f57c56cd1ac9ce4556567f254dc264R45

packages/hosted-credit-card-integration/src/components/HostedCreditCardComponent.test.tsx
serhii-tkachenko: Will do

packages/hosted-credit-card-integration/src/components/HostedCreditCardComponent.tsx
bc-peng: Could you please do a quick to test on if removing some or all `useCallback` has any impact on re-render times of ` <CreditCardPaymentMethodComponent />`? 
As I am working on converting class components into function components, I noticed some `useCallback` usages do not make a difference.

serhii-tkachenko: Hi @bc-peng,
I measured performance of component rendering with and without useCallback methods, and it seems like useCallback doing its job well, since the component with useCallback function usages loads faster then without it. 

With useCallback:
Screenshot 2025-09-02 at 11 15 07
Screenshot 2025-09-02 at 11 15 19


Without useCallback:
Screenshot 2025-09-02 at 11 27 14
Screenshot 2025-09-02 at 11 27 24


So in this case I prefer to left code as it is.

animesh1987: I agree with @bc-peng here, we should remove `useCallback` here. Since in react 19 and above react itself is starting to remove it. So one less thing to worry about.

packages/hosted-credit-card-integration/src/components/HostedCreditCardComponent.tsx
```diff
@@ -0,0 +1,278 @@
...
+        (name) => {
+            return `${compact([method.gateway, method.id]).join('-')}-${name}`;
+        },
+        [method],
+    );
+
+    const getHostedFormOptions: (
```
animesh1987, line 50: Can this method be extracted out to a util? Also I will prefer we remove useCallback here given the dependencies this has.

packages/hosted-credit-card-integration/src/components/HostedCreditCardComponent.tsx
serhii-tkachenko: Peng asked me to measure the performance of the component and you can see that component with useCallback for now is better based on numbers. I think useCallback should be moved in React 19 migration project

packages/hosted-credit-card-integration/src/components/HostedCreditCardComponent.tsx
serhii-tkachenko: I think this refactoring is outside of the main scope of the PR, since the only effort here is to move a part of code from `packages/hosted-credit-card-integration/src/HostedCreditCardPaymentMethod.tsx` as it is, so it can be used in other packages to reduce duplications

serhii-tkachenko: Also, as you can see I have covered component with tests. As you can see tests were not added for this component after moving to RTL

serhii-tkachenko: I am happy to leave refactoring for you and your team, so you can update it in a way you see 😄

packages/hosted-credit-card-integration/src/components/HostedCreditCardComponent.tsx
```diff
@@ -0,0 +1,278 @@
...
+    } = method;
+    const isCardCodeRequired = cardCode || cardCode === null;
+    const isCardHolderNameRequired = showCardHolderName ?? true;
+
+    const getHostedFieldId: (name: string) => string = useCallback(
+        (name) => {
+            return `${compact([method.gateway, method.id]).join('-')}-${name}`;
+        },
+        [method],
+    );
```
bc-peng, line 48: @serhii-tkachenko Thanks for testing the performance diligently.
I think one theoretical improvement that could be made is the following.
```
const hostedFieldIdPrefix = useMemo(
        () => compact([method.gateway, method.id]).join('-'),
        [method],
    );

const getHostedFieldId = (name: string) => `${hostedFieldIdPrefix}-${name}`;
```
The `useCallback` is not actually needed, but `useMemo` for making `hostedFieldIdPrefix` should be helpful.

cc @animesh1987

serhii-tkachenko: Updated 👍

serhii-tkachenko: I would not do this because eslint throws an error:
Screenshot 2025-09-03 at 10 09 32

animesh1987 (approved):

bc-launchbay:
success [20250903140350](https://launchbay.bigcommerce.net/projects/535/releases/262903) by Launchbay

packages/hosted-credit-card-integration/src/components/HostedCreditCardComponent.tsx
bc-peng: The eslint error should be eliminated by removing `getHostedFieldId` from the dependencies as it looks like this.  
```
const getHostedFieldId = (name: string) => `${hostedFieldIdPrefix}-${name}`;
```
However, I understand any dependency update is risky, I am happy to improve this when it comes to react 19 upgrade.

PR #2532, opened 2025-08-25
chore(checkout): CHECKOUT-9460 Update styles from review

animesh1987:
## What/Why?
Update styles for theme v2 based on product/design feedback.

## Rollout/Rollback
- revert this PR

## Testing
- Screenshot
Screenshot 2025-08-25 at 3 47 15 pm


packages/core/src/app/checkout/CheckoutStepHeader.tsx
```diff
@@ -56,13 +56,23 @@ const CheckoutStepHeader: FunctionComponent<CheckoutStepHeaderProps> = ({
...
+            }
+
+            {!themeV2 &&
+                <div
+                    className="stepHeader-body stepHeader-column optimizedCheckout-contentPrimary"
+                    data-test="step-info"
+                >
+                    {!isActive && isComplete && summary}
+                </div>
+            }
```
richapatel1510: What is the difference here? We attach div in older theme regardless of the content? 
Also did we intentionally skip 'body-regular' className?

packages/core/src/scss/themeV2/_themeV2.scss
```diff
@@ -2,6 +2,9 @@
 // THEME V2
 // =============================================================================
 
+// Mixins
+@import "mixins";
+
```
bc-peng, line 7: Apart from Richa’s comment, the rest looks good to me.

packages/core/src/app/checkout/CheckoutStepHeader.tsx
animesh1987: I missed the class, have updated it. The main reason I went with separate elements as I moved the `!isActive && isComplete` to top level as well. Since in present OOPC we render an empty element 
```
<div
    className="stepHeader-body stepHeader-column optimizedCheckout-contentPrimary"
    data-test="step-info">
</div>
```

which causes elements to not be centered in the parent div.

bc-launchbay:
success [20250825115050](https://launchbay.bigcommerce.net/projects/535/releases/262198) by Launchbay

PR #2540, opened 2025-08-27
feat(checkout): CHECKOUT-9454 Convert AddressForm

richapatel1510:
## What/Why?

Convert class component `AddressForm` into function component.

## Rollout/Rollback

Revert this PR.

## Testing

- CI checks

### DIFF
--- a/packages/core/src/app/address/AddressForm.tsx
+++ b/packages/core/src/app/address/AddressForm.tsx
@@ -40,9 +40,9 @@ const AddressForm: React.FC<AddressFormProps> = ({
     const config = getConfig();
     const countries = (type === AddressType.Billing 
         ? getBillingCountries() 
-        : getShippingCountries()) 
-        || EMPTY_ARRAY;
-        const googleMapsApiKey = config?.checkoutSettings.googleMapsApiKey || '';
+        : getShippingCountries()
+    ) || EMPTY_ARRAY;
+    const googleMapsApiKey = config?.checkoutSettings.googleMapsApiKey || '';
     const isFloatingLabelEnabledValue = config ? isFloatingLabelEnabled(config.checkoutSettings) : false;
     const countriesWithAutocomplete = ['US', 'CA', 'AU', 'NZ', 'GB'];
 


packages/core/src/app/address/AddressForm.tsx
```diff
@@ -1,214 +1,74 @@
...
+        googleMapsApiKey,
+        onAutocompleteToggle,
+        shouldShowSaveAddress,
+        isFloatingLabelEnabled,
+        countries,
+        setFieldValue = noop,
+        onChange = noop,
+    }) => {
+    const { language } = useLocale();
+    const context = useContext(ThemeContext);
```
bc-peng: Can we use`useThemeContext();`?

packages/core/src/app/address/AddressForm.tsx
```diff
@@ -224,28 +84,102 @@ class AddressForm extends Component<AddressFormProps & WithLanguageProps> {
...
-            .map((field) => field.name);
+        return translatedPlaceholderId && language.translate(translatedPlaceholderId);
+    }, [language]);
 
-        if (fieldName === AUTOCOMPLETE_FIELD_NAME || dateFormFieldNames.includes(fieldName)) {
-            setFieldValue(fieldName, value);
-        }
+    if (!context) {
+        throw Error('Need to wrap in style context');
+    }
```
bc-peng: Think we should be able to remove this.

packages/core/src/app/address/AddressFormType.ts
```diff
@@ -0,0 +1,54 @@
...
+    countries?: Country[];
+    formFields: FormField[];
+    googleMapsApiKey?: string;
+    shouldShowSaveAddress?: boolean;
+    isFloatingLabelEnabled?: boolean;
+    onAutocompleteSelect?(address: Partial<Address>): void;
+    onAutocompleteToggle?(state: { inputValue: string; isOpen: boolean }): void;
+    onChange?(fieldName: string, value: string | string[]): void;
+    setFieldValue?(fieldName: string, value: string | string[]): void;
+}
```
bc-peng: Thought we should be able to get rid of at least`countries`, `countriesWithAutocomplete`, ` googleMapApiKey` and `isFloatingLableEnabled` by reading them from checkout context directly.
*`countriesWithAutocomplete` should be read as a constant directly.
Can we do this in this PR or in another follow-up PR as part of the effort to reduce prop drilling?

packages/core/src/app/address/AddressForm.tsx
```diff
@@ -1,214 +1,89 @@
...
+            data: { getConfig, getBillingCountries, getShippingCountries }
+        }
+    } = useCheckout();
+    
+    const config = getConfig();
+    const countries = (type === AddressType.Billing 
+        ? getBillingCountries() 
+        : getShippingCountries()) 
+        || EMPTY_ARRAY;
+        const googleMapsApiKey = config?.checkoutSettings.googleMapsApiKey || '';
```
animesh1987: Is this a formatting issue?

packages/core/src/app/address/AddressForm.tsx
```diff
@@ -1,214 +1,89 @@
...
+    const isFloatingLabelEnabledValue = config ? isFloatingLabelEnabled(config.checkoutSettings) : false;
+    const countriesWithAutocomplete = ['US', 'CA', 'AU', 'NZ', 'GB'];
```
bc-peng, line 47: `const countriesWithAutocomplete = ['US', 'CA', 'AU', 'NZ', 'GB']` has a same copy in `BillingForm.tsx`, can we consolidate them into one place one copy?

bc-peng (approved):

packages/core/src/app/address/AddressForm.tsx
richapatel1510: Its already removed from `BillingForm.tsx` [here](https://github.com/bigcommerce/checkout-js/pull/2540/files#diff-d6dabe25c48ac22be31a89292cacb4e84b06d8b02060968261de0b2145d31459)

PR #2549, opened 2025-09-01
refactor(checkout): CHECKOUT-9465 Convert Billing

bc-peng:
## What/Why?

Convert class component `Billing` into function component.

## Rollout/Rollback

Revert.

## Testing
- CI.
- Manually tested major use cases.

### DIFF
--- a/packages/core/src/app/billing/Billing.tsx
+++ b/packages/core/src/app/billing/Billing.tsx
@@ -1,4 +1,4 @@
-import type {CheckoutSelectors} from '@bigcommerce/checkout-sdk';
+import type { CheckoutSelectors } from '@bigcommerce/checkout-sdk';
 import React, { type ReactElement, useEffect } from 'react';
 
 import { TranslatedString } from '@bigcommerce/checkout/locale';


packages/core/src/app/billing/BillingForm.tsx
```diff
@@ -101,9 +105,15 @@ const BillingForm = ({
...
     const handleUseNewAddress = () => {
-        handleSelectAddress({});
+        void handleSelectAddress({});
     };
 
+    // Below should be removed once <AddressForm /> is able to reduce prop drilling
+    const countries = getBillingCountries() || EMPTY_ARRAY;
+    const countriesWithAutocomplete = ['US', 'CA', 'AU', 'NZ', 'GB'];
+    const { googleMapsApiKey } = config.checkoutSettings;
+    const isFloatingLabelEnabled = getIsFloatingLabelEnabled(config.checkoutSettings)
```
bc-peng, line 115: @richapatel1510 Something about the `<AddressForm />`.

packages/core/src/app/billing/Billing.tsx
```diff
@@ -1,100 +1,64 @@
...
-            updateCheckout,
-            customerMessage,
-            billingAddress,
-            navigateNextStep,
-            onUnhandledError,
-        } = this.props;
-
+    const isInitializing  = isLoadingBillingCountries();
+    const themeV2  = isExperimentEnabled(config.checkoutSettings,
+        'CHECKOUT-7962.update_font_style_on_checkout_page');
```
animesh1987: Why are we reading this from setting, shouldn't this be read from the context?

bc-peng: I got it wrong. Yes, I should be able to get it from the context.

packages/core/src/app/billing/Billing.tsx
```diff
@@ -1,100 +1,62 @@
...
-                    <BillingForm
-                        {...props}
-                        onSubmit={this.handleSubmit}
-                        updateAddress={updateAddress}
-                    />
-                </div>
-            </AddressFormSkeleton>
-        );
+    if (!config || !customer || !checkout || !cart) {
+        throw new Error('Unable to access checkout data')
```
serhii-tkachenko, line 43: Should the message be translated, so the end user can understand what is going on? (I understand that this is a rare error)

packages/core/src/app/billing/Billing.tsx
```diff
@@ -1,100 +1,62 @@
...
-    type Address,
-    type CheckoutRequestBody,
-    type CheckoutSelectors,
-    type Country,
-    type Customer,
-    type FormField,
-} from '@bigcommerce/checkout-sdk';
-import { noop } from 'lodash';
-import React, { Component, type ReactNode } from 'react';
+import type {CheckoutSelectors} from '@bigcommerce/checkout-sdk';
```
serhii-tkachenko:
```suggestion
import type { CheckoutSelectors } from '@bigcommerce/checkout-sdk';
```

packages/core/src/app/billing/BillingForm.tsx
```diff
@@ -33,43 +33,44 @@ import StaticBillingAddress from './StaticBillingAddress';
...
+    const {
+        data: { getCustomer, getConfig, getBillingCountries, getCart },
+        statuses: { isUpdatingBillingAddress, isUpdatingCheckout },
+    } = checkoutState;
+    const customer = getCustomer();
+    const config = getConfig();
+    const cart = getCart();
+
+    if (!config || !customer || !cart) {
+        throw new Error('checkout data is not available');
```
serhii-tkachenko, line 69: Same here

serhii-tkachenko (commented):
I have several questions regarding this refactoring:
1. Should these refactoring be coverer with experiment, so it can be turned off if there any issues on production?
2. Do you have automation tests which checks if billing address form works as expected?

packages/core/src/app/billing/Billing.tsx
animesh1987: These errors are the errors in console I believe.

animesh1987 (approved):

bc-peng:
> 1. Should these refactoring be coverer with experiment, so it can be turned off if there any issues on production?

We can roll back quickly by using [this experiment](https://app.launchdarkly.com/projects/default/flags/CHECKOUT-9275.checkout_app_version_override/targeting?env=production&selected-env=production) now. 

>Do you have automation tests which checks if billing address form works as expected?

As a benefit of migrating to RTL earlier this year, our existing tests coverage is good enough to cover the refactored the component. Refactored component should not change the end-user behaviour/experience in any way.

packages/core/src/app/billing/Billing.tsx
bc-peng: Throwing an error here is simply for making TS happy.
In theory, as soon as we stop showing the checkout loading skeleton, all these values are defined.
However, we historically allow them to be `undefined`, hence we need this check quite often.

richapatel1510 (approved):

bc-launchbay:
success [20250903010707](https://launchbay.bigcommerce.net/projects/535/releases/262853) by Launchbay

bc-launchbay:
success [20250903010707](https://launchbay.bigcommerce.net/projects/535/releases/262853) by Launchbay

PR #2553, opened 2025-09-03
feat(checkout): CHECKOUT-9454 Convert GoogleAutocomplete

richapatel1510:
## What/Why?

Convert class component `GoogleAutocomplete` into function component.

## Rollout/Rollback

Revert this PR.

## Testing

- CI checks

### DIFF
--- a/packages/core/src/app/address/googleAutocomplete/GoogleAutocomplete.tsx
+++ b/packages/core/src/app/address/googleAutocomplete/GoogleAutocomplete.tsx
@@ -1,5 +1,5 @@
 import { noop } from 'lodash';
-import React, { useEffect, useRef, useState } from 'react';
+import React, { useRef, useState } from 'react';
 
 import { Autocomplete, type AutocompleteItem } from '../../ui/autocomplete';
 
@@ -49,10 +49,9 @@ const GoogleAutocomplete: React.FC<GoogleAutocompleteProps> = ({
     const [autoComplete, setAutoComplete] = useState<string>('off');
     const googleAutocompleteServiceRef = useRef<GoogleAutocompleteService>();
 
-    // Initialize the service
-    useEffect(() => {
+    if (!googleAutocompleteServiceRef.current) {
         googleAutocompleteServiceRef.current = new GoogleAutocompleteService(apiKey);
-    }, [apiKey]);
+    }
 
     const onSelectHandler = (item: AutocompleteItem) => {
         const service = googleAutocompleteServiceRef.current;


packages/core/src/app/address/googleAutocomplete/GoogleAutocomplete.tsx
```diff
@@ -21,53 +21,46 @@ export interface GoogleAutocompleteProps {
...
+    nextElement,
+    isAutocompleteEnabled,
+    onChange = noop,
+    componentRestrictions,
+    types,
+    apiKey,
+}) => {
+    const [items, setItems] = useState<AutocompleteItem[]>([]);
+    const [autoComplete, setAutoComplete] = useState<string>('off');
+    const googleAutocompleteServiceRef = useRef<GoogleAutocompleteService>();
```
animesh1987, line 50: Do we need to create a ref here? Feel we can just create an instance when the component mounts?

packages/core/src/app/address/googleAutocomplete/GoogleAutocomplete.tsx
```diff
@@ -21,53 +21,46 @@ export interface GoogleAutocompleteProps {
...
+
+    // Initialize the service
+    useEffect(() => {
```
animesh1987: Why do we need to have an effect? Since `apiKey` should not change for an instance we can instantiate it at the useRef declaration itself if that's the implementation we want to go with. Also we can check with an if statement if service exists and create it only then.

packages/core/src/app/address/googleAutocomplete/GoogleAutocomplete.tsx
animesh1987: Ignore this comment.

packages/core/src/app/address/googleAutocomplete/GoogleAutocomplete.tsx
```diff
@@ -21,53 +21,46 @@ export interface GoogleAutocompleteProps {
...
+    onChange = noop,
+    componentRestrictions,
+    types,
+    apiKey,
+}) => {
+    const [items, setItems] = useState<AutocompleteItem[]>([]);
+    const [autoComplete, setAutoComplete] = useState<string>('off');
+    const googleAutocompleteServiceRef = useRef<GoogleAutocompleteService>();
+
+    // Initialize the service
```
bc-peng:
```suggestion
```

animesh1987 (approved):

PR #2578, opened 2025-09-16
refactor(checkout): CHECKOUT-9475 Update eslint rule

bc-peng:
## What/Why?

Update the eslint rule to enforce spaces inside curly braces, improving code readability and consistency.

## Rollout/Rollback

Revert.

## Testing

CI.

### DIFF
--- a/packages/core/src/app/billing/Billing.tsx
+++ b/packages/core/src/app/billing/Billing.tsx
@@ -19,7 +19,7 @@ interface BillingProps {
     onUnhandledError(error: Error): void;
 }
 
-const Billing = ({navigateNextStep, onReady, onUnhandledError}:BillingProps): ReactElement => {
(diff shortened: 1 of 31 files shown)


packages/core/src/app/billing/Billing.tsx
```diff
@@ -19,7 +19,7 @@ interface BillingProps {
     onUnhandledError(error: Error): void;
 }
 
-const Billing = ({navigateNextStep, onReady, onUnhandledError}:BillingProps): ReactElement => {
+const Billing = ({ navigateNextStep, onReady, onUnhandledError }:BillingProps): ReactElement => {
```
bc-peng, line 22: As a side effect, the spacing rule also applies to braces outside of imports. 😂

animesh1987 (approved):

PR #2580, opened 2025-09-17
feat(checkout): CHECKOUT-9428 Remove Scroll Bars from Error Modal

bc-peng:
## What/Why?

Remove unnecessary scroll bars from the checkout page's generic error modal.

Please note: in MacOS, scroll bars only always show up when "Show scroll bars" is set to "Always".

image

## Rollout/Rollback

CI.

#### Before
Screenshot 2025-09-17 at 14 48 32

#### After
Screenshot 2025-09-17 at 14 48 08

### DIFF
--- a/packages/core/src/scss/components/foundation/modal/_modal.scss
+++ b/packages/core/src/scss/components/foundation/modal/_modal.scss
@@ -174,7 +174,7 @@
 // Specific styles for displaying cart summary on mobile
 body.has-activeModal {
     .modal-body {
-        overflow: scroll;
+        overflow: auto;
         padding-bottom: 20%; // This is to fill the space occupied by Safari and Chrome's navigation bar on mobile
 
         @include breakpoint("small") {


packages/core/src/scss/components/foundation/modal/_modal.scss
```diff
@@ -174,7 +174,7 @@
 // Specific styles for displaying cart summary on mobile
 body.has-activeModal {
     .modal-body {
-        overflow: scroll;
+        overflow: auto;
```
bc-peng, line 177: [overflow](https://www.w3schools.com/cssref/pr_pos_overflow.php) has four properties.

> `visible` - Default. The overflow is not clipped. The content renders outside the element's box
> `hidden` - The overflow is clipped, and the rest of the content is hidden
> `scroll` - Scrollbars are added. User must scroll to see all content
> `auto` - Similar to scroll, but adds scrollbars only when necessary

`hidden` will cut the message text, `visible` will break the page layout, `scroll` looks redundant. 

`auto` is ideal, however, as the error modal always auto resizes itself, scroll bars should not be seen.

bc-launchbay:
success [20250917054040](https://launchbay.bigcommerce.net/projects/535/releases/263875) by Launchbay

bc-launchbay:
success [20250917054040](https://launchbay.bigcommerce.net/projects/535/releases/263875) by Launchbay

PR #2582, opened 2025-09-17
feat(checkout): CHECKOUT-9468 convert SingleShippingForm

richapatel1510:
## What/Why?

Convert class component `SingleShippingForm` into function component.

## Rollout/Rollback

Revert this PR.

## Testing

- CI checks


packages/core/src/app/shipping/SingleShippingForm.tsx
```diff
@@ -119,231 +123,190 @@ class SingleShippingForm extends PureComponent<
...
+                shippingFormRenderTimestamp={shippingFormRenderTimestamp}
+                shouldDisableSubmit={shouldDisableSubmit()}
+                shouldShowOrderComments={shouldShowOrderComments}
+                shouldShowShippingOptions={isValid}
+            />
+        </Form>
+    );
+};
 
 export default withLanguage(
```
animesh1987, line 297: We should be able to get rid of withLanguage HOC?

packages/core/src/app/shipping/SingleShippingForm.tsx
```diff
@@ -119,231 +123,190 @@ class SingleShippingForm extends PureComponent<
...
-        } finally {
-            this.setState({ isResettingAddress: false });
         }
-    };
-
-    private onUseNewAddress: () => void = async () => {
-        const { deleteConsignments, onUnhandledError = noop, setValues, values } = this.props;
+    }, [shippingFormRenderTimestamp]);
+
+    const handleAddressSelect = useCallback(
```
animesh1987: I think we can get rid of `useCallback` here.

packages/core/src/app/shipping/SingleShippingForm.tsx
```diff
@@ -119,231 +123,190 @@ class SingleShippingForm extends PureComponent<
...
-                    shouldShowOrderComments={shouldShowOrderComments}
-                    shouldShowShippingOptions={isValid}
-                />
-            </Form>
-        );
-    }
+        return () => {
+            debouncedUpdateAddressRef.current?.cancel();
+        };
+    }, [updateAddress, shippingAutosaveDelay]);
```
bc-peng:
```suggestion
    }, []);
```
We only want it to run once, don't we?

packages/core/src/app/shipping/SingleShippingForm.tsx
```diff
@@ -119,231 +123,190 @@ class SingleShippingForm extends PureComponent<
...
-        const { updateAddress, onUnhandledError = noop, values, setValues } = this.props;
-
-        this.setState({ isResettingAddress: true });
-
-        try {
-            await updateAddress(address);
-
+    }, [
+        getFields,
+        setFieldValue,
```
bc-peng: `setFieldValue` will change `values`.
Is it possible that this dependency array causes an infinite `useEffect` call?

packages/core/src/app/shipping/SingleShippingForm.tsx
richapatel1510: We use language for validation schema of formik in withFormikExtended, hence need to keep it here.

packages/core/src/app/shipping/SingleShippingForm.tsx
richapatel1510: Moved this to the mount useEffect, we only need to set this once on mount.

PR #2596, opened 2025-09-23

bc-peng:
## What/Why?

Convert class component `Payment` into function component.

## Rollout/Rollback

Revert.

## Testing
CI.

#### Stripe CC

https://github.com/user-attachments/assets/25bef230-b0a2-4527-9150-118287e2af06

### DIFF
--- a/packages/core/src/app/payment/Payment/Payment.tsx
+++ b/packages/core/src/app/payment/Payment/Payment.tsx
@@ -4,7 +4,7 @@ import {
     type PaymentMethod,
 } from '@bigcommerce/checkout-sdk';
 import { isEmpty, noop } from 'lodash';
-import React, { type ReactElement, useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import React, { type ReactNode, useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { type ObjectSchema } from 'yup';
 
 import { useAnalytics } from '@bigcommerce/checkout/analytics';
@@ -59,7 +59,7 @@ const Payment = (
         onUnhandledError = noop,
         checkEmbeddedSupport = noop,
     }: PaymentProps,
-): ReactElement => {
+): ReactNode => {
     const [didExceedSpamLimit, setDidExceedSpamLimit] = useState(false);
     const [isReady, setIsReady] = useState(false);
     const [selectedMethod, setSelectedMethodState] = useState<PaymentMethod | undefined>();
@@ -105,8 +105,8 @@ const Payment = (
     const paymentProviderCustomer = getPaymentProviderCustomer();
     const { isComplete = false } = getOrder() || {};
 
-    if(!checkout || !customer || !config || !consignments || isComplete) {
-        throw new Error('Checkout data is not available or order is complete');
+    if (!checkout || !customer || !config || !consignments) {
+        throw new Error('Checkout data is not available');
     }
 
     const paymentMethods = getPaymentMethods();
@@ -394,6 +394,10 @@ const Payment = (
 
         checkEmbeddedSupport(methods.map(({ id }) => id));
     }, [methods]);
+
+    if (isComplete) {
+        return null;
+    }
 
     return (
         <PaymentContext.Provider value={paymentContextValue}>


packages/core/src/app/payment/Payment/Payment.tsx
```diff
@@ -0,0 +1,432 @@
...
+    if(!checkout || !customer || !config) {
+        throw new Error('Checkout data is not available');
+    }
+
+    const { isStoreCreditApplied, shouldExecuteSpamCheck } = checkout;
+    const applyStoreCredit = checkoutService.applyStoreCredit;
+    const availableStoreCredit = customer.storeCredit;
+    const cartUrl = config.links.cartLink;
+    const paymentMethods = getPaymentMethods();
+    const { defaultMethod, methods } = useMemo(()=>getDefaultPaymentMethodAndFilteredMethods(checkoutState, paymentMethods),[paymentMethods]);
```
bc-peng: Calculating the default method and filtering payment methods turned out to be very expensive.
They were taking up most of the cost in `mapToPaymentProps`. I’ve moved them out and added `useMemo`.
I'm happy to break `checkoutState` into the required objects, just want to confirm this structure first.

packages/core/src/app/payment/Payment/Payment.tsx
```diff
@@ -0,0 +1,432 @@
...
+    const {
+        enableTermsAndConditions,
+        orderTermsAndConditions,
+        orderTermsAndConditionsType: termsConditionsType,
+        orderTermsAndConditionsLink: termsCondtitionsUrl,
+    } = config.checkoutSettings as CheckoutSettings & { orderTermsAndConditionsLocation: string };
+    const isTermsConditionsRequired = enableTermsAndConditions;
+    const shouldLocaliseErrorMessages = isExperimentEnabled(
+        config.checkoutSettings,
+        'PAYMENTS-6799.localise_checkout_payment_error_messages',
```
serhii-tkachenko: This is outdated experiment that was added in 2021. Also I can't find it in LD. Do we still need to keep it here?

packages/core/src/app/payment/Payment/getDefaultPaymentMethodAndFilteredMethods.ts
```diff
@@ -0,0 +1,98 @@
...
+    let filteredMethods;
+
+    filteredMethods = methods.filter((method: PaymentMethod) => {
+        if (method.id === PaymentMethodId.Bolt && method.initializationData) {
+            return Boolean(method.initializationData.showInCheckout);
+        }
+
+        return method.id !== PaymentMethodId.BraintreeLocalPaymentMethod;
+
+
```
serhii-tkachenko: Please remove these empty lines

packages/core/src/app/payment/Payment/getDefaultPaymentMethodAndFilteredMethods.ts
```diff
@@ -0,0 +1,98 @@
...
+        const stripeUpePaymentMethod = methods.filter(method =>
+            method.id === 'card' && method.gateway === PaymentMethodId.StripeUPE
+        );
+
+        methods = stripeUpePaymentMethod.length ? stripeUpePaymentMethod : methods;
+    }
+
+    if (!checkout || !config || !customer || isComplete) {
+        throw new Error('checkout data is not available or order is complete');
+    }
```
serhii-tkachenko: It might be better to move this block and place it before Stripe related condition, to avoid unnecessary code to be triggered when major data is not available

packages/core/src/app/payment/Payment/Payment.tsx
bc-peng: Removed this const variable and all of its usage.

packages/core/src/app/payment/Payment/getDefaultPaymentMethodAndFilteredMethods.ts
bc-peng: Removed, also used regex to check all code.

packages/core/src/app/payment/Payment/getDefaultPaymentMethodAndFilteredMethods.ts
bc-peng: Yes, it is best to have major data ready/check first.

bc-peng: On a second thought, I have moved `if (!checkout || !config || !customer || isComplete) {` out to Payment.tsx as part of replacing CheckoutSelectors with required objects.

packages/test-framework/src/react-testing-library-support/CheckoutPageNodeObject.ts
```diff
@@ -269,7 +269,7 @@ export class CheckoutPageNodeObject {
     }
 
     async waitForPaymentStep(): Promise<void> {
-        await waitFor(() => screen.getByText(/place order/i));
+        await waitFor(() => screen.getByText(/place order/i), { timeout: 20000 });
```
bc-peng, line 272: Just for the record:
The default timeout here is `1000 ms`, which causes `Payment.test.tsx` to fail unpredictably, so I’m updating it.

animesh1987 (approved):

packages/core/src/app/payment/Payment/OrderErrorModal.tsx
```diff
@@ -0,0 +1,48 @@
...
+    onClose,
+}: OrderErrorModalProps): ReactElement | null => {
+    const error = submitOrderError || finalizeOrderError;
+
+    if (!error || ('type' in error &&(
+        error.type === 'order_finalization_not_required' ||
+        error.type === 'payment_cancelled' ||
+        error.type === 'payment_invalid_form' ||
+        error.type === 'spam_protection_not_completed' ||
+        error.type === 'invalid_hosted_form_value'
```
richapatel1510, line 30: Should we have enum for these types?

bc-peng: As they are only being used in this file, think we can keep them.

richapatel1510 (dismissed):

bc-launchbay:
success [20251001011313](https://launchbay.bigcommerce.net/projects/535/releases/264923) by Launchbay

PR #2602, opened 2025-09-29
feat(checkout): CHECKOUT-9477 Convert Shipping component (Part 1)

richapatel1510:
## What/Why?

Convert class component `Shipping` into function component.

## Rollout/Rollback

Revert.

## Testing
CI.

### DIFF
--- a/packages/core/src/app/shipping/Shipping.tsx
+++ b/packages/core/src/app/shipping/Shipping.tsx
@@ -14,7 +14,7 @@ import {
     type ShippingRequestOptions,
 } from '@bigcommerce/checkout-sdk';
 import { noop } from 'lodash';
-import React, { Component, type ReactNode } from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import { createSelector } from 'reselect';
 
 import { type ExtensionContextProps, withExtension } from '@bigcommerce/checkout/checkout-extension';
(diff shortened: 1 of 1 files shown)


packages/core/src/app/shipping/Shipping.tsx
```diff
@@ -309,7 +213,79 @@ class Shipping extends Component<ShippingProps & WithCheckoutShippingProps & Ext
...
+                    shouldShowCloseButton={false}
+                />
+                <ShippingHeader
+                    cartHasPromotionalItems={cartHasPromotionalItems}
+                    isMultiShippingMode={isMultiShippingMode}
+                    onMultiShippingChange={handleMultiShippingModeSwitch}
+                    shouldShowMultiShipping={shouldShowMultiShipping}
+                />
+                <ShippingForm
+                    {...shippingFormProps}
```
bc-peng, line 271: Think we should avoid using a combined prop like `{...shippingFormProps}` whenever possible, as it reduces code readability and maintainability. 

Could we list out all the props explicitly here and remove `{...shippingFormProps}`?

packages/core/src/app/shipping/Shipping.tsx
```diff
@@ -309,7 +213,79 @@ class Shipping extends Component<ShippingProps & WithCheckoutShippingProps & Ext
                 onUnhandledError(error);
             }
         }
-    };
+    }, [props]);
```
bc-peng, line 214: I see we are using `[props]` 3 times in this file.
We may want to revisit this dependency once we are able to update `props: ShippingProps & WithCheckoutShippingProps & ExtensionContextProps`.

packages/core/src/app/shipping/Shipping.tsx
```diff
@@ -88,161 +88,75 @@ export interface WithCheckoutShippingProps {
...
 
-    private handleSingleShippingSubmit: (values: SingleShippingFormValues) => void = async ({
-        billingSameAsShipping,
-        shippingAddress: addressValues,
-        orderComment,
-    }) => {
+    const handleSwitchToSingleShipping = useCallback(async () => {
+        setIsMultiShippingUnavailableModalOpen(false);
+        await handleMultiShippingModeSwitch();
+    }, [handleMultiShippingModeSwitch]);
```
bc-peng: Is it possible that `handleMultiShippingModeSwitch` will not be changed during the life time of this component? So we can use `[]` here.

packages/core/src/app/shipping/Shipping.tsx
```diff
@@ -88,161 +88,75 @@ export interface WithCheckoutShippingProps {
...
-            isMultiShippingMode,
-            step,
-            shouldRenderStripeForm,
-            cartHasPromotionalItems,
-            extensionState: { shippingFormRenderTimestamp } = {},
-            setIsMultishippingMode,
-            ...shippingFormProps
-        } = this.props;
+        return () => { isMounted = false; };
+    }, [loadShippingAddressFields, loadShippingOptions, loadBillingAddressFields, onReady, onUnhandledError, cartHasPromotionalItems, isMultiShippingMode]);
```
bc-peng: Is this `useEffect` intended to run once?

packages/core/src/app/shipping/Shipping.tsx
```diff
@@ -88,161 +88,75 @@ export interface WithCheckoutShippingProps {
...
-            deinitializeShippingMethod,
-            isMultiShippingMode,
-            step,
-            shouldRenderStripeForm,
-            cartHasPromotionalItems,
-            extensionState: { shippingFormRenderTimestamp } = {},
-            setIsMultishippingMode,
-            ...shippingFormProps
-        } = this.props;
+        return () => { isMounted = false; };
```
bc-peng: If the component is unmounted, does the value of `isMounted` matter?

packages/core/src/app/shipping/Shipping.tsx
richapatel1510: This will be gone in follow up PR to get rid of mapToShippingProps

packages/core/src/app/shipping/Shipping.tsx
richapatel1510: This will be gone in follow up PR to get rid of mapToShippingProps

packages/core/src/app/shipping/Shipping.tsx
richapatel1510: Yes, that's right. Updated the code, we don't need `isMounted`

packages/core/src/app/shipping/Shipping.tsx
```diff
@@ -88,161 +88,73 @@ export interface WithCheckoutShippingProps {
...
+
+                if (cartHasPromotionalItems && isMultiShippingMode) {
+                    setIsMultiShippingUnavailableModalOpen(true);
+                }
+
+                onReady();
+            } catch (error) {
+               onUnhandledError(error);
+            } finally {
+                setIsInitializing(false);
```
bc-peng, line 118:
```suggestion
               setIsInitializing(false);
```

bc-peng: This can be handled in the following PR.

animesh1987 (approved):

PR #2606, opened 2025-09-29
fix(payment): added data-test attribute

andriiVitvitskyi1990:
## What/Why?
- Added data-test attribute to fix form visibility issue

We have https://www.zagg.com/ store have custom script that relies on checkout html markup `data-test="credit-cart-payment-method"`. The issue could be reproduced only for logged in customers and caused because custom script didn't find data-test mentioned before.

Screenshot 2025-09-29 at 19 26 00

image

Screenshot 2025-09-29 at 20 07 01

## Rollout/Rollback
Revert

## Testing
Unit


animesh1987 (approved):

bc-launchbay:
success [20250930072249](https://launchbay.bigcommerce.net/projects/535/releases/264843) by Launchbay

PR #2609, opened 2025-09-30
feat(checkout): CHECKOUT-9477 refactor Shipping component by adding useShipping hook

richapatel1510:
## What/Why?

Refactor shipping component to use the new hook `useShipping` instead of `mapToShippingProps` and avoid prop drilling as much as possible.

## Rollout/Rollback

Revert this PR.

## Testing

- CI checks


animesh1987 (dismissed):

packages/core/src/app/shipping/hooks/useShipping.ts
```diff
@@ -0,0 +1,130 @@
...
+        hasMultiShippingEnabled && !methodId && shippableItemsCount > 1;
+
+    const shippingAddress =
+        !shouldShowMultiShipping && consignments.length > 1 ? undefined : getShippingAddress();
+
+    const providerWithCustomCheckout = getProviderWithCustomCheckout(
+        config.checkoutSettings.providerWithCustomCheckout,
+    );
+
+    return {
```
bc-peng, line 97: Things like cart, consignments, isGuest are not related to shipping, can we make `useShipping` for shipping only?

animesh1987: I think all this information is needed for shipping component. And consignments is essentially what drives shipping 🤔

richapatel1510: I prefer to keep it here because if we use 2 different hooks in the component we will have to add conditions such as 
```
if (!checkout || !config || !customer || !cart) {
        throw new Error('Unable to access checkout data');
    }
```

in the component again. And we need some of the cart and consignment data to calculate other flags in this hook.

bc-peng: Oh, make sense. I see the Shipping.tsx is much shorter than before.

bc-peng (approved):

PR #2610, opened 2025-09-30
chore(customer): CHECKOUT-9474 Move guest customer form out of customer component

animesh1987:
## What/Why?
Refactor and create a `GuestFormContainer` to replace the render method.

## Rollout/Rollback
- revert this PR

## Testing
- CI

### DIFF
--- a/packages/core/src/app/customer/GuestFormContainer.tsx
+++ b/packages/core/src/app/customer/GuestFormContainer.tsx
@@ -1,0 +1,132 @@
+import React from 'react';
+import { PaymentMethodId, useCheckout } from '@bigcommerce/checkout/payment-integration-api';
+import CheckoutButtonList from './CheckoutButtonList';
+import GuestForm, { GuestFormValues } from './GuestForm';
+import { isPayPalFastlaneMethod } from '@bigcommerce/checkout/paypal-fastlane-integration';
+import getProviderWithCustomCheckout from '../payment/getProviderWithCustomCheckout';
+import { Cart } from '@bigcommerce/checkout-sdk';
+import { shouldUseStripeLinkByMinimumAmount } from '@bigcommerce/checkout/instrument-utils';
+import StripeGuestForm from './StripeGuestForm';
+import CheckoutStepStatus from '../checkout/CheckoutStepStatus';
+
+interface GuestFormContainerProps {
+    email?: string;
+    isFloatingLabelEnabled?: boolean;
+    isWalletButtonsOnTop: boolean;
+    isSubscribed: boolean;
+    step: CheckoutStepStatus;
+    checkEmbeddedSupport?(methodIds: string[]): void;
+    handleChangeEmail(email: string): void;
+    handleContinueAsGuest(formValues: GuestFormValues): void;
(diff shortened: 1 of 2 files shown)


packages/core/src/app/customer/GuestFormContainer.tsx
```diff
@@ -0,0 +1,137 @@
...
+import React from 'react';
+
+import type CheckoutStepStatus from '../checkout/CheckoutStepStatus';
+import getProviderWithCustomCheckout from '../payment/getProviderWithCustomCheckout';
+
+import CheckoutButtonList from './CheckoutButtonList';
+import GuestForm, { type GuestFormValues } from './GuestForm';
+import StripeGuestForm from './StripeGuestForm';
+
+
```
bc-peng:
```suggestion

```

packages/core/src/app/customer/GuestFormContainer.tsx
```diff
@@ -0,0 +1,137 @@
...
+interface GuestFormContainerProps {
+    email?: string;
+    isFloatingLabelEnabled?: boolean;
```
bc-peng, line 16:
```suggestion
    email: string | undefined;
    isFloatingLabelEnabled: boolean | undefined;
```
This component is only being used at one place, we will always provide these props, right?

packages/core/src/app/customer/GuestFormContainer.tsx
```diff
@@ -0,0 +1,137 @@
...
+        deinitializeCustomer,
+        initializeCustomer,
+    }  = checkoutService;
+
+    const config = getConfig();
+    const cart = getCart();
+    const isLoadingGuestForm = isContinuingAsGuest() || isExecutingPaymentMethodCheckout();
+
+    if (!config || !cart) {
+        return null;
```
bc-peng, line 70: Because we only show `CheckoutPage` component when we have config and cart locally, thought instead of returning `null`, we can throw an error here.
https://github.com/bigcommerce/checkout-js/blob/af87b4084eab2f0fe2aa7f73c3a82c02be47435b/packages/core/src/app/checkout/CheckoutIntermediate.tsx#L34

packages/core/src/app/customer/GuestFormContainer.tsx
animesh1987: Yes only in one place, I prefer the `?` approach in defining interfaces, props are coming from parent so can be undefined still.

packages/core/src/app/customer/GuestFormContainer.tsx
animesh1987: We have a pattern of `null` in other places as well from what I see, so maybe keep this for now?

bc-launchbay:
success [20251001044810](https://launchbay.bigcommerce.net/projects/535/releases/264937) by Launchbay

bc-launchbay:
success [20251001044810](https://launchbay.bigcommerce.net/projects/535/releases/264937) promoted by https://launchbay.bigcommerce.net/projects/535/releases/264955 being deployed. Deployed by Launchbay

PR #2623, opened 2025-10-07
chore(customer): CHECKOUT-9474 Convert customer component to functional component

animesh1987:
## What/Why?
Convert customer component to functional component

## Rollout/Rollback
- revert this PR

## Testing
- CI

### DIFF
--- a/packages/core/src/app/customer/Customer.tsx
+++ b/packages/core/src/app/customer/Customer.tsx
@@ -1,17 +1,6 @@
 import { type AnalyticsContextProps } from '@bigcommerce/checkout/analytics';
-import { shouldUseStripeLinkByMinimumAmount } from '@bigcommerce/checkout/instrument-utils';
-import { type CheckoutContextProps } from '@bigcommerce/checkout/payment-integration-api';
 import {
-    type CheckoutPaymentMethodExecutedOptions,
-    type CheckoutSelectors,
-    type CustomerAccountRequestBody,
     type CustomerCredentials,
-    type CustomerInitializeOptions,
-    type CustomerRequestOptions,
-    type ExecutePaymentMethodCheckoutOptions,
-    type FormField,
-    type GuestCredentials,
-    type SignInEmail,
 } from '@bigcommerce/checkout-sdk';
 import { createBigCommercePaymentsFastlaneCustomerStrategy } from '@bigcommerce/checkout-sdk/integrations/bigcommerce-payments';
 import { createBoltCustomerStrategy } from '@bigcommerce/checkout-sdk/integrations/bolt';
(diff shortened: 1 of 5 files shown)


packages/core/src/app/customer/Customer.tsx
```diff
@@ -59,293 +46,123 @@ export interface CustomerProps {
...
 export interface CustomerState {
     isEmailLoginFormOpen: boolean;
     isReady: boolean;
     hasRequestedLoginEmail: boolean;
+    draftEmail?: string;
 }
 
-class Customer extends Component<CustomerProps & WithCheckoutCustomerProps & AnalyticsContextProps, CustomerState> {
-    state: CustomerState = {
+const Customer: React.FC<CustomerProps & AnalyticsContextProps> = ({
```
bc-peng, line 57:
```suggestion
const Customer: React.FC<CustomerProps> = ({
```
Can we get AnalyticsContext from the hook?

packages/core/src/app/customer/Customer.tsx
```diff
@@ -59,293 +46,123 @@ export interface CustomerProps {
...
+                try {
+                    await customerData.actions.deinitializeCustomer({ 
+                        methodId: customerData.data.providerWithCustomCheckout 
+                    });
+                } catch (error) {
+                    onUnhandledError(error);
+                }
+            };
+
+            cleanup();
```
bc-peng:
```suggestion
            void cleanup();
```

packages/core/src/app/customer/Customer.tsx
```diff
@@ -59,293 +46,123 @@ export interface CustomerProps {
...
+        };
+    }, [customerData.actions.deinitializeCustomer, customerData.data.providerWithCustomCheckout, onUnhandledError]);
```
bc-peng, line 137:
```suggestion
    }, []);
```
Just need to run once.

packages/core/src/app/customer/Customer.tsx
```diff
@@ -393,195 +208,167 @@ class Customer extends Component<CustomerProps & WithCheckoutCustomerProps & Ana
...
+                handleShowLogin={handleShowLogin}
+                isFloatingLabelEnabled={customerData.data.isFloatingLabelEnabled}
+                isSubscribed={isSubscribed}
+                isWalletButtonsOnTop={isWalletButtonsOnTop}
+                onUnhandledError={onUnhandledError}
+                onWalletButtonClick={onWalletButtonClick}
+                step={step}
+            />
+        );
+    }, [state.draftEmail, customerData.data.email, customerData.data.isFloatingLabelEnabled, handleChangeEmail, handleContinueAsGuest, handleShowLogin, isSubscribed, isWalletButtonsOnTop, onUnhandledError, onWalletButtonClick, step]);
```
bc-peng, line 302: Can we place the component directly into the `return`?

packages/core/src/app/customer/Customer.tsx
```diff
@@ -393,195 +208,167 @@ class Customer extends Component<CustomerProps & WithCheckoutCustomerProps & Ana
...
-    } = config;
-
-    const providerWithCustomCheckout = getProviderWithCustomCheckout(
-        config.checkoutSettings.providerWithCustomCheckout,
+    return (
+        <>
+            {state.isEmailLoginFormOpen && renderEmailLoginLinkForm()}
+            {shouldRenderLoginForm && renderLoginForm()}
+            {shouldRenderGuestForm && renderGuestForm()}
+            {shouldRenderCreateAccountForm && renderCreateAccountForm()}
```
bc-peng, line 369: Thought we should avoid function calls in the returned component.

packages/core/src/app/customer/Customer.tsx
animesh1987: I will do this as a follow up.

packages/core/src/app/customer/Customer.tsx
animesh1987: Yep you are correct, I am going to do all this as a follow up to this PR.

packages/core/src/app/customer/Customer.tsx
animesh1987: Do it in a follow up PR.

bc-peng (approved):

PR #2629, opened 2025-10-09

bc-peng:
## What/Why?

The bare minimum to convert `Payment` from class component to function component.
- Use `useEffect`.
- Convert private methods into arrow functions.

## Rollout/Rollback

Revert.

## Testing
- CI.
- Manually tested failure paths previously.

### DIFF
--- a/packages/core/src/app/payment/Payment.tsx
+++ b/packages/core/src/app/payment/Payment.tsx
@@ -8,7 +8,14 @@ import {
 } from '@bigcommerce/checkout-sdk';
 import { memoizeOne } from '@bigcommerce/memoize';
 import { compact, find, isEmpty, noop } from 'lodash';
-import React, { Component, type ReactNode } from 'react';
+import React, {
+  type ReactElement,
+  type ReactNode,
+  useCallback,
+  useEffect,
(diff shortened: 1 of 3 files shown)


packages/core/src/app/payment/Payment.tsx
```diff
@@ -454,101 +331,204 @@ class Payment extends Component<
...
+                const order = state.data.getOrder();
+
+                onFinalize(order?.orderId);
+            } catch (error) {
+                if (isErrorWithType(error) && error.type !== 'order_finalization_not_required') {
+                    onFinalizeError(error);
+                }
+            }
+
+            grandTotalChangeUnsubscribe.current = checkoutServiceSubscribe(
```
animesh1987, line 458: Do you know what this does?

bc-peng: `checkoutServiceSubscribe` is `checkoutService.subscribe` under the hood.

> Subscribes to any changes to the current state.
> The method registers a callback function and executes it every time there
is a change in the checkout state.

Therefore, here is for handling a cart total change during the payment flow.

PR #2630, opened 2025-10-10
feat(others): CHECKOUT-9427 Update styling for checklist item

animesh1987:
## What/Why?
Update styling for checklist item

## Rollout/Rollback
Hidden behind theme v2 changes

## Testing

https://github.com/user-attachments/assets/ecc1d5a3-3afc-4efd-aeb9-f2fc66f632f5

### DIFF
--- a/packages/core/src/scss/themeV2/components/checkout/_checklist.scss
+++ b/packages/core/src/scss/themeV2/components/checkout/_checklist.scss
@@ -6,8 +6,8 @@
 
         .form-checklist-item {
             border: 1px solid $color-greyMedium;
-            border-radius: 3px;
-            margin: 3px 0;
+            border-radius: spacing("sixth");
+            margin: spacing("sixth") 0;
 
             &:first-child {
                 margin-top: 0;


packages/core/src/scss/themeV2/components/checkout/_checklist.scss
```diff
@@ -0,0 +1,70 @@
+.form-checklist {
+    @include themeV2 {
+        border: none;
+        display: flex;
+        flex-direction: column;
+
+        .form-checklist-item {
+            border: 1px solid $color-greyMedium;
+            border-radius: 3px;
+            margin: 3px 0;
```
bc-peng: The PR looks good. 
Just to confirm, we won't be using 3px anywhere else, right?

animesh1987: Actually good point, I need to update that with spacing.

packages/core/src/scss/themeV2/components/checkout/_checklist.scss
```diff
@@ -0,0 +1,70 @@
+.form-checklist {
+    @include themeV2 {
+        border: none;
+        display: flex;
+        flex-direction: column;
+
+        .form-checklist-item {
+            border: 1px solid $color-greyMedium;
```
bc-peng, line 8: I see three different border values here.
> border: 1px solid $color-greyMedium;
> border: $input-border-width $input-border-style $color-black;
> border: 2px $input-border-style $input-border-color;

How many border styles do we have for V2 theme?

I'm asking because if we manage them through few variables like `v2BorderStyleGrey`, `v2BorderStyleBlack`, it would be easier to change styles globally in the future.

animesh1987: I am still using these from old themes as my assumption for now is V2 is overrides.

bc-launchbay:
success [20251016234145](https://launchbay.bigcommerce.net/projects/535/releases/266029) by Launchbay

PR #2633, opened 2025-10-13

bc-peng:
## What/Why?

Fixed the issues discovered when releasing bigcommerce/checkout-js#2632.

Two issues:
- `handleSubmit` missed a dependency, `props.isPaymentDataRequired()`, which created a big difference when not applying store credit.
-  The state value of the `Payment` component was captured in `handleCartTotalChange`  closure. The captured state value prevented the `PaymentForm` from re-rendering itself in the previous PR.

## Rollout/Rollback

Revert.

## Testing
- CI.
- Manually tested failure paths previously.

### DIFF
--- a/packages/core/src/app/payment/Payment.tsx
+++ b/packages/core/src/app/payment/Payment.tsx
@@ -104,16 +104,9 @@ const Payment= (props: PaymentProps & WithCheckoutPaymentProps & WithLanguagePro
         validationSchemas: {},
         submitFunctions: {},
     });
-    const grandTotalChangeUnsubscribe = useRef<() => void>();
 
-    const getContextValue = memoizeOne(() => {
-        return {
-            disableSubmit,
-            setSubmit,
-            setValidationSchema,
-            hidePaymentSubmitButton,
-        };
-    });
+    const isReadyRef = useRef(state.isReady);
+    const grandTotalChangeUnsubscribe = useRef<() => void>();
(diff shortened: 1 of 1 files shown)


packages/core/src/app/payment/Payment.tsx
```diff
@@ -454,101 +324,217 @@ class Payment extends Component<
 
             onSubmitError(error);
         }
-    };
+    }, [props.defaultMethod, state.selectedMethod, props.isPaymentDataRequired()]);
```
bc-peng, line 327: ### Issue1
`handleSubmit` missed a dependency, props.isPaymentDataRequired(), which created a big difference when not applying store credit.

packages/core/src/app/payment/Payment.tsx
```diff
@@ -454,101 +324,217 @@ class Payment extends Component<
...
             }
         } catch (error) {
             onUnhandledError(error);
         }
     }
 
-    private async handleCartTotalChange(): Promise<void> {
-        const { isReady } = this.state;
+    const handleCartTotalChange = async (): Promise<void> => {
+        const isReady = isReadyRef.current;
```
bc-peng, line 410: ### Issue2
The state value of the Payment component was captured in `handleCartTotalChange` closure before.
The captured state value prevented the PaymentForm from re-rendering itself.

PR #2637, opened 2025-10-14
feat(checkout): CHECKOUT-9547 Add CaptureMessage to Sentry Logger

bc-peng:
## What/Why?

Create a solution to log the usage of `CreditCardPaymentMethod` in `core` and `CreditCardFieldset` in `instrument-utils/src` via Sentry so that we can decide whether to keep these components.

## Rollout/Rollback

Revert.

### Testing capture
https://bigcommerce.sentry.io/issues/6942942622/
image

### DIFF
--- a/packages/core/src/app/payment/paymentMethod/CreditCardPaymentMethod.tsx
+++ b/packages/core/src/app/payment/paymentMethod/CreditCardPaymentMethod.tsx
@@ -212,7 +212,7 @@ class CreditCardPaymentMethod extends Component<
             ? isInstrumentCardCodeRequiredProp(selectedInstrument, method)
             : false;
 
-        const SentryMessage = method ? `CoreCreditCardPaymentMethod method "${method.method}" (ID: ${method.id}, type: ${method.type}) ${method.gateway ? `via ${method.gateway}` : 'with no gateway'}, supports ${method.supportedCards?.length ? method.supportedCards.join(', ') : 'no specific cards'}, ${method.skipRedirectConfirmationAlert ? 'skips' : 'requires'} redirect confirmation${method.returnUrl ? `, return URL: ${method.returnUrl}` : ''}.` : '';
+        const SentryMessage = method ? `DataCoreCreditCardPaymentMethod  ${JSON.stringify(method)}` : '';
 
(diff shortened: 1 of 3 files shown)


packages/core/src/app/checkout/CheckoutApp.tsx
```diff
@@ -53,13 +53,13 @@ const CheckoutApp = (props: CheckoutAppProps): ReactElement => {
     }, []);
 
     return (
-        <ErrorBoundary logger={errorLogger}>
+        <ErrorBoundary errorLogger={errorLogger}>
```
bc-peng, line 56: Consolidate the logger name to `errorLogger`.

packages/core/src/app/common/error/SentryErrorLogger.ts
```diff
@@ -97,6 +97,14 @@ export default class SentryErrorLogger implements ErrorLogger {
         });
     }
 
+    logMessage(
+        message: string,
+    ): void {
+        this.loadSentry().then(() => {
+            Sentry.captureMessage(message);
```
bc-peng, line 104: Sentry no longer provides the `captureEvent` method, so using `captureMessage` here.
https://docs.sentry.io/platforms/javascript/usage/

packages/payment-integration-api/src/CaptureMessageComponent.tsx
```diff
@@ -0,0 +1,29 @@
+import { type ReactNode, useEffect } from 'react';
+
+import { useCheckout } from './';
+
+interface CaptureMessageComponentProps {
+    message: string;
+}
+
+export const CaptureMessageComponent = ({ message }: CaptureMessageComponentProps): ReactNode => {
```
bc-peng, line 9: Thought creating a component would make the usage easier as the component will handle the context access.

animesh1987: We can keep it around for future use cases.

bc-launchbay:
success [20251015015401](https://launchbay.bigcommerce.net/projects/535/releases/265835) by Launchbay

PR #2668, opened 2025-10-27
feat(payment): customer step buttons height 40px

bc-dronov:
## What/Why?
Increased payment button height on customer step to 40px
Also includes PR:
https://github.com/bigcommerce/checkout-sdk-js/pull/3046

## Rollout/Rollback
Revert this PR

## Testing
Screenshot 2025-10-28 at 15 18 41

### DIFF
--- a/packages/core/src/app/customer/CheckoutButton.tsx
+++ b/packages/core/src/app/customer/CheckoutButton.tsx
@@ -2,7 +2,7 @@ import { type CustomerInitializeOptions, type CustomerRequestOptions } from '@bi
 import { noop } from 'lodash';
 import React, { type ReactElement, useEffect } from 'react';
 
-const WALLET_BUTTON_HEIGHT = 36;
+const WALLET_BUTTON_HEIGHT = 40;
 
 export interface CheckoutButtonProps {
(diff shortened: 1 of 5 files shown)


packages/core/src/app/customer/CheckoutButton.tsx
```diff
@@ -2,7 +2,7 @@ import { type CustomerInitializeOptions, type CustomerRequestOptions } from '@bi
 import { noop } from 'lodash';
 import React, { type ReactElement, useEffect } from 'react';
 
-const WALLET_BUTTON_HEIGHT = 36;
+const WALLET_BUTTON_HEIGHT = 40;
```
animesh1987, line 5: Is there a reason why we are doing this?

cc @JessicaHu51

bc-dronov: Hi @animesh1987 ! We need to implement a Stripe Link V2 button on the customer step. Its minimum height is 40px, and there's no other option to set it to 36px, so we decided to increase the height of all other buttons to 40px.

animesh1987 (approved):

PR #2683, opened 2025-11-04
chore(checkout): CHECKOUT-9619 Move cart summary section to its own component

animesh1987:
## What/Why?
Move cart summary section to its own component

## Rollout/Rollback
- revert this PR

## Testing
- CI
- screencast

https://github.com/user-attachments/assets/86b7f24e-6dc7-4647-a628-84d83cf46cb4

### DIFF
--- a/packages/core/src/app/checkout/components/CartSummary.tsx
+++ b/packages/core/src/app/checkout/components/CartSummary.tsx
@@ -46,7 +46,7 @@ export const CartSummary: React.FC<CartSummaryProps> = ({ isMultiShippingMode })
 
                 return (
                     <LazyContainer loadingSkeleton={<CartSummarySkeleton />}>
-                        <aside className="layout-cart">
+                        <aside aria-label="Cart Summary" className="layout-cart">
                             <CartSummaryComponent isMultiShippingMode={isMultiShippingMode} />
                             <Extension region={ExtensionRegion.SummaryAfter} />
                         </aside>


packages/core/src/app/checkout/components/CartSummary.tsx
```diff
@@ -0,0 +1,60 @@
...
+                        <LazyContainer loadingSkeleton={<></>}>
+                            <Extension region={ExtensionRegion.SummaryAfter} />
+                            <CartSummaryDrawer isMultiShippingMode={isMultiShippingMode} />
+                        </LazyContainer>
+                    );
+                }
+
+                return (
+                    <LazyContainer loadingSkeleton={<CartSummarySkeleton />}>
+                        <aside className="layout-cart">
```
TomA-R: While you're here, should we add an aria label?
```suggestion
                        <aside className="layout-cart" role="complementary" aria-label="Cart summary">
```

animesh1987: Good suggestion. 🧠

animesh1987: Can't do `role="complementary"` as got a lint error.
> error  The element aside has an implicit role of complementary. Defining this explicitly is redundant and should be avoided  jsx-a11y/no-redundant-roles

bc-launchbay:
success [20251104233859](https://launchbay.bigcommerce.net/projects/535/releases/267527) by Launchbay

PR #2707, opened 2025-11-13
CHECKOUT-9450: Pass strategies as part of order finalisation

davidchin:
## What/Why?

Pass strategies that require client-side orchestration into the `finalizeOrder` method; otherwise, they won’t be executed once the fallback is removed.

## Rollout/Rollback
Revert or turn off `CHECKOUT-9450.lazy_load_payment_strategies`

## Testing

See https://github.com/bigcommerce/checkout-sdk-js/pull/3080

### DIFF
--- a/packages/core/src/app/payment/Payment.tsx
+++ b/packages/core/src/app/payment/Payment.tsx
@@ -3,9 +3,16 @@ import {
     type CheckoutSelectors,
     type CheckoutService,
     type CheckoutSettings,
+    type OrderFinalizeOptions,
     type OrderRequestBody,
     type PaymentMethod,
 } from '@bigcommerce/checkout-sdk';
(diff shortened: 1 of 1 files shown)


packages/core/src/app/payment/Payment.tsx
```diff
@@ -448,7 +455,20 @@ const Payment= (props: PaymentProps & WithCheckoutPaymentProps & WithLanguagePro
             await loadPaymentMethodsOrThrow();
 
             try {
-                const state = await finalizeOrderIfNeeded();
+                const state = await finalizeOrderIfNeeded({
+                    integrations: [
```
animesh1987, line 459: These seem quite limited, do we need to pass more?

davidchin: I scanned through the codebase, these are the strategies that that implement the finalize method. If there are more, we will see in the Sentry log.

bc-peng (approved):

animesh1987 (approved):

PR #2716, opened 2025-12-01
fix(checkout): CHECKOUT-9630 Display error if no shipping countries are configured

animesh1987:
## What/Why?
Display an error if no shipping countries are configured, so shipping step is not loaded.

## Rollout/Rollback
- revert this PR

## Testing
- CI
- Screencast

https://github.com/user-attachments/assets/748cbee7-03a5-46ae-9c0c-b2108818263c

### DIFF
--- a/packages/core/src/app/shipping/ShippingForm.tsx
+++ b/packages/core/src/app/shipping/ShippingForm.tsx
@@ -58,7 +58,7 @@ const ShippingForm = ({
     const { extensionState: { shippingFormRenderTimestamp } } = useExtensions();
 
     const config = getConfig();
-    const isNoCountriesErrorOnCheckoutEnabled = isExperimentEnabled(config?.checkoutSettings, 'CHECKOUT-9630.no_countries_error_on_checkout');
+    const isNoCountriesErrorOnCheckoutEnabled = isExperimentEnabled(config?.checkoutSettings, 'CHECKOUT-9630.no_countries_error_on_checkout', false);
 
     useEffect(() => {
         if (shippingFormRenderTimestamp) {


packages/core/src/app/shipping/ShippingForm.tsx
```diff
@@ -68,6 +73,17 @@ const ShippingForm = ({
...
+    // Handle error when no countries are available - use useEffect to avoid setState during render
+    useEffect(() => {
+        if (countries.length === 0 && isNoCountriesErrorOnCheckoutEnabled) {
+            onUnhandledError(new CustomError({
+                name: 'no_countries_available',
+                message: getLanguageService().translate('shipping.no_countries_available_message'),
+                title: getLanguageService().translate('shipping.no_countries_available_heading'),
+            }));
+        }
+    }, [countries.length, isNoCountriesErrorOnCheckoutEnabled, onUnhandledError]);
```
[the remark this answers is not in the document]
bc-peng: Think we can remove `onUnhandledError` and `isNoCountriesErrorOnCheckoutEnabled` from dependencies as by the time we render this form, those values are fixed.

packages/core/src/app/shipping/Shipping.test.tsx
```diff
@@ -828,4 +828,76 @@ describe('Shipping step', () => {
...
+            jest.spyOn(checkoutService.getState().data, 'getShippingCountries').mockReturnValue([]);
+
+            // Spy on errorLogger to verify error handling
+            jest.spyOn(defaultProps.errorLogger, 'log');
+
+            render(<CheckoutTest {...defaultProps} />);
+
+            await checkout.waitForShippingStep();
+
+            // Verify that error was NOT logged when experiment is disabled
```
bc-peng, line 894: I am seeing 8 comments in this PR, could you please confirm are they all necessary?

packages/core/src/app/shipping/ShippingForm.tsx
```diff
@@ -79,6 +95,10 @@ const ShippingForm = ({
         />;
     };
 
+    if (countries.length === 0 && isNoCountriesErrorOnCheckoutEnabled) {
```
richapatel1510, line 98: We can may be move this condition to a const using useMemo and then use at both the places.

packages/core/src/app/shipping/Shipping.test.tsx
animesh1987: I will remove them when I clean up the experiment code.

packages/core/src/app/shipping/ShippingForm.tsx
animesh1987: I will keep this as is, we can't throw the error from here since `onUnhandledError` sets state in CheckoutPage component. If we throw the error from here, then state is set multiple times during render cycle which causes the app to break.

bc-peng (dismissed):

richapatel1510 (dismissed):

bc-peng (approved):

PR #2718, opened 2025-12-02
refactor(checkout): CHECKOUT-9620 Refactor CheckoutPage

bc-peng:
## What/Why?
JavaScript closures were causing stale state values in the checkout flow, preventing updates to consignments from being handled correctly.

This issue surfaced in an earlier E2E test.

Related areas potentially affected by closure behavior have also been refactored.

## Rollout/Rollback

Revert.

### Before (shipping step disappears)

https://github.com/user-attachments/assets/87483bab-4b06-44f1-b6cb-45c460287d05

### After (showing error modal then going to shipping step)

https://github.com/user-attachments/assets/e4187589-4f1a-433d-b306-4112066977d1

### DIFF
--- a/packages/core/src/app/checkout/CheckoutPage.tsx
+++ b/packages/core/src/app/checkout/CheckoutPage.tsx
@@ -135,16 +135,6 @@ const Checkout = ({
                       subscribeToConsignments,
                       themeV2
                   }: CheckoutPageProps):ReactElement => {
-    const stepsRef = useRef<CheckoutStepStatus[]>(steps);
-    const embeddedMessenger = useRef<EmbeddedCheckoutMessenger>();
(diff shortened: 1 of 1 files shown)


packages/core/src/app/checkout/CheckoutPage.tsx
```diff
@@ -369,6 +375,10 @@ const Checkout = ({
         window.location.reload();
     }, []);
 
+    const handleSetIsMultishippingMode = useCallback((value: boolean): void => {
+        setState(prevState => ({ ...prevState, isMultiShippingMode: value }));
+    }, []);
+
```
bc-peng, line 382: This is not related to the e2e test.
Just a refactoring to prevent unnecessary shipping step re-render.

PR #2721, opened 2025-12-08
feat(checkout): CHECKOUT-9677 Enable Adding or Removing Coupons

bc-peng:
## What/Why?

Enable adding or removing multi coupons and gift certificates.

## Rollout/Rollback

Revert.

### Applying/removing coupons.

https://github.com/user-attachments/assets/7f203c1c-f675-4d1a-a70f-9886761cfb6c

### Applying/removing gift certificates.

https://github.com/user-attachments/assets/3334c72f-40f8-409c-b3a7-2b03ee195b96

### DIFF
--- a/packages/core/src/app/coupon/components/AppliedCouponOrGiftCertificate.tsx
+++ b/packages/core/src/app/coupon/components/AppliedCouponOrGiftCertificate.tsx
@@ -2,12 +2,12 @@ import React, { type FunctionComponent } from 'react';
 
 import { IconCoupon, IconRemoveCoupon } from '@bigcommerce/checkout/ui';
 
-interface AppliedCouponProps {
+interface AppliedCouponOrGiftCertificateProps {
     code: string;
-    onClick?: () => void;
+    onClick: () => void;
 }
 
-export const AppliedCouponOrGiftCertificate: FunctionComponent<AppliedCouponProps> = ({
+export const AppliedCouponOrGiftCertificate: FunctionComponent<AppliedCouponOrGiftCertificateProps> = ({
     code,
     onClick,
 }) => (


packages/core/src/app/coupon/MultiCoupon.tsx
bc-peng, line 1: Renaming the file to `NewOrderSummarySubtotals.tsx` as the file contains complete order summary details.

packages/core/src/app/coupon/components/AppliedCouponOrGiftCertificate.tsx
```diff
@@ -0,0 +1,20 @@
+import React, { type FunctionComponent } from 'react';
+
+import { IconCoupon, IconRemoveCoupon } from '@bigcommerce/checkout/ui';
+
+interface AppliedCouponProps {
+    code: string;
+    onClick?: () => void;
+}
+
+export const AppliedCouponOrGiftCertificate: FunctionComponent<AppliedCouponProps> = ({
```
animesh1987: The interface name doesn't match the component name here, is that intentional?

packages/core/src/app/coupon/components/CouponForm.tsx
```diff
@@ -3,30 +3,55 @@ import React, { type FunctionComponent, useState } from 'react';
...
+
+    const submitForm = async () => {
+        try {
+            await applyCouponOrGiftCertificate(code);
+
+            setCode('');
+        } catch (error) {
+            // TODO: Handle different error types accordingly
+            // eslint-disable-next-line no-console
+            console.log(error);
```
animesh1987, line 33: Can we use the logger that is part of checkout, since that links to sentry as well I believe.

packages/core/src/app/coupon/components/AppliedCouponOrGiftCertificate.tsx
```diff
@@ -0,0 +1,20 @@
+import React, { type FunctionComponent } from 'react';
+
+import { IconCoupon, IconRemoveCoupon } from '@bigcommerce/checkout/ui';
+
+interface AppliedCouponProps {
+    code: string;
+    onClick?: () => void;
```
animesh1987: Is this expected to be optional?

packages/core/src/app/coupon/components/AppliedCouponOrGiftCertificate.tsx
bc-peng: The name is updated now, it should be changed when I was renaming the component.😅

packages/core/src/app/coupon/components/CouponForm.tsx
bc-peng: This is temporary, the catch block will be replaced with a proper error handling in the next PR.

packages/core/src/app/coupon/components/AppliedCouponOrGiftCertificate.tsx
bc-peng: Updated, it should be required.

PR #2724, opened 2025-12-08
BC Checkout fixes

dostranderIAPP:
We have a custom BC checkout experience that is sub optimal. The best practice for custom checkouts is to use the BC "Optimized One Page Checkout" which we are implementing here. 

This repo requires a Linux environment to compile the npm build so we are also introducing a Docker container to manage that part. Please check the README.md for more information.


packages/core/src/app/customer/LoginForm.tsx
```diff
@@ -41,24 +37,16 @@ export interface LoginFormProps {
 }
 
 export interface LoginFormValues {
```
tdimambro1, line 39: can this interface be removed now that it's empty?

PR #2726, opened 2025-12-09
feat(checkout): CHECKOUT-9679 Handle Coupon Error

bc-peng:
## What/Why?

Handle errors when applying a coupon or gift certificate.

## Rollout/Rollback

Revert.

### When a coupon is invalid

https://github.com/user-attachments/assets/ada7e08d-2cf6-48c4-8a2c-5fa083bd994f

### When maximum coupon usage reached

https://github.com/user-attachments/assets/c8b7f86b-bd94-44d7-a9d7-76f87b2909f0

### DIFF
--- a/packages/core/src/app/coupon/components/CouponForm.tsx
+++ b/packages/core/src/app/coupon/components/CouponForm.tsx
@@ -33,6 +33,10 @@ export const CouponForm: FunctionComponent = () => {
     };
 
     const submitForm = async () => {
+        if (!code) {
+            return;
+        }
+
         try {
             await applyCouponOrGiftCertificate(code);
 
--- a/packages/core/src/app/order/OrderSummary.tsx
+++ b/packages/core/src/app/order/OrderSummary.tsx
@@ -46,7 +46,7 @@ const OrderSummary: FunctionComponent<OrderSummaryProps & OrderSummarySubtotalsP
     const { checkoutSettings } = checkoutState.data.getConfig() ?? {};
 
     // TODO: When removing the experiment, rename `NewOrderSummarySubtotals` to `OrderSummarySubtotals`.
-    const isMultiCouponEnabled = isExperimentEnabled(checkoutSettings, 'PROJECT-7321-5991.multi-coupon-cart-checkout', false);
+    const isMultiCouponEnabled = isExperimentEnabled(checkoutSettings, 'CHECKOUT-9674.multi_coupon_cart_checkout', false);
 
     const nonBundledLineItems = useMemo(() => removeBundledItems(lineItems), [lineItems]);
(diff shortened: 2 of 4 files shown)


packages/core/src/app/coupon/useMultiCoupon.ts
```diff
@@ -6,14 +8,26 @@ interface UseMultiCouponValues {
...
+    isCouponCodeCollapsed: boolean;
     removeCoupon: (code: string) => Promise<void>;
     removeGiftCertificate: (giftCertificateCode: string) => Promise<void>;
-    isCouponCodeCollapsed: boolean;
+    setCouponError: (error: string | null) => void;
+    shouldDisableCouponForm: boolean;
 }
 
 export const useMultiCoupon = (): UseMultiCouponValues => {
+    const errorMessageTimeout = 7000;
```
bc-peng: Open to adjustments during product testing.

packages/core/src/app/coupon/components/CouponForm.tsx
```diff
@@ -28,9 +32,9 @@ export const CouponForm: FunctionComponent = () => {
 
             setCode('');
         } catch (error) {
-            // TODO: Handle different error types accordingly
-            // eslint-disable-next-line no-console
-            console.log(error);
+            if (error instanceof Error) {
+                setCouponError(error.message);
+            }
```
bc-peng, line 47: We just relay the error message from the API here.
If it is not a standard error, then it will escalate to the `ErrorBoundries` and go to Sentry eventually.

packages/core/src/app/coupon/useMultiCoupon.ts
```diff
@@ -53,12 +67,27 @@ export const useMultiCoupon = (): UseMultiCouponValues => {
         await checkoutService.removeGiftCertificate(code);
     };
 
+    useEffect(() => {
```
animesh1987: Do we need to attach this in the custom hook? Can we not keep this in the component itself, and only set error when we have the error passed in?

packages/core/src/app/coupon/useMultiCoupon.ts
animesh1987: Why do we need to set this timeout?

bc-peng: I have removed timed auto-dismiss, pending design confirmation.

packages/core/src/app/coupon/useMultiCoupon.ts
bc-peng: Removed as part of auto-dismiss removal.

richapatel1510 (approved):

PR #2906, opened 2026-03-26
feat(checkout): CHECKOUT-9900 [Accessibility] Resolve broken ARIA reference for address fields on the checkout

bc-charlesho:
### What?
Ensure `aria-labelledby` reference always resolves to a DOM element in the checkout's address step.

### Why?
[WAVE](https://wave.webaim.org/) accessibility evaluation tool (via Chrome extension) reported broken ARIA reference on the address fields. This PR resolves the issue in order to adhere to WCAG compliance standards.

## Rollout/Rollback

Revert the PR

## Testing

- [x] Unit tests pass (5 new/updated tests)
- [x] Run WAVE browser extension and confirm the 'Broken ARIA reference' error is gone
- [x] Verify error messages still display correctly after form submission with invalid input

### Before:
Before

Screenshot 2026-03-26 at 10 45 22 am

### After:
After

Screenshot 2026-03-26 at 12 19 03 pm

Screenshot 2026-03-26 at 12 19 45 pm

---

> [!NOTE]
> **Low Risk**
> Low risk UI/accessibility change that adjusts error rendering markup and test coverage; main risk is minor regression in when/where validation messages appear.
> 
> **Overview**
> Fixes broken `aria-labelledby` references by ensuring every field’s `*-field-error-message` id is always present in the DOM via `FormFieldError` rendering a screen-reader-only `<span>` when no error is shown.
> 
> Updates `FormFieldError` to only render the visible alert after submit *and* when the field has both `touched` and an error, and revises unit tests plus Braintree ACH snapshots to match the new error-container markup.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit fa72ab6d8a67619d7995f2e6f4012f4419b71f84. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/ui/src/form/FormFieldError/FormFieldError.tsx
+++ b/packages/ui/src/form/FormFieldError/FormFieldError.tsx
@@ -1,4 +1,4 @@
-import { ErrorMessage } from 'formik';
+import { ErrorMessage, getIn, useFormikContext } from 'formik';
 import React, { type FunctionComponent, memo, useCallback } from 'react';
 
 import { FormContext } from '../contexts';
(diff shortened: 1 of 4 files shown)


packages/ui/src/form/FormFieldError/FormFieldError.tsx
```diff
@@ -10,30 +10,40 @@ export interface FormFieldErrorProps {
...
-        [errorId, name, testId],
+        [errorId, name],
     );
 
     return (
         <FormContext.Consumer>
-            {({ isSubmitted }) =>
-                isSubmitted && <ErrorMessage name={name} render={renderMessage} />
-            }
+            {({ isSubmitted }) => (
```
animesh1987, line 36: This seems like a bigger change as compared to just adding aria labels?

bc-charlesho: The concern is valid.
The aria-labelledby attributes on form inputs already reference {id}-field-error-message, but the element with that ID was only rendered when isSubmitted && hasError. When there's no error, the referenced element didn't exist in the DOM at all, causing the WAVE "Broken ARIA reference" violation.

Simply adding aria labels wouldn't fix it — the problem is that the target element of the existing aria-labelledby is conditionally rendered. The PR ensures the list  wrapper is always present, and when there's no error, a hidden placeholder  renders so the reference always resolves. When there is an error, it renders the same  as before.

This matches the pattern already used in packages/core/src/app/ui/form/FormFieldError.tsx (lines 34-49), which has the same approach in place.

bc-charlesho:
Will re-implement based on the latest code base.

PR #2965, opened 2026-04-23
feat(payment): PI-5142 [Oney] Add Oney payment method

bc-ania:
## What/Why?

**Add select and render grouped Oney payment method variant options**

When groupedMethods is detected in initializationData, the component takes a lightweight path: renders only the select variant selector, a hidden SDK mount point, and the additional-action Modal (for BNPL redirect/3DS). The Adyen strategy is initialized and deinitialized directly via useEffect — no HostedWidgetPaymentComponent, no hosted card fields. On variant change the strategy is reinitialized and methodIdOverride is updated. MethodIdOverride is cleared on unmount. The non-grouped path (credit card, BCMC, etc.) is completely unchanged.

Introduced a generic prefix-based grouping mechanism via a standalone groupMethodsByPrefix function, applied in getDefaultPaymentMethod through GROUPED_METHOD_ID_PREFIXES.reduce(groupMethodsByPrefix, filteredMethods). Methods sharing a common ID prefix (facilypay_) are sorted numerically and collapsed into a single representative — the numerically first variant at its original position in the list. Non-representative variants are removed from filteredMethods entirely. The representative carries all variants in initializationData.groupedMethods and has its numeric prefix stripped from displayName (e.g. "3x Oney by bank card" → "Oney by bank card"). 
Adding support for a new group in the future requires only adding a prefix to GROUPED_METHOD_ID_PREFIXES.

## Rollout/Rollback

Revert PR or disable experiment PAYMENTS-5142.payment_method_grouping

## Testing 

https://github.com/user-attachments/assets/bf62fcef-4145-4bc7-9afc-83e7b0aad6bf

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes how payment methods are filtered/rendered and how Adyen V3 initializes/deinitializes payment for grouped variants, which can affect checkout payment selection and submission behavior when the experiment is enabled.
> 
> **Overview**
> Adds **experiment-gated payment method grouping** (`PAYMENTS-5142.payment_method_grouping`) that collapses multiple `facilypay_*` methods into a single representative method with `initializationData.groupedMethods`, including numeric-suffix sorting and display-name cleanup.
> 
> Updates `AdyenV3PaymentMethod` to detect grouped methods and render a lightweight variant selector + Adyen mount points, managing `initializePayment`/`deinitializePayment` directly and using `paymentForm`’s `methodIdOverride` to switch variants (cleared on unmount).
> 
> Expands unit/integration coverage to verify grouping behavior in filtering/UI and the grouped Adyen variant flow.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a4e1a91b7127364fc60612c61ab1b74e4883a1e5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/app/payment/Payment.tsx
+++ b/packages/core/src/app/payment/Payment.tsx
@@ -4,7 +4,6 @@ import {
     type CartStockPositionsChangedError,
     type CheckoutSelectors,
     type CheckoutService,
-    type CheckoutSettings,
     type Consignment,
     type OrderFinalizeOptions,
     type OrderRequestBody,
@@ -77,7 +76,6 @@ export interface PaymentProps {
 interface WithCheckoutPaymentProps {
     availableStoreCredit: number;
     cart?: Cart;
-    checkoutSettings: CheckoutSettings;
     consignments?: Consignment[];
     cartUrl: string;
     defaultMethod?: PaymentMethod;
@@ -681,10 +679,7 @@ export function mapToPaymentProps(
         return null;
     }
 
-    const checkoutSettings = config.checkoutSettings & {
-        orderTermsAndConditionsLocation: string,
-    };
-
+    const checkoutSettings = config.checkoutSettings;
     const {
         enableTermsAndConditions: isTermsConditionsEnabled,
         features,
@@ -708,7 +703,6 @@ export function mapToPaymentProps(
         applyStoreCredit: checkoutService.applyStoreCredit,
         availableStoreCredit: customer.storeCredit,
         cart: getCart(),
-        checkoutSettings,
         consignments,
         cartUrl: config.links.cartLink,
         clearError: checkoutService.clearError,
--- a/packages/core/src/app/payment/Payment.tsx
+++ b/packages/core/src/app/payment/Payment.tsx
@@ -681,7 +681,7 @@ export function mapToPaymentProps(
         return null;
     }
 
-    const checkoutSettings = config.checkoutSettings as CheckoutSettings & {
+    const checkoutSettings = config.checkoutSettings & {
         orderTermsAndConditionsLocation: string;
     };
 
--- a/packages/core/src/app/payment/Payment.tsx
+++ b/packages/core/src/app/payment/Payment.tsx
@@ -77,6 +77,7 @@ export interface PaymentProps {
 interface WithCheckoutPaymentProps {
     availableStoreCredit: number;
     cart?: Cart;
+    checkoutSettings: CheckoutSettings;
     consignments?: Consignment[];
     cartUrl: string;
     defaultMethod?: PaymentMethod;
@@ -480,7 +481,6 @@ const Payment= (props: PaymentProps & WithCheckoutPaymentProps & WithLanguagePro
             const checkout = updatedState.data.getCheckout();
             const config = updatedState.data.getConfig();
             const methods = updatedState.data.getPaymentMethods() || EMPTY_ARRAY;
-
             const defaultMethod = (checkout && config)
                 ? getFilteredPaymentMethodsWithDefault({
                       checkout,
@@ -681,13 +681,17 @@ export function mapToPaymentProps(
         return null;
     }
 
+    const checkoutSettings = config.checkoutSettings as CheckoutSettings & {
+        orderTermsAndConditionsLocation: string;
+    };
+
     const {
         enableTermsAndConditions: isTermsConditionsEnabled,
         features,
         orderTermsAndConditionsType: termsConditionsType,
         orderTermsAndConditions: termsCondtitionsText,
         orderTermsAndConditionsLink: termsCondtitionsUrl,
-    } = config.checkoutSettings as CheckoutSettings & { orderTermsAndConditionsLocation: string };
+    } = checkoutSettings;
 
     const isTermsConditionsRequired = isTermsConditionsEnabled;
     const { isStoreCreditApplied } = checkout;
@@ -704,6 +708,7 @@ export function mapToPaymentProps(
         applyStoreCredit: checkoutService.applyStoreCredit,
         availableStoreCredit: customer.storeCredit,
         cart: getCart(),
+        checkoutSettings,
         consignments,
         cartUrl: config.links.cartLink,
         clearError: checkoutService.clearError,
@@ -721,7 +726,7 @@ export function mapToPaymentProps(
         shouldExecuteSpamCheck: checkout.shouldExecuteSpamCheck,
         shouldLocaliseErrorMessages:
             features['PAYMENTS-6799.localise_checkout_payment_error_messages'],
-        shouldShowSubmitPaymentButton: isExperimentEnabled(config.checkoutSettings, 'CHECKOUT-9729.show_submit_button_when_payment_not_required', false),
+        shouldShowSubmitPaymentButton: isExperimentEnabled(checkoutSettings, 'CHECKOUT-9729.show_submit_button_when_payment_not_required', false),
         submitOrder: checkoutService.submitOrder,
         submitOrderError: getSubmitOrderError(),
         checkoutServiceSubscribe: checkoutService.subscribe,
--- a/packages/core/src/app/payment/Payment.tsx
+++ b/packages/core/src/app/payment/Payment.tsx
@@ -682,7 +682,7 @@ export function mapToPaymentProps(
     }
 
     const checkoutSettings = config.checkoutSettings & {
-        orderTermsAndConditionsLocation: string;
+        orderTermsAndConditionsLocation: string,
     };
 
     const {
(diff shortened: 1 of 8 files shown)


packages/core/src/app/payment/Payment.tsx
```diff
@@ -802,4 +825,4 @@ export function mapToPaymentProps({
     };
 }
 
-export default withAnalytics(withLanguage(withCheckout(mapToPaymentProps)(Payment)));
+export default withAnalytics(withLanguage(withCheckout(mapToPaymentProps)(Payment)));
```
bc-peng:
```suggestion
export default withAnalytics(withLanguage(withCheckout(mapToPaymentProps)(Payment)));

```

packages/core/src/app/payment/groupPaymentMethodsByPrefix.ts
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        if (m.id === representativeSourceId) {
+            return [representative];
+        }
+
+        return [];
+    });
+
+export const groupMethodsByPrefix = (methods: PaymentMethod[], prefix: string): PaymentMethod[] => {
```
bc-peng:
```suggestion
export const groupPaymentMethodsByPrefix = (methods: PaymentMethod[], prefix: string): PaymentMethod[] => {
```
For keeping it in line with the file name.

packages/core/src/app/payment/groupPaymentMethodsByPrefix.ts
```diff
@@ -0,0 +1,69 @@
...
+            displayName: first.config.displayName?.replace(/^\d+x\s+/i, '') ?? first.config.displayName,
+        },
+        initializationData: {
+            ...(first.initializationData as Record<string, unknown>),
+            groupedMethods: sortedGroup,
+        },
+    };
+};
+
+const spliceGroupedPaymentMethodRepresentativeIntoList = (
```
bc-peng:
```suggestion
const flatMapGroupedPaymentMethodRepresentativeIntoList = (
```

packages/core/src/app/payment/groupPaymentMethodsByPrefix.ts
```diff
@@ -0,0 +1,69 @@
...
+    const [first] = sortedGroup;
+
+    return {
+        ...first,
+        config: {
+            ...first.config,
+            displayName: first.config.displayName?.replace(/^\d+x\s+/i, '') ?? first.config.displayName,
+        },
+        initializationData: {
+            ...(first.initializationData as Record<string, unknown>),
```
bc-peng: Is it possible we avoid casting here?

packages/core/src/app/payment/groupPaymentMethodsByPrefix.ts
```diff
@@ -0,0 +1,74 @@
...
+
+function isRecord(value: unknown): value is Record<string, unknown> {
+    return typeof value === 'object' && value !== null && !Array.isArray(value);
+}
+
+const selectAndSortPaymentMethodsByPrefix = (
+    methods: PaymentMethod[],
+    prefix: string,
+): PaymentMethod[] | undefined => {
+    const group = methods.filter((m) => m.id.startsWith(prefix));
```
animesh1987, line 10: Since we are actively filtering some methods and removing them? Could this cause an issue?

packages/core/src/app/payment/groupPaymentMethodsByPrefix.ts
```diff
@@ -0,0 +1,74 @@
...
+    return [...group].sort((a, b) => {
+        const toNum = (id: string) => parseInt(id.slice(prefix.length), 10) || 0;
+
+        return toNum(a.id) - toNum(b.id);
+    });
+};
+
+const buildGroupedPaymentMethodRepresentative = (sortedGroup: PaymentMethod[]): PaymentMethod => {
+    const [first] = sortedGroup;
+    const initializationData = isRecord(first.initializationData) ? first.initializationData : {};
```
animesh1987: Are we just checking here that if `initializationData` is an object or not?

packages/core/src/app/payment/groupPaymentMethodsByPrefix.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    return {
+        ...first,
+        config: {
+            ...first.config,
+            displayName: first.config.displayName?.replace(/^\d+x\s+/i, '') ?? first.config.displayName,
+        },
+        initializationData: {
+            ...initializationData,
+            groupedMethods: sortedGroup,
```
animesh1987, line 34: Here does `groupedMethods` also have the first method?

packages/core/src/app/payment/groupPaymentMethodsByPrefix.ts
```diff
@@ -0,0 +1,74 @@
...
+    if (!sortedGroup) {
+        return methods;
+    }
+
+    const representative = buildGroupedPaymentMethodRepresentative(sortedGroup);
+
+    return flatMapGroupedPaymentMethodRepresentativeIntoList(
+        methods,
+        prefix,
+        sortedGroup[0].id,
```
animesh1987: Can this be `representative.id` as well?

packages/core/src/app/payment/groupPaymentMethodsByPrefix.ts
```diff
@@ -0,0 +1,74 @@
...
+    };
+};
+
+const flatMapGroupedPaymentMethodRepresentativeIntoList = (
+    methods: PaymentMethod[],
+    prefix: string,
+    representativeSourceId: string,
+    representative: PaymentMethod,
+): PaymentMethod[] =>
+    methods.flatMap((m) => {
```
animesh1987, line 45: Do we need a flatMap here or a simple map will do?

animesh1987: I got my answer, had to lookup flatmap in more detail. It will simply remove the empty array.

packages/core/src/app/payment/groupPaymentMethodsByPrefix.ts
bc-ania: It shouldn't cause an issue, the goal is to have one representative method, and put the whole group in representative.initializationData.groupedMethods

packages/core/src/app/payment/groupPaymentMethodsByPrefix.ts
bc-ania: yes, and to avoid cast `as Record<string, unknown>`

packages/core/src/app/payment/groupPaymentMethodsByPrefix.ts
bc-ania: yes, it has all the methods, the first one as well

bc-ania: Screenshot 2026-04-28 at 15 36 38

packages/core/src/app/payment/Payment.tsx
```diff
@@ -746,18 +763,23 @@ export function mapToPaymentProps({
         return null;
     }
 
+    const checkoutSettings = config.checkoutSettings as CheckoutSettings & {
+        orderTermsAndConditionsLocation: string;
+    };
```
bc-wpietrzak: Is this type assertion needed? I tried running lint without it and there were no errors

packages/core/src/app/payment/groupPaymentMethodsByPrefix.ts
bc-wpietrzak: Is `isRecord` necessary? I tried running test/lint without `isRecord` or type assertion, and there are no issues found.

packages/core/src/app/payment/groupPaymentMethodsByPrefix.ts
animesh1987: yes, that's the bit that confused me. Since grouped methods which is the sortedList goes under initialization data of first method. Is that correct? Feels like a bit of recursive loop.

bc-ania: GroupedMethods points to the same methods that were already in checkout before grouping. We need to add them so that select can see every variant. The first sorted method is a copy of the first variant in the list, and the full list next to it is attached so nothing is lost. The representative object is a shallow copy of first, so groupedMethods doesn't contain the representative because it holds the original first reference.

bc-peng (approved):

bc-wpietrzak (approved):

bc-launchbay:
success [20260505091950](https://launchbay.bigcommerce.net/projects/535/releases/283740) by Launchbay

bc-launchbay:
success [20260505091950](https://launchbay.bigcommerce.net/projects/535/releases/283740) by Launchbay

bc-launchbay:
success [20260505091950](https://launchbay.bigcommerce.net/projects/535/releases/283740) by Launchbay

PR #3025, opened 2026-05-21
feat(ui): CHECKOUT-9019 Add phone number validation to Checkout forms

bc-maxy:
## What/Why?

**What this does**:

Adds phone number validation to our address forms using a new library called `intl-tel-input`, which handles international phone formats.

**How it works:**

- The phone field now validates that numbers are entered correctly
- If a phone number field is optional and left blank, it skips validation. However, if something is typed in, it still checks that it's a valid number. Let me know if you think this should work in a different way.
- The country code in the phone field automatically matches whatever country is selected in the shipping form. Also let me know if this is not what you expected - I can adjust.

**Behind-the-scenes technical changes:**

Adding the `intl-tel-input` required some extra setup for Jest and Webpack to handle it properly. This is a fully modular package so we need to transform it for Jest. I added exclusion to transformer logic, as node_modules are excluded by default. I also added css and webp handlers, as that's what the library uses for styling and flag images internally.

All new functionality is hidden by `CHECKOUT-9019.use_new_phone_number_validation` experiment.

## Rollout/Rollback

Set `CHECKOUT-9019.use_new_phone_number_validation` to `false`.

## Testing

New CI tests + Manual testing (see below).

**When phone number field is optional:**

https://github.com/user-attachments/assets/94699b3e-88b3-450b-82dc-c1a6b9e3f35d

**When phone number is required:**

https://github.com/user-attachments/assets/c99cb628-6fb5-481f-9307-9577e6209db9

**Multi-shipping:**

https://github.com/user-attachments/assets/72bffafc-1813-438e-92da-6d8c82b4667d

**When the flag is set to false:**

https://github.com/user-attachments/assets/f55e17c0-5e3f-4c0d-9682-5455b3a02005

---

> [!NOTE]
> **Medium Risk**
> Touches checkout address phone capture and validation (customer-facing data); rollout is experiment-gated but wrong validation or country sync could block orders when enabled.
> 
> **Overview**
> Adds **international phone validation** on checkout address forms behind experiment `CHECKOUT-9019.use_new_phone_number_validation`, using **`intl-tel-input`** (React wrapper + utils chunk).
> 
> When the flag is on, telephone fields render **`PhoneFormField`** / **`IntlTelInput`** instead of a plain `tel` input. **Formik** validates non-empty values via `isValidNumber()` and shows **`address.phone_number_invalid_error`**. Empty optional phones skip format checks. **`AddressForm`** passes **`selectedCountry`** so the widget’s default country follows the address country when the field is empty.
> 
> **`DynamicFormField`** delegates rendering to new **`DynamicFormFieldSelector`**; **`FormField`** gains optional **`validate`**. **Focus-on-error** ignores hidden/search inputs and picks the first visible errored control (e.g. phone vs country search).
> 
> **Jest** transforms/maps the ESM package (CSS stub, country data stub); **webpack** adds a **`.css`** rule and **`.webp`** assets for the library. **SCSS** styles `.iti` / `.iti-wrapper` for floating labels. Renames enum typo **`DROPDOWM` → `DROPDOWN`** in a few call sites.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 78c50cc6ec77ef89e2fe1001c7c013cfb92faef6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


bc-maxy:
cursor review

bc-maxy:
For the optional phone number case, maybe I should relax validation rules and remove the valid phone check (see optional case video)? And then keep it only for the case when the phone is required.. What do you think?

richapatel1510:
I think the implementation for optional case is as expected, we should validate the phone number if it is entered.

packages/ui/src/form/DynamicFormField/DynamicInput.tsx
```diff
@@ -28,14 +31,18 @@ export interface DynamicInputProps extends InputProps {
...
     isFloatingLabelEnabled?: boolean;
     inputDateFormat?: string;
+    selectedCountry?: string;
+    isNewPhoneFieldWithValidation?: boolean;
 }
 
 const DynamicInput: FunctionComponent<DynamicInputProps & WithDateProps> = ({
     fieldType,
     id,
+    itiRef,
```
bc-peng: Can we name it `intlTelInputRef` as `itiRef` may not explain itself well?

packages/ui/src/form/DynamicFormField/DynamicInput.tsx
```diff
@@ -44,10 +51,24 @@ const DynamicInput: FunctionComponent<DynamicInputProps & WithDateProps> = ({
...
     ...rest
 }) => {
     const inputFormat = inputDateFormat || date.inputFormat || '';
 
+    useEffect(() => {
+        // auto-set phone number country based on shipping address
+        if (isNewPhoneFieldWithValidation) {
+            if (!selectedCountry || value) return;
+
+            const selectedCountryInIsoFormat = selectedCountry.toLowerCase() as Iso2;
```
bc-peng: Is it possible to avoid `as Iso2`?

packages/ui/.eslintrc.json
```diff
@@ -5,6 +5,7 @@
     "@typescript-eslint/naming-convention": "off",
     "@typescript-eslint/consistent-type-assertions": "off",
     "@typescript-eslint/no-unsafe-call": "off",
-    "@nx/enforce-module-boundaries": "off"
+    "@nx/enforce-module-boundaries": "off",
+    "complexity": "off" // TODO: refactor DynamicInput and remove
```
bc-peng: Can we refactor it now or create a PR to refactor it before merging this PR?

packages/ui/src/form/DynamicFormField/DynamicInput.test.tsx
```diff
@@ -1,12 +1,44 @@
...
+        return (
+            <input
+                data-test="iti-phone-input"
+                {...inputProps}
+                onChange={(e) => onChangeNumber?.(e.target.value)}
+                value={value ?? ''}
+            />
+        );
+    }),
+}));
```
richapatel1510: 💅 Can move it out as a shared test util outside so can use it in both `DynamicFormField.test.tsx` and `DynamicInput.test.tsx`

packages/ui/.eslintrc.json
richapatel1510: If not, then may be just add a line-level disable in `DynamicInput` and not the whole package?

packages/ui/src/form/DynamicFormField/DynamicFormField.tsx
```diff
@@ -111,24 +144,30 @@ const DynamicFormField: FunctionComponent<DynamicFormFieldProps> = ({
                 fieldType={dynamicFormFieldType}
                 id={fieldInputId}
                 isFloatingLabelEnabled={isFloatingLabelSupportedFieldType}
+                isNewPhoneFieldWithValidation={isNewPhoneFieldWithValidation}
```
animesh1987: This feels fine, but maybe we can have a future ticket to have a specific component for phone number that we can load at parent level? Possibly a fast follow refactor, so we can move to have a specific component.

packages/ui/src/form/DynamicFormField/DynamicFormFieldType.ts
```diff
@@ -1,6 +1,6 @@
 enum DynamicFormFieldType {
     TELEPHONE = 'tel',
-    DROPDOWM = 'dropdown',
```
animesh1987, line 3: 😏

packages/ui/src/form/DynamicFormField/DynamicInput.tsx
```diff
@@ -44,10 +51,24 @@ const DynamicInput: FunctionComponent<DynamicInputProps & WithDateProps> = ({
...
+    selectedCountry,
+    isNewPhoneFieldWithValidation = false,
     ...rest
 }) => {
     const inputFormat = inputDateFormat || date.inputFormat || '';
 
+    useEffect(() => {
+        // auto-set phone number country based on shipping address
+        if (isNewPhoneFieldWithValidation) {
+            if (!selectedCountry || value) return;
```
animesh1987:
```suggestion
            if (!selectedCountry || value) {
                return;
            }
```

packages/ui/src/form/DynamicFormField/DynamicInput.tsx
animesh1987: Please use type guards and avoid `as`.

packages/ui/src/form/DynamicFormField/DynamicFormFieldSelector.tsx
```diff
@@ -0,0 +1,116 @@
...
+                autocomplete,
+                dynamicFormFieldType,
+                isFloatingLabelEnabled,
+                max,
+                maxLength,
+                min,
+                options,
+                placeholder,
+            ],
+        );
```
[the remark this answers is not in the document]
bc-maxy: Rendering this conditionally would break rules of hooks. We could extract the logic into a yet another new component but I think we are over-optimizing at this point and the callback should be extremely cheap 😅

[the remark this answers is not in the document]
bc-maxy: This is not a change introduced in this branch btw, that's existing behavior

packages/ui/.eslintrc.json
bc-maxy: Refactored in https://github.com/bigcommerce/checkout-js/pull/3025/commits/0c1d212f44170a032afebcf537be602a7be26425 😇

packages/ui/src/form/DynamicFormField/DynamicFormField.tsx
bc-maxy: Refactored https://github.com/bigcommerce/checkout-js/pull/3025/commits/0c1d212f44170a032afebcf537be602a7be26425 :)

packages/ui/src/form/PhoneFormField/PhoneFormField.tsx
```diff
@@ -0,0 +1,84 @@
+import { type IntlTelInputRef } from '@intl-tel-input/react';
+import 'intl-tel-input/styles';
```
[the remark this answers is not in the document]
bc-maxy: Oops 😄

packages/ui/src/form/DynamicFormField/DynamicInput.test.tsx
bc-maxy: I can't think of a way to use utils for this because of Jest hoisting, which is when Jest hoists jest.mock above all imports.
And so if we have a code like this:
```
import { intlTelInputMock } from '../mocks/intlTelInput.mock';

jest.mock('@intl-tel-input/react', () => intlTelInputMock);
```
The jest.mock is hoisted above intlTelInputMock import and then tests start failing with unresolved intlTelInputMock references. 

I think we can work around this by using manual mocks (in `__mocks__` folder) but then we would need to define `mockIsValidNumber` and `mockSetCountry` in this folder as well.. Which means that they would need to be imported by all relevant test files and we need to make sure the values are cleared between test scenarios. 
That's my understanding, but let me know if you have simpler suggestions! 
Anyways, I thought maybe it's not worth it for this case and the duplication is simpler and providing test isolation out of the box 😄

bc-maxy:
Thank you for all the comments, this should be ready for re-review @animesh1987 @richapatel1510 @bc-peng 🙏 🙏 🙏

packages/ui/src/form/PhoneFormField/PhoneInput.tsx
```diff
@@ -0,0 +1,72 @@
...
+    field: { name, value, onBlur, onChange },
+    id,
+    autocomplete,
+    maxLength,
+    isFloatingLabelEnabled,
+    selectedCountry,
+    intlTelInputRef,
+}) => {
+    useEffect(() => {
+        // auto-set phone number country based on shipping address
```
bc-peng:
```suggestion
```

This one seems self-explainable.

packages/ui/src/form/PhoneFormField/PhoneFormField.tsx
```diff
@@ -0,0 +1,83 @@
...
+        isFloatingLabelEnabled,
+        selectedCountry,
+        onChange,
+    }) => {
+        const { language } = useLocale();
+        const intlTelInputRef = useRef<IntlTelInputRef>(null);
+
+        const validatePhone = useCallback(
+            (value: string) => {
+                let isPhoneNumberValid: boolean | null | undefined;
```
bc-peng: I see it is only being set once.
Can we consider `const` by moving L47-51 to the `try` block?

packages/ui/src/form/DynamicFormField/DynamicFormField.test.tsx
```diff
@@ -1,13 +1,47 @@
...
+import { fireEvent, render, screen, waitFor } from '@bigcommerce/checkout/test-utils';
+
+import { FormProvider } from '../contexts';
 
 import DynamicFormField from './DynamicFormField';
 
+const mockIsValidNumber = jest.fn();
+const mockSetCountry = jest.fn();
+
+jest.mock('@intl-tel-input/react', () => ({
```
bc-peng, line 17: I see `jest.mock('@intl-tel-input/react', () => ({` appears twice.
Is it possible that we don't mock it and use the real lib to do the testing?

packages/ui/src/form/DynamicFormField/DynamicFormField.tsx
```diff
@@ -102,35 +103,9 @@ const DynamicFormField: FunctionComponent<DynamicFormFieldProps> = ({
...
-            max,
-            maxLength,
-            min,
-            options,
-            placeholder,
-        ],
-    );
+    const isNewPhoneFieldWithValidation =
+        isNewPhoneValidationExperimentEnabled &&
+        dynamicFormFieldType === DynamicFormFieldType.TELEPHONE;
```
richapatel1510: This can probably be moved to `DynamicFormFieldSelector` now?

packages/ui/src/form/PhoneFormField/PhoneInput.tsx
```diff
@@ -0,0 +1,71 @@
...
+            return;
+        }
+
+        const selectedCountryInIsoFormat = selectedCountry.toLowerCase();
+
+        if (isIso2(selectedCountryInIsoFormat)) {
+            intlTelInputRef.current?.getInstance()?.setCountry(selectedCountryInIsoFormat);
+        }
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [selectedCountry]);
```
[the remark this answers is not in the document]
bc-maxy: This is intentional 🙂

packages/ui/src/form/DynamicFormField/DynamicFormField.test.tsx
bc-maxy: The only reason for that is that this is an ESM library while our checkout-js project runs Jest in CommonJS mode, and so to make it work we need to add babel transforms which is not super-straightforward 😓 

I tried adding it here: https://github.com/bigcommerce/checkout-js/pull/3025/commits/6da7fc30f63e83d32bd2485389f0b462e415ff02, and that worked locally but not in CI (even though node / babel versions are similar).

Let me know if you have any ideas on how to fix! In the meantime - mocks. Which, on the positive, makes it a proper unit test that isolates and tests the component logic. On the negative, it is quite verbose.

bc-peng: CI is broken likely due to CPU.

image

Thanks for trying. We can see mocking is a logical choice now.

bc-peng:
The PR looks good to me.
Just let @bc-charlesho know this PR will add a subtle animation to the text field when a phone number is incorrect.
Please refer to the second video above(around 00:26) for the details.

bc-maxy:
> The PR looks good to me. Just let @bc-charlesho know this PR will add a subtle animation to the text field when a phone number is incorrect. Please refer to the second video above(around 00:26) for the details.

Thank you for reviewing @bc-peng 🙏  @bc-charlesho this animation is also configurable on the library side, so can be removed if needed

bc-maxy:
@bc-peng @bc-charlesho No more field explosion animation 💥💥💥 it's now removed:)

bc-peng (dismissed):
Nice work👏👏👏

richapatel1510 (dismissed):
👏 👏

packages/ui/src/form/PhoneFormField/PhoneInput.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+    return (
+        <span className="iti-wrapper">
+            <IntlTelInput
+                inputProps={{
+                    'aria-labelledby': `${id}-label ${id}-field-error-message`,
+                    autoComplete: autocomplete,
```
animesh1987, line 53: What does autocomplete do for this input?

bc-maxy: In this case it just gets passed down to the actual input component with the value "tel" (https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete) which.. I assume says to a browser that "this is a phone field" and so browser can pre-fill this with phone data (when saved for a user)

packages/ui/src/form/PhoneFormField/PhoneInput.tsx
```diff
@@ -0,0 +1,76 @@
...
+                        /* webpackChunkName: "intl-tel-input-utils" */
+                        'intl-tel-input/utils'
+                    )
+                }
+                onChangeNumber={handleChangeNumber}
+                ref={intlTelInputRef}
+                separateDialCode={false}
+                strictRejectAnimation={false}
+                value={value ? String(value) : ''}
+            />
```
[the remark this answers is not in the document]
bc-maxy: This is expected - unfortunately there is no easy way to determine what initialCountry should be passed for a user, so the useEffect is essentially a workaround for this.

richapatel1510 (approved):

bc-peng (approved):

bc-launchbay:
success [20260610012340](https://launchbay.bigcommerce.net/projects/535/releases/287549) by Launchbay

bc-launchbay:
success [20260610012340](https://launchbay.bigcommerce.net/projects/535/releases/287549) by Launchbay

bc-launchbay:
success [20260610012340](https://launchbay.bigcommerce.net/projects/535/releases/287549) by Launchbay

PR #3038, opened 2026-05-27
feat(checkout): CHECKOUT-9823 billing and shipping extra fields enhancements

richapatel1510:
## What/Why?

CHECKOUT-9823: Billing and shipping enhancements for B2B address-book customers.

- **Auto-select default address on load**: when the `hasCompanyAddressBook` capability is enabled and the checkout has no billing/shipping address set, `Billing.tsx` and `Shipping.tsx` now look up the customer's `isDefaultBilling` / `isDefaultShipping` address and apply it during initialization.
- **Filter address selector by type**: `SearchableAddressSelectComponent` now filters the address list by `isBilling` / `isShipping` based on the current `AddressType` before applying the search query, so the billing selector only shows billing-flagged addresses and shipping only shows shipping-flagged ones.
- **Drop session-storage extra-field merging in `AddressSelect`**: the `storageKey` prop and `getAddressWithExtraFields` helper are removed from `AddressSelect`. Extra-field merging from session storage now happens centrally inside `mapAddressToFormValues`, which also handles the SDK's `B2B_EXTRA_FIELD_PREFIX` mismatch between schema field names and raw `extraFields[i].fieldId`s. Callers in `BillingForm` and `ShippingAddressForm` no longer pass `storageKey`.
- **Re-sync billing form values after address update**: with `enableReinitialize: true`, the form rehydrates via `mapPropsToValues` whenever the billing address prop changes. `getAddressWithCustomerExtraFields` grafts the matching customer address's extraFields onto the billing address (the SDK strips them on updateBillingAddress), so extra fields and other address-derived values stay in sync with the freshly applied address.
- **Render `address.label` in `StaticAddress`**: when an address has a `label`, it is rendered as a new line in the static summary.
- **Bump `@bigcommerce/checkout-sdk` to `^1.918.0`** (required for `B2B_EXTRA_FIELD_PREFIX` export).
- Adds `getB2BCustomer()` mock with mixed billing/shipping flag combinations to back the new test cases in `AddressSelect`, `SearchableAddressSelectComponent`, and `mapAddressToFormValues`.

## Test plan

- [x] Unit tests pass: `AddressSelect.test.tsx`, `SearchableAddressSelectComponent.test.tsx`, `mapAddressToFormValues.test.ts`.
- [x] As a B2B customer with `hasCompanyAddressBook` enabled and no prior billing/shipping address on the cart, load checkout and confirm the default billing and default shipping addresses are auto-applied.
- [x] In the billing address selector, confirm only addresses with `isBilling: true` are listed; in the shipping selector, only `isShipping: true`.
- [x] Select an address with extra fields from the address book on billing — confirm the form re-renders with the extra-field values populated and submits successfully.
- [x] Pre-populate session-storage extra fields, then load checkout — confirm `mapAddressToFormValues` falls back to stored values when the API address has none, and that API values still win when present.
- [x] Confirm an address with a `label` renders the label line in the static address summary on the review step.
- [x] Regression: non-B2B (guest / no `hasCompanyAddressBook`) checkout flow behaves identically to before.

## Rollout/Rollback

Revert this PR

## Testing

- CI checks
- Manual testing (TBA)

**1. B2B Extra Fields Happy Path**

https://github.com/user-attachments/assets/f31ec4b7-98de-4080-8b77-8a5226c24a05

**2. B2B Valid/Invalid Extra Fields behaviour**

https://github.com/user-attachments/assets/0e75649c-8d5d-485e-9ae4-afe72051d5c3

**3. B2C OOPC Flow**

https://github.com/user-attachments/assets/d2bccc3e-930a-4318-adeb-c6f6a1695c8a

---

> [!NOTE]
> **Medium Risk**
> Touches billing/shipping initialization and B2B extra-field/session-storage behavior across checkout steps; failures on default-address apply are non-blocking but could leave empty addresses in edge cases.
> 
> **Overview**
> Improves **B2B company address book** checkout: default addresses on load, type-filtered selectors, and a clearer path for **extra fields** and **labels** without merging session storage at address selection.
> 
> When **`hasCompanyAddressBook`** is on and billing/shipping is empty, **`Billing`** and **`Shipping`** apply the customer’s **`isDefaultBilling`** / **`isDefaultShipping`** address during init (failures are swallowed so checkout isn’t blocked). **`SearchableAddressSelectComponent`** filters options by **`isBilling`** / **`isShipping`** before search.
> 
> **`AddressSelect`** no longer takes **`storageKey`** or merges extra fields on pick; **`mapAddressToFormValues`** now maps raw **`fieldId`** values (stripping **`B2B_EXTRA_FIELD_PREFIX`**), prefers address values over session storage, and forms still pass billing/shipping storage keys from **`BillingForm`** / **`SingleShippingForm`**. **`getAddressWithCustomerExtraFields`** grafts book **`extraFields`** onto checkout billing/shipping addresses when the SDK doesn’t round-trip them; **`getAddressWithLabel`** and **`StaticAddress`** show **`label`**. **`isEqualAddress`** ignores B2B metadata so re-select and comparisons stay stable. “Use new address” clears B2B session keys where **`hasAddressExtraFields`** applies.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4059947798da0e42d33c4ce8d59029202f8d0a38. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/app/address/AddressSelect.tsx
+++ b/packages/core/src/app/address/AddressSelect.tsx
@@ -75,7 +75,6 @@ const AddressSelect = ({
                     }
                 >
                     <AddressSelectButton
-                        addresses={addresses}
                         placeholderText={placeholderText}
                         selectedAddress={selectedAddress}
                         showSingleLineAddress={showSingleLineAddress}
--- a/packages/core/src/app/address/AddressSelect.tsx
+++ b/packages/core/src/app/address/AddressSelect.tsx
@@ -11,7 +11,6 @@ import { DropdownTrigger } from '@bigcommerce/checkout/ui';
 import AddressSelectButton from './AddressSelectButton';
 import { AddressSelectComponent } from './AddressSelectComponent';
 import type AddressType from './AddressType';
-import { B2BExtraFieldsSessionStorage } from './B2BExtraFieldsSessionStorage';
 import isEqualAddress from './isEqualAddress';
 import { SearchableAddressSelectComponent } from './SearchableAddressSelectComponent';
 
@@ -22,7 +21,6 @@ export interface AddressSelectProps {
     selectedAddress?: Address;
     type: AddressType;
     showSingleLineAddress?: boolean;
-    storageKey?: string;
     onSelectAddress(address: Address): void;
     onUseNewAddress(currentAddress?: Address): void;
     placeholderText?: ReactNode;
@@ -33,7 +31,6 @@ const AddressSelect = ({
     selectedAddress,
     type,
     showSingleLineAddress,
-    storageKey,
     onSelectAddress,
     onUseNewAddress,
     placeholderText,
@@ -44,10 +41,8 @@ const AddressSelect = ({
     const { shouldShowPayPalFastlaneLabel } = usePayPalFastlaneAddress();
 
     const handleSelectAddress = (newAddress: Address) => {
-        const addressWithExtraFields = getAddressWithExtraFields(newAddress, storageKey);
-
-        if (!isEqualAddress(selectedAddress, addressWithExtraFields)) {
-            onSelectAddress(addressWithExtraFields);
+        if (!isEqualAddress(selectedAddress, newAddress)) {
+            onSelectAddress(newAddress);
         }
     };
 
@@ -93,21 +88,5 @@ const AddressSelect = ({
         </div>
     );
 };
-
-function getAddressWithExtraFields(address: Address, storageKey?: string): Address {
-    if (!storageKey) return address;
-
-    const storedExtraFields = B2BExtraFieldsSessionStorage.getFields(storageKey);
-
-    if (!storedExtraFields) return address;
-
-    return {
-        ...address,
-        extraFields: Object.entries(storedExtraFields).map(([fieldId, fieldValue]) => ({
-            fieldId,
-            fieldValue: typeof fieldValue === 'number' ? fieldValue : String(fieldValue),
-        })),
-    };
-}
 
 export default memo(AddressSelect);
--- a/packages/core/src/app/address/getAddressWithCustomerExtraFields.ts
+++ b/packages/core/src/app/address/getAddressWithCustomerExtraFields.ts
@@ -1,0 +1,31 @@
+import { type Address, type CustomerAddress } from '@bigcommerce/checkout-sdk';
+
+import isEqualAddress from './isEqualAddress';
+
+/**
+ * `updateBillingAddress` / `updateShippingAddress` don't round-trip B2B
+ * `extraFields` through the checkout state. Look up the matching customer
+ * address book entry and graft its `extraFields` onto the current address so
+ * the form and validation see the values. No-op for B2C
+ * (`hasAddressExtraFields=false`) and when the address already carries its
+ * own `extraFields`.
+ */
+export default function getAddressWithCustomerExtraFields<T extends Address>(
+    address: T | undefined,
+    customerAddresses: CustomerAddress[] | undefined,
+    hasAddressExtraFields: boolean,
+): T | undefined {
+    if (!hasAddressExtraFields || !address) {
+        return address;
+    }
+
+    if (address.extraFields?.length) {
+        return address;
+    }
+
+    const match = customerAddresses?.find((customerAddress) =>
+        isEqualAddress(customerAddress, address),
+    );
+
+    return match?.extraFields?.length ? { ...address, extraFields: match.extraFields } : address;
+}
--- a/packages/core/src/app/address/getAddressWithLabel.ts
+++ b/packages/core/src/app/address/getAddressWithLabel.ts
@@ -10,6 +10,12 @@ export default function getAddressWithLabel<T extends Address>(
         return address;
     }
 
+    // We match only against the locally loaded `customerAddresses`, not the
+    // full server-side company address book. A company may have many more
+    // addresses on the server than are returned here, so an address selected
+    // from the full book may not appear in this subset and will render without
+    // a label. We accept that gap rather than round-tripping to the API to
+    // resolve a purely cosmetic label.
     const match = customerAddresses.find((customerAddress) =>
         isEqualAddress(address, customerAddress),
     );
--- a/packages/core/src/app/address/getAddressWithLabel.ts
+++ b/packages/core/src/app/address/getAddressWithLabel.ts
@@ -1,0 +1,18 @@
+import { type Address, type CustomerAddress } from '@bigcommerce/checkout-sdk';
+
+import isEqualAddress from './isEqualAddress';
+
+export default function getAddressWithLabel<T extends Address>(
+    address: T,
+    customerAddresses: CustomerAddress[] = [],
+): T {
+    if (address.label) {
+        return address;
+    }
+
+    const match = customerAddresses.find((customerAddress) =>
+        isEqualAddress(address, customerAddress),
+    );
+
+    return match?.label ? { ...address, label: match.label } : address;
+}
--- a/packages/core/src/app/billing/BillingForm.tsx
+++ b/packages/core/src/app/billing/BillingForm.tsx
@@ -56,6 +56,7 @@ const BillingForm = ({
     getFields,
     billingAddress,
     setFieldValue,
+    setValues,
     values,
     onUnhandledError,
 }: BillingFormProps & WithLanguageProps & FormikProps<BillingFormValues>) => {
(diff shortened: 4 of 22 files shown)


bc-0dp:
This only introduce default address for B2B customers not regular B2C customers right?

packages/core/src/app/address/AddressSelect.tsx
```diff
@@ -94,20 +88,4 @@ const AddressSelect = ({
...
-
-    return {
-        ...address,
-        extraFields: Object.entries(storedExtraFields).map(([fieldId, fieldValue]) => ({
-            fieldId,
-            fieldValue: typeof fieldValue === 'number' ? fieldValue : String(fieldValue),
-        })),
-    };
-}
-
```
richapatel1510, line 112: Removed the `B2BExtraFieldsSessionStorage` read from `AddressSelect.handleSelectAddress` — it was duplicating work already done elsewhere. The picked address comes from `customer.addresses`, which already includes `extraFields` from the address book API, so there's nothing to hydrate at selection time. Session storage is still read as a fallback inside `mapAddressToFormValues` (for unsaved in-progress edits), and `Billing.tsx / useShipping.ts` graft `extraFields` from the address book onto the active billing/shipping address to cover the SDK round-trip gap. Those two paths fully cover the cases the `AddressSelect` read was trying to handle.

packages/core/src/app/shipping/Shipping.tsx
```diff
@@ -79,6 +80,20 @@ function Shipping({
...
+                    );
+
+                    if (defaultShippingAddress) {
+                        try {
+                            await updateShippingAddress(defaultShippingAddress);
+                        } catch {
+                            /* Do nothing: we should not block shoppers from buying. */
+                        }
+                    }
+                }
```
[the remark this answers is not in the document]
richapatel1510: In practice this path is cold: the backend pre-fills `consignment.shippingAddress` with the customer's default address for signed-in B2B customers, so the `!shippingAddress` guard skips the auto-apply. The client-side branch is kept as a defensive fallback for cases where that backend guarantee doesn't hold (e.g. guest-to-signed-in mid-checkout).

packages/core/src/app/shipping/Shipping.tsx
```diff
@@ -79,6 +80,20 @@ function Shipping({
                     loadBillingAddressFields(),
                 ]);
 
+                if (hasCompanyAddressBook && !shippingAddress) {
```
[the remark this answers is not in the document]
richapatel1510: The backend normally pre-fills `consignment.shippingAddress` with the customer's default address, so this branch is cold in practice. Its kept only in case if a user signs in mid session and the address is not prefilled.

packages/core/src/app/address/AddressSelect.tsx
richapatel1510: Also based on current b2b behaviour, 
-> If someone selects an incomplete address from dropdown -> updates some extra fields -> then again selects same or different address from the selector -> the extra fields gets reset to the value from the selected address and doesn't include the values from session storage.

richapatel1510:
cursor review

packages/core/src/app/billing/BillingForm.tsx
```diff
@@ -120,6 +121,10 @@ const BillingForm = ({
     };
 
     const handleUseNewAddress = () => {
+        if (hasAddressExtraFields) {
```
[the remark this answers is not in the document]
richapatel1510: Not needed. Updated the description. 

About `meaning stale session-storage extra-field values from a previously edited address can leak in as fallback values when switching to a different address.`

Thanks for flagging, but this isn't reachable in tested flows:

1. `getAddressWithCustomerExtraFields` (packages/core/src/app/address/getAddressWithCustomerExtraFields.ts) grafts the matching customer address's `extraFields` onto the billing address before `mapPropsToValues` runs, so `extraFieldValue` is populated from the address itself.

2. Even for "skipped" optional extras, the backend stores an empty string, which short-circuits the `??` chain (`??` only falls through on `null`/`undefined`). Verified manually — selecting an address with an invalid/empty extra shows the address's own value, not session storage.

3. The fallback is only reachable when an address has no entry at all for a `fieldId`. Shipping has the same exposure (`SingleShippingForm.handleAddressSelect` doesn't clear `SHIPPING_KEY` either) and is accepted as-is.

Leaving the code unchanged. Happy to revisit if a real repro surfaces.

packages/core/src/app/address/getAddressWithLabel.ts
```diff
@@ -0,0 +1,18 @@
...
+): T {
+    if (address.label) {
+        return address;
+    }
+
+    const match = customerAddresses.find((customerAddress) =>
+        isEqualAddress(address, customerAddress),
+    );
+
+    return match?.label ? { ...address, label: match.label } : address;
```
bc-peng, line 23: Could you please doc the decision where we don't match against the API side of the company address book?

packages/core/src/app/shipping/Shipping.tsx
```diff
@@ -79,6 +80,20 @@ function Shipping({
...
+                    const defaultShippingAddress = customer.addresses?.find(
+                        ({ isDefaultShipping }) => isDefaultShipping,
+                    );
+
+                    if (defaultShippingAddress) {
+                        try {
+                            await updateShippingAddress(defaultShippingAddress);
+                        } catch {
+                            /* Do nothing: we should not block shoppers from buying. */
```
bc-peng, line 92: Just want to see the real case of this, could you add a video to show the behaviour?

bc-peng (dismissed):
Nice job.👏👏

packages/core/src/app/shipping/Shipping.tsx
richapatel1510: Ideally this would never happen because backend always sets default shipping address when available.

packages/core/src/app/address/getAddressWithLabel.ts
richapatel1510: https://github.com/bigcommerce/checkout-js/pull/3038/commits/d5261e61ada3c2a995dac3ee3a505e2d0dd47c41

bc-peng (dismissed):

packages/core/src/app/address/AddressSelect.tsx
animesh1987: Can we please capture this behaviour in the ADR.

packages/core/src/app/address/getAddressWithCustomerExtraFields.ts
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * `updateBillingAddress` / `updateShippingAddress` don't round-trip B2B
+ * `extraFields` through the checkout state. Look up the matching customer
+ * address book entry and graft its `extraFields` onto the current address so
+ * the form and validation see the values. No-op for B2C
+ * (`hasAddressExtraFields=false`) and when the address already carries its
+ * own `extraFields`.
+ */
+export default function getAddressWithCustomerExtraFields<T extends Address>(
+    address: T | undefined,
```
animesh1987, line 14: Why is address of generic type?

packages/core/src/app/address/getAddressWithCustomerExtraFields.ts
```diff
@@ -0,0 +1,31 @@
...
+    customerAddresses: CustomerAddress[] | undefined,
+    hasAddressExtraFields: boolean,
+): T | undefined {
+    if (!hasAddressExtraFields || !address) {
+        return address;
```
animesh1987, line 19: So here if 
- we don't have `hasAddressExtraFields` we return `address` 
- `address` is `undefined`, we return undefined
?

packages/core/src/app/address/getAddressWithCustomerExtraFields.ts
richapatel1510: The generic `<T extends Address>` preserves the caller's subtype. `getBillingAddress()` returns `BillingAddress` (extends `Address` with `email/id`), and shipping has its own variant — without the generic, the return would widen to `Address` and callers would lose those fields. The spread on line 30 keeps the runtime shape intact.

packages/core/src/app/address/getAddressWithCustomerExtraFields.ts
richapatel1510: Yes, that's correct. For `B2C` customers we don't have `hasAddressExtraFields`  so we return `address` as it is, also if no `address` its already `undefined` for `B2B/B2C` customers,

bc-peng (approved):

animesh1987 (approved):

bc-launchbay:
success [20260601055122](https://launchbay.bigcommerce.net/projects/535/releases/286618) by Launchbay

bc-launchbay:
success [20260601055122](https://launchbay.bigcommerce.net/projects/535/releases/286618) by Launchbay

bc-launchbay:
success [20260601055122](https://launchbay.bigcommerce.net/projects/535/releases/286618) by Launchbay

PR #3058, opened 2026-06-02
feat(checkout): CHECKOUT-10009 add credit limit checks on PO payment method

richapatel1510:
## What/Why?

Adds credit limit and currency validation to the Purchase Order (cheque) payment method to prevent buyers from checking out when their order would exceed the PO credit limit available on their account, or when the cart currency doesn't match the currency configured on the PO. Today the PO method is selectable in all cases, and any guard against credit limits happens server-side only after submission — this surfaces the constraint up-front in the UI.

The check reads `poConfig` (`{ creditLimit, currency }`) from the capabilities context. When the cheque method is in the list, a new hook `usePoMethodDisabledReason` computes one of two `PoDisabledReason` values:

- `creditLimit` — when `grandTotal > poConfig.creditLimit`
- `currencyMismatch` — when `poConfig.currency` differs from the cart currency (case-insensitive)

If a reason is returned, the cheque method is disabled in `PaymentMethodList` and an explanatory message is rendered in `PaymentMethodTitle`. The same hook is also called inside `PaymentForm` against the currently selected method so the submit button is blocked when a PO method is pre-selected (e.g. via `defaultMethod`). `Payment.tsx` falls back to `props.defaultMethod` when no method has been explicitly selected so the gate still applies on first render.

Pure logic lives in `getPoMethodDisabledReason` so it can be unit-tested without React/context. Two English strings were added under `payment.errors.disabled_PO_number_credit` and `payment.errors.disabled_PO_number_currency_mismatch`, plus matching `paymentProviderHeader-poDisabled*` styles.

## Rollout/Rollback

Its behind B2B feature flag is gating this change — the check is driven entirely by the presence of `poConfig` in the capabilities response. 
Rollback is code-only: reverting this PR removes the disabled-state UI and the submit gate.

## Testing

- CI checks
- Manual testing (TBA)

**B2B flow**

https://github.com/user-attachments/assets/a1b817e9-3f73-4a8a-993d-12ca5d1a67f0

**OOPC B2C flow**

https://github.com/user-attachments/assets/b68e42cf-3f3e-4a68-87d4-9052a9a86e28

---

> [!NOTE]
> **Medium Risk**
> Changes checkout payment selection and submit gating for B2B PO flows; logic is guarded by incomplete/missing `poConfig` and cart data but affects order completion paths when enabled.
> 
> **Overview**
> **Purchase Order (cheque) payment** is now gated in checkout using `poConfig` from capabilities (`creditLimit`, `currency`). A new `getPoMethodDisabledReason` / `usePoMethodDisabledReason` flow disables the cheque option when the cart currency does not match (case-insensitive, checked first) or when `grandTotal` exceeds the credit limit; explanatory copy and styles were added under `payment.errors.disabled_PO_*`.
> 
> The **payment step** wires this through the list (`PaymentMethodList` / `PaymentMethodTitle`), blocks **Place Order** when the selected method is invalid (`PaymentForm`), and treats **`defaultMethod`** as the effective selection in `Payment` so pre-selected PO is covered on first paint. Unit tests cover the pure disable logic.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 22e89e952e51b1cb8fea45a0096bd69c5899ba19. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/app/payment/PaymentForm.tsx
+++ b/packages/core/src/app/payment/PaymentForm.tsx
@@ -125,8 +125,8 @@ const PaymentForm: FunctionComponent<
 
     const { checkoutState } = useCheckout();
     const { checkoutSettings } = checkoutState.data.getConfig() ?? {};
-    const selectedMethodDisabledReason = usePoMethodDisabledReason(selectedMethod);
-    const isSubmitDisabled = shouldDisableSubmit || Boolean(selectedMethodDisabledReason);
+    const poMethodDisabledReason = usePoMethodDisabledReason(selectedMethod);
+    const isSubmitDisabled = shouldDisableSubmit || Boolean(poMethodDisabledReason);
     const shouldShowSubmitButtonWhenPaymentNotRequired = isExperimentEnabled(
         checkoutSettings,
         'CHECKOUT-9729.show_submit_button_when_payment_not_required',
--- a/packages/core/src/app/payment/PaymentForm.tsx
+++ b/packages/core/src/app/payment/PaymentForm.tsx
@@ -31,6 +31,7 @@ import {
     getUniquePaymentMethodId,
     PaymentMethodId,
     PaymentMethodList,
+    usePoMethodDisabledReason,
 } from './paymentMethod';
 import PaymentRedeemables from './PaymentRedeemables';
 import PaymentSubmitButton from './PaymentSubmitButton';
@@ -124,6 +125,8 @@ const PaymentForm: FunctionComponent<
 
     const { checkoutState } = useCheckout();
     const { checkoutSettings } = checkoutState.data.getConfig() ?? {};
+    const selectedMethodDisabledReason = usePoMethodDisabledReason(selectedMethod);
+    const isSubmitDisabled = shouldDisableSubmit || Boolean(selectedMethodDisabledReason);
     const shouldShowSubmitButtonWhenPaymentNotRequired = isExperimentEnabled(
         checkoutSettings,
         'CHECKOUT-9729.show_submit_button_when_payment_not_required',
@@ -207,7 +210,7 @@ const PaymentForm: FunctionComponent<
                             selectedMethod && selectedMethod.initializationStrategy?.type
                         }
                         isComplete={!!selectedMethod?.initializationData?.isComplete}
-                        isDisabled={shouldDisableSubmit}
+                        isDisabled={isSubmitDisabled}
                         methodGateway={selectedMethod && selectedMethod.gateway}
                         methodId={selectedMethodId}
                         methodName={
--- a/packages/core/src/app/payment/paymentMethod/PaymentMethodList.tsx
+++ b/packages/core/src/app/payment/paymentMethod/PaymentMethodList.tsx
@@ -13,6 +13,7 @@ import getPaymentMethodName from './getPaymentMethodName';
 import getUniquePaymentMethodId, { parseUniquePaymentMethodId } from './getUniquePaymentMethodId';
 import PaymentMethodTitle, { getPaymentMethodTitle } from './PaymentMethodTitle';
 import PaymentMethodV2 from './PaymentMethodV2';
+import { type PoDisabledReason, usePoMethodDisabledReason } from './usePoMethodDisabledReason';
 
 export interface PaymentMethodListProps {
     isEmbedded?: boolean;
@@ -53,6 +54,9 @@ const PaymentMethodList: FunctionComponent<
     } = useCheckout();
 
     const config = getConfig();
+
+    const chequeMethod = useMemo(() => find(methods, { id: 'cheque' }), [methods]);
+    const chequeDisabledReason = usePoMethodDisabledReason(chequeMethod);
 
     const titleText = useMemo(() => {
         if (config && values.paymentProviderRadio) {
@@ -109,6 +113,9 @@ const PaymentMethodList: FunctionComponent<
 
                     return (
                         <PaymentMethodListItem
+                            disabledReason={
+                                method === chequeMethod ? chequeDisabledReason : undefined
+                            }
                             isDisabled={isInitializingPayment}
                             isEmbedded={isEmbedded}
                             isUsingMultiShipping={isUsingMultiShipping}
@@ -125,6 +132,7 @@ const PaymentMethodList: FunctionComponent<
 };
 
 interface PaymentMethodListItemProps {
+    disabledReason?: PoDisabledReason;
     isDisabled?: boolean;
     isEmbedded?: boolean;
     isUsingMultiShipping?: boolean;
@@ -134,6 +142,7 @@ interface PaymentMethodListItemProps {
 }
 
 const PaymentMethodListItem: FunctionComponent<PaymentMethodListItemProps> = ({
+    disabledReason,
     isDisabled,
     isEmbedded,
     isUsingMultiShipping,
@@ -155,12 +164,13 @@ const PaymentMethodListItem: FunctionComponent<PaymentMethodListItemProps> = ({
     const renderPaymentMethodTitle = useCallback(
         (isSelected: boolean) => (
             <PaymentMethodTitle
+                disabledReason={disabledReason}
                 isSelected={isSelected}
                 method={method}
                 onUnhandledError={onUnhandledError}
             />
         ),
-        [method],
+        [disabledReason, method],
     );
 
     if (method.initializationData?.isCustomChecklistItem) {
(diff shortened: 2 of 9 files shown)


packages/core/src/app/payment/paymentMethod/usePoMethodDisabledReason.ts
```diff
@@ -0,0 +1,74 @@
...
+    if (
+        method.id !== 'cheque' ||
+        !poConfig ||
+        poConfig.creditLimit == null ||
+        !poConfig.currency ||
+        grandTotal == null ||
+        !cartCurrencyCode
+    ) {
+        return null;
+    }
```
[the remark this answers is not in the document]
richapatel1510: This is expected. `creditLimit: null` means credit limit check is not configured and doesn't need to be considered.

packages/core/src/app/payment/paymentMethod/usePoMethodDisabledReason.ts
```diff
@@ -0,0 +1,74 @@
...
+    grandTotal,
+    cartCurrencyCode,
+}: GetPoMethodDisabledReasonArgs): PoDisabledReason | null {
+    if (
+        method.id !== 'cheque' ||
+        !poConfig ||
+        poConfig.creditLimit == null ||
+        !poConfig.currency ||
+        grandTotal == null ||
+        !cartCurrencyCode
```
bc-maxy: nit: this is a bit hard to read, what do you think about splitting this into several constants (or several conditions) to make it easier?

packages/core/src/app/payment/paymentMethod/usePoMethodDisabledReason.ts
```diff
@@ -0,0 +1,74 @@
...
+}
+
+export function usePoMethodDisabledReason(
+    method: PaymentMethod | undefined,
+): PoDisabledReason | undefined {
+    const {
+        checkoutState: {
+            data: { getCart, getCheckout },
+        },
+    } = useCheckout();
```
bc-maxy: Here we can use a selector with useCheckout to narrow subscription (similar to https://github.com/bigcommerce/checkout-js/pull/3056)

packages/core/src/app/payment/PaymentForm.tsx
```diff
@@ -124,6 +125,8 @@ const PaymentForm: FunctionComponent<
 
     const { checkoutState } = useCheckout();
     const { checkoutSettings } = checkoutState.data.getConfig() ?? {};
+    const selectedMethodDisabledReason = usePoMethodDisabledReason(selectedMethod);
```
bc-maxy: I think the name is a bit misleading because this const is purchase order specific.
Maybe `poMethodDisabledReason` or something like that?

bc-maxy (commented):
Looks great overall, but added a few comments!

packages/core/src/app/payment/paymentMethod/usePoMethodDisabledReason.ts
```diff
@@ -0,0 +1,74 @@
+import { type PaymentMethod } from '@bigcommerce/checkout-sdk';
+import { useMemo } from 'react';
+
+import { useCapabilities, useCheckout } from '@bigcommerce/checkout/contexts';
+
+export type PoDisabledReason =
+    | { kind: 'creditLimit' }
+    | { kind: 'currencyMismatch'; expectedCurrency: string };
```
bc-peng: Can we simplify the reason type to `string` here by getting `expectedCurrency` from the capability once the reason is determined?

The rest PR looks good.

packages/core/src/app/payment/paymentMethod/PaymentMethodList.tsx
```diff
@@ -54,6 +55,9 @@ const PaymentMethodList: FunctionComponent<
 
     const config = getConfig();
 
+    const chequeMethod = useMemo(() => find(methods, { id: 'cheque' }), [methods]);
+    const chequeDisabledReason = usePoMethodDisabledReason(chequeMethod);
```
animesh1987, line 59: Does this need to be a hook, this can simply be a util method that we pass cheque method and capabilities to?

animesh1987: I think we can just call this https://github.com/bigcommerce/checkout-js/pull/3058/changes#diff-81d8c422f5528716de4c81576ba4efcfc795d490f1ff5b3583f77f0e8b3484a8R66. Reduce one layer.

richapatel1510: Thats a good idea. But we are using this hook at 2 places here and in PaymentForm, so if we remove the hook we need to add useMemo and some duplicate logic in both the files.

packages/core/src/app/payment/paymentMethod/usePoMethodDisabledReason.ts
richapatel1510: https://github.com/bigcommerce/checkout-js/pull/3058/commits/83bf962bc19e8f50c0b7764c1071141a4c7803e4

packages/core/src/app/payment/PaymentForm.tsx
richapatel1510: https://github.com/bigcommerce/checkout-js/pull/3058/commits/83bf962bc19e8f50c0b7764c1071141a4c7803e4

packages/core/src/app/payment/paymentMethod/usePoMethodDisabledReason.ts
richapatel1510: https://github.com/bigcommerce/checkout-js/pull/3058/commits/83bf962bc19e8f50c0b7764c1071141a4c7803e4

packages/core/src/app/payment/paymentMethod/usePoMethodDisabledReason.ts
richapatel1510: https://github.com/bigcommerce/checkout-js/pull/3058/commits/83bf962bc19e8f50c0b7764c1071141a4c7803e4

packages/core/src/app/payment/paymentMethod/PaymentMethodList.tsx
animesh1987: That should be ok imo. Hook runs on every render as well 🤔 so if we memoize the value for a session it should be ok.

richapatel1510: We  will need to keep the same logic at both places to use the useCheckout & useCapability hooks, and getting all the data and then feed it to the function.

richapatel1510: https://github.com/bigcommerce/checkout-js/pull/3058/changes/BASE..365623b70c4c7b4e9d21f3a112c0050fa0058686#diff-1dd716d897bc91f0ece6c0fdfb716115419774327a43cf0bd2e46380a2918c68R59

&

https://github.com/bigcommerce/checkout-js/pull/3058/changes/BASE..365623b70c4c7b4e9d21f3a112c0050fa0058686#diff-ead2e1dd86d0270e470bc599425d898871244aeec6e7908c55622cc5522b0ae6R128

packages/core/src/app/payment/paymentMethod/PaymentMethodList.tsx
```diff
@@ -54,6 +55,9 @@ const PaymentMethodList: FunctionComponent<
 
     const config = getConfig();
 
+    const chequeMethod = useMemo(() => find(methods, { id: 'cheque' }), [methods]);
```
bc-maxy: You probably don't need a useMemo here? Unless you are planning on building on this logic and adding something besides `find` going forward?

bc-maxy (dismissed):
Looks great to me otherwise 🤩

animesh1987 (dismissed):

bc-peng (dismissed):

bc-maxy (approved):

bc-launchbay:
success [20260604004844](https://launchbay.bigcommerce.net/projects/535/releases/287073) by Launchbay

bc-launchbay:
success [20260604004844](https://launchbay.bigcommerce.net/projects/535/releases/287073) by Launchbay

bc-launchbay:
success [20260604004844](https://launchbay.bigcommerce.net/projects/535/releases/287073) by Launchbay

PR #3070, opened 2026-06-09
feat(checkout): CHECKOUT-10048 CHECKOUT-10050 disable gift certificate and coupon capabilities

bc-mackxu:
## Summary

- **CHECKOUT-10048**: Implement `userJourney.disableGiftCertificate` capability — when true, removes any applied gift certificates on checkout load, hides applied GC display, skips `applyGiftCertificate` in the code input, and updates the coupon form label to show "Coupon" only
- **CHECKOUT-10050**: Implement `userJourney.disableCoupon` capability — when true, removes any applied coupons on checkout load, skips `applyCoupon` in the code input, and updates the coupon form label to show "Gift certificate" only
- When **both** flags are true, the entire coupon/GC input form is hidden
- Bumps `@bigcommerce/checkout-sdk` to 1.922.0 which adds `disableGiftCertificate` and `disableCoupon` to the `userJourney` capability type

## What changed

| File | Change |
|------|--------|
| `useMultiCoupon.ts` | `applyCouponOrGiftCertificate` skips GC/coupon apply based on flags |
| `NewOrderSummarySubtotals.tsx` | Removes applied GCs/coupons on mount via `useEffect`; updated label; hides form when both disabled |
| `Redeemable.tsx` (payment/mobile) | Returns null when both disabled; updated label; `handleSubmit` respects flags |
| `PaymentRedeemables.tsx` | Passes capability flags down to `Redeemable` |
| `en.json` | No new keys needed — reuses `redeemable.coupon_text` and `redeemable.gift_certificate_text` |

## Notes for reviewers

- The removal logic runs in a `useEffect` with an empty dependency array (runs once on mount). Inline render was attempted first but caused a duplicate DELETE API call — `checkoutService.remove*` triggers a checkout state update synchronously during render, which React warns about (`Cannot update a component while rendering a different component`) and then re-renders the component, firing the removal a second time. `useEffect` avoids this by running after the render phase, same pattern as `disableStoreCredit` in `Payment.tsx`.
- The `eslint-disable react-hooks/exhaustive-deps` is intentional — capabilities are set once via GRPC at page load and never change, so empty deps is correct.
- Coupon field continues to work when only `disableGiftCertificate` is true (per ticket requirement).
- GC field continues to work when only `disableCoupon` is true.
- Removal is skipped on the order confirmation page (`isOrderConfirmation` guard) — order confirmation does not allow modifications so this is purely defensive.

## Test plan

- [x] Unit tests added for `useMultiCoupon`, `NewOrderSummarySubtotals`, and `Redeemable` covering all new flag combinations
- [x] Manual test: set `disableGiftCertificate: true` — label shows "Coupon", GC apply is skipped; test with a GC already applied (confirm it is removed with a single DELETE call) and without one (confirm no errors or unexpected API calls)
- [x] Manual test: set `disableGiftCertificate: true`, submit an **invalid** coupon code — confirm error message displays correctly (no GC fallback attempted)

https://github.com/user-attachments/assets/8d7edcf5-ec69-4653-8bf9-6a87e3e7415b

- [x] Manual test: set `disableCoupon: true` — label shows "Gift certificate", coupon apply is skipped; test with a coupon already applied (confirm it is removed with a single DELETE call) and without one (confirm no errors or unexpected API calls)
- [x] Manual test: set `disableCoupon: true`, submit an **invalid** GC code — confirm error propagates and displays (no coupon fallback attempted)
- [x] Manual test: each flag individually and both true — verify correct label, form visibility, and behaviour on both desktop and mobile (including payment step)

https://github.com/user-attachments/assets/4e0d9d5e-1107-4410-97cf-686d186ae65e

---

> [!NOTE]
> **Medium Risk**
> Changes checkout discount application paths (coupon vs gift certificate fallback) across order summary and payment redeemables; wrong flag handling could block valid codes or show incorrect UI, but behavior is covered by new unit tests.
> 
> **Overview**
> Wires **`userJourney.disableCoupon`** and **`userJourney.disableGiftCertificate`** from `useCapabilities` into cart checkout redeemable UI and apply logic.
> 
> **`useMultiCoupon.applyCouponOrGiftCertificate`** no longer always tries gift certificate then coupon: it skips GC when disabled, skips coupon fallback when only GC is allowed, and rethrows GC errors when coupons are disabled. **`Redeemable`** mirrors that submit behavior via new optional props, returns **null** when both are disabled, and uses **`getRedeemableLabelId`** for toggle copy (coupon-only, gift-certificate-only, or default). **`NewOrderSummarySubtotals`** hides the coupon section when both flags are true and picks the same label; **`PaymentRedeemables`** passes the capability flags into **`Redeemable`**.
> 
> Adds **`getRedeemableLabelId`** in coupon utils and unit tests for **`Redeemable`**, **`useMultiCoupon`**, and **`NewOrderSummarySubtotals`** covering labels, visibility, and which apply APIs run.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e7b41e8f758240317a4c1a78890a938e77142764. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/app/cart/Redeemable.tsx
+++ b/packages/core/src/app/cart/Redeemable.tsx
@@ -31,6 +31,8 @@ import {
     Toggle,
 } from '@bigcommerce/checkout/ui';
 
+import { getRedeemableLabelId } from '../coupon/utils';
+
 import AppliedRedeemables, { type AppliedRedeemablesProps } from './AppliedRedeemables';
 
 export interface RedeemableFormValues {
@@ -49,6 +51,8 @@ export type ReedemableChildrenProps = Pick<
 
 export type RedeemableProps = {
     appliedRedeemableError?: RequestError;
+    disableCoupon?: boolean;
+    disableGiftCertificate?: boolean;
     isApplyingRedeemable?: boolean;
     isRemovingRedeemable?: boolean;
     removedRedeemableError?: RequestError;
@@ -61,7 +65,19 @@ export type RedeemableProps = {
 
 const Redeemable: FunctionComponent<
     RedeemableProps & WithLanguageProps & FormikProps<RedeemableFormValues>
-> = ({ shouldCollapseCouponCode, showAppliedRedeemables, ...formProps }) => {
+> = ({
+    disableCoupon,
+    disableGiftCertificate,
+    shouldCollapseCouponCode,
+    showAppliedRedeemables,
+    ...formProps
+}) => {
+    if (disableCoupon && disableGiftCertificate) {
+        return null;
+    }
+
+    const toggleLabelId = getRedeemableLabelId(disableGiftCertificate, disableCoupon);
+
     return (
         <Toggle openByDefault={!shouldCollapseCouponCode}>
             {({ toggle, isOpen }): ReactNode => (
@@ -75,12 +91,12 @@ const Redeemable: FunctionComponent<
                             href="#"
                             onClick={preventDefault(toggle)}
                         >
-                            <TranslatedString id="redeemable.toggle_action" />
+                            <TranslatedString id={toggleLabelId} />
                         </a>
                     )}
                     {!shouldCollapseCouponCode && (
                         <div className="redeemable-label body-cta">
-                            <TranslatedString id="redeemable.toggle_action" />
+                            <TranslatedString id={toggleLabelId} />
                         </div>
                     )}
                     {(isOpen || !shouldCollapseCouponCode) && (
@@ -235,19 +251,35 @@ export default withLanguage(
 
         async handleSubmit(
             { redeemableCode },
-            { props: { applyCoupon, applyGiftCertificate, clearError } },
+            {
+                props: {
+                    applyCoupon,
+                    applyGiftCertificate,
+                    clearError,
+                    disableCoupon,
+                    disableGiftCertificate,
+                },
+            },
         ) {
             const code = redeemableCode.trim();
 
-            try {
-                await applyGiftCertificate(code);
-            } catch (error) {
-                if (error instanceof Error) {
-                    clearError(error);
-                }
+            if (!disableGiftCertificate) {
+                try {
+                    await applyGiftCertificate(code);
 
-                applyCoupon(code);
+                    return;
+                } catch (error) {
+                    if (disableCoupon) {
+                        throw error;
+                    }
+
+                    if (error instanceof Error) {
+                        clearError(error);
+                    }
+                }
             }
+
+            await applyCoupon(code);
         },
 
         validationSchema({ language }: RedeemableProps & WithLanguageProps) {
--- a/packages/core/src/app/coupon/NewOrderSummarySubtotals.tsx
+++ b/packages/core/src/app/coupon/NewOrderSummarySubtotals.tsx
@@ -1,6 +1,7 @@
 import type { Fee, OrderFee, Tax } from '@bigcommerce/checkout-sdk';
 import React, { type FunctionComponent, useRef, useState } from 'react';
 
+import { useCapabilities } from '@bigcommerce/checkout/contexts';
 import { preventDefault } from '@bigcommerce/checkout/dom-utils';
 import { TranslatedString } from '@bigcommerce/checkout/locale';
 import { CollapseCSSTransition } from '@bigcommerce/checkout/ui';
@@ -9,6 +10,7 @@ import { isOrderFee, OrderSummaryDiscount, OrderSummaryPrice } from '../order';
 
 import { AppliedGiftCertificates, CouponForm, Discounts } from './components';
 import { useMultiCoupon } from './useMultiCoupon';
+import { getRedeemableLabelId } from './utils';
 
 interface MultiCouponProps {
     fees?: Fee[] | OrderFee[];
@@ -35,6 +37,10 @@ const NewOrderSummarySubtotals: FunctionComponent<MultiCouponProps> = ({
         uiDetails: { shipping, shippingBeforeDiscount },
     } = useMultiCoupon();
 
+    const {
+        userJourney: { disableCoupon, disableGiftCertificate },
+    } = useCapabilities();
+
     const [isCouponFormVisible, setIsCouponFormVisible] = useState(!isCouponFormCollapsed);
     const couponFormRef = useRef<HTMLDivElement>(null);
 
@@ -44,7 +50,7 @@ const NewOrderSummarySubtotals: FunctionComponent<MultiCouponProps> = ({
 
     return (
         <>
-            {!isOrderConfirmation && (
+            {!isOrderConfirmation && !(disableCoupon && disableGiftCertificate) && (
                 <section className="cart-section optimizedCheckout-orderSummary-cartSection">
                     <a
                         aria-controls="coupon-form-collapsable"
@@ -54,7 +60,9 @@ const NewOrderSummarySubtotals: FunctionComponent<MultiCouponProps> = ({
                         href="#"
                         onClick={preventDefault(toggleCouponForm)}
                     >
-                        <TranslatedString id="redeemable.toggle_action" />
+                        <TranslatedString
+                            id={getRedeemableLabelId(disableGiftCertificate, disableCoupon)}
+                        />
                     </a>
 
                     <CollapseCSSTransition isVisible={isCouponFormVisible} nodeRef={couponFormRef}>
--- a/packages/core/src/app/coupon/useMultiCoupon.ts
+++ b/packages/core/src/app/coupon/useMultiCoupon.ts
@@ -1,7 +1,7 @@
 import { type Coupon } from '@bigcommerce/checkout-sdk';
 import { useState } from 'react';
 
-import { useCheckout, useLocale } from '@bigcommerce/checkout/contexts';
+import { useCapabilities, useCheckout, useLocale } from '@bigcommerce/checkout/contexts';
 
 import { EMPTY_ARRAY } from '../common/utility';
 import { hasSelectedShippingOptions } from '../shipping';
@@ -43,6 +43,9 @@ export const useMultiCoupon = (): UseMultiCouponValues => {
 
     const { checkoutState, checkoutService } = useCheckout();
     const { language } = useLocale();
+    const {
+        userJourney: { disableCoupon, disableGiftCertificate },
+    } = useCapabilities();
 
     const {
         data: { getConfig, getCheckout, getOrder },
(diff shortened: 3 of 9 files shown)


packages/core/src/app/coupon/NewOrderSummarySubtotals.tsx
```diff
@@ -35,6 +36,34 @@ const NewOrderSummarySubtotals: FunctionComponent<MultiCouponProps> = ({
...
+        }
+
+        if (disableCoupon) {
+            const coupons = checkoutState.data.getCoupons() ?? [];
+
+            for (const { code } of coupons) {
+                void checkoutService.removeCoupon(code);
+            }
+        }
+    }, []);
```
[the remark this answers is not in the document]
bc-mackxu: Not addressing — `multi_coupon_cart_checkout` is 100% rolled out, the legacy `OrderSummarySubtotals` path is never rendered in production.

packages/core/src/app/cart/Redeemable.tsx
```diff
@@ -75,12 +93,12 @@ const Redeemable: FunctionComponent<
...
                             onClick={preventDefault(toggle)}
                         >
-                            <TranslatedString id="redeemable.toggle_action" />
+                            <TranslatedString id={toggleLabelId} />
                         </a>
                     )}
                     {!shouldCollapseCouponCode && (
                         <div className="redeemable-label body-cta">
-                            <TranslatedString id="redeemable.toggle_action" />
+                            <TranslatedString id={toggleLabelId} />
```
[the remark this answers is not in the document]
bc-mackxu: Intentional — the `useEffect` fires on mount and removes applied GCs/coupons immediately. The brief flash before async removal completes is acceptable; adding conditional guards on `AppliedGiftCertificates`/`AppliedRedeemables` was explicitly decided against since the removal handles it.

packages/core/src/app/cart/Redeemable.tsx
```diff
@@ -61,7 +63,23 @@ export type RedeemableProps = {
...
+}) => {
+    if (disableCoupon && disableGiftCertificate) {
+        return null;
+    }
+
+    const toggleLabelId = disableGiftCertificate
+        ? 'redeemable.coupon_text'
+        : disableCoupon
+          ? 'redeemable.gift_certificate_text'
+          : 'redeemable.toggle_action';
```
bc-maxy: What do you think about moving this to a util function and avoiding the three-way ternary? This is also used in 2 places in the code - so could be a good candidate.

packages/core/src/app/coupon/NewOrderSummarySubtotals.tsx
```diff
@@ -35,6 +36,34 @@ const NewOrderSummarySubtotals: FunctionComponent<MultiCouponProps> = ({
         uiDetails: { shipping, shippingBeforeDiscount },
     } = useMultiCoupon();
 
+    const {
+        userJourney: { disableCoupon, disableGiftCertificate },
+    } = useCapabilities();
+
+    const { checkoutService, checkoutState } = useCheckout();
```
bc-maxy: To narrow checkout state subscription:

```suggestion
    const { checkoutService, checkoutState } = useCheckout(({ data }) => ({
        giftCertificates: data.getGiftCertificates(),
        coupons: data.getCoupons(),
    }));
```

packages/core/src/app/coupon/NewOrderSummarySubtotals.tsx
```diff
@@ -35,6 +36,34 @@ const NewOrderSummarySubtotals: FunctionComponent<MultiCouponProps> = ({
...
+    useEffect(() => {
+        if (isOrderConfirmation) {
+            return;
+        }
+
+        if (disableGiftCertificate) {
+            const giftCertificates = checkoutState.data.getGiftCertificates() ?? [];
+
+            for (const { code } of giftCertificates) {
+                void checkoutService.removeGiftCertificate(code);
```
bc-maxy: Can this fail? How do we handle this right now from user's perspective?

packages/core/src/app/cart/Redeemable.tsx
bc-mackxu: Good call — extracted to `getRedeemableLabelId(disableGiftCertificate, disableCoupon)` in `coupon/utils/` and used it in both `Redeemable.tsx` and `NewOrderSummarySubtotals.tsx`.

packages/core/src/app/coupon/NewOrderSummarySubtotals.tsx
bc-mackxu: Not applying — the `useEffect` has empty deps `[]` and reads state only once on mount, so subscription narrowing doesn't reduce re-renders here. The selector pattern would be useful if we were subscribing to ongoing state changes, but that's not the case.

packages/core/src/app/coupon/NewOrderSummarySubtotals.tsx
bc-mackxu: Yes it can fail (e.g. network error). Currently `void` silently discards the rejection — if removal fails, the GC/coupon stays applied but the user can no longer apply new ones via the form (since it's hidden or skipped). This is acceptable as a best-effort on-mount cleanup; the capability is set server-side and the page state will be consistent on a reload. Happy to add error handling if you'd prefer.

packages/core/src/app/coupon/useMultiCoupon.ts
```diff
@@ -69,15 +72,23 @@ export const useMultiCoupon = (): UseMultiCouponValues => {
...
+                }
 
-            await applyCoupon(code);
+                if (error instanceof Error) {
+                    await clearError(error);
+                }
+            }
         }
+
+        await applyCoupon(code);
```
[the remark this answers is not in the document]
bc-mackxu: Not a bug — when both flags are true the form is hidden entirely (`Redeemable` returns `null`, `NewOrderSummarySubtotals` skips rendering the input section), so `applyCouponOrGiftCertificate` is never reachable by the user. The function only runs when at least one flag is false.

packages/core/src/app/coupon/NewOrderSummarySubtotals.tsx
bc-maxy: This is to limit component re-renders which is driven by `useSyncExternalStore` and not `useEffect`?

bc-mackxu: Applied — my previous reply misread the concern. You're right, the selector narrows the `useSyncExternalStore` subscription so the component only re-renders when `giftCertificates` or `coupons` change, not on every checkout state update.

packages/core/src/app/coupon/NewOrderSummarySubtotals.tsx
bc-maxy: Sorry just for me to understand - if this code tries to remove the coupon and fails, we do not notify the user in any way, correct?
Do we reject the order on the backend in case this client-side removal fails and user tries to submit it?

bc-mackxu: This is for the checkout page load. If `userJourney.disableGiftCertificate` is true and the cart already has a gift certificate applied, we'll need to remove the gift certificate from the cart.

Since shoppers won't be able to apply gift certificates in this scenario, I think it's ok to remove it silently during checkout loaded.

animesh1987: Good question @bc-maxy, on the whole unapply flow. Since we already know on backend about capabilities shall we unapply any GC, SC before the page load. That way we can avoid roundtrips to the server.

cc @valentindellangela @bc-peng

bc-peng: >we do not notify the user in any way, correct?

Since this runs on checkout page load, showing an error before any action creates a poor shopper experience.

We may consider merging this PR first, as It also includes the required UI changes. Then remove the API calls after the backend work is ready.

packages/core/src/app/cart/Redeemable.tsx
```diff
@@ -235,19 +251,35 @@ export default withLanguage(
...
+            if (!disableGiftCertificate) {
+                try {
+                    await applyGiftCertificate(code);
 
-                applyCoupon(code);
+                    return;
+                } catch (error) {
+                    if (disableCoupon) {
+                        throw error;
+                    }
```
bc-peng, line 274: In order to keep the same logic, should this be?

```suggestion
                    if (!disableCoupon) {
                         await applyCoupon(code);
                    }
```

packages/core/src/app/coupon/useMultiCoupon.ts
```diff
@@ -69,15 +72,23 @@ export const useMultiCoupon = (): UseMultiCouponValues => {
...
-            }
+        if (!disableGiftCertificate) {
+            try {
+                await applyGiftCertificate(code);
+
+                return;
+            } catch (error) {
+                if (disableCoupon) {
+                    throw error;
+                }
```
bc-peng, line 83: Same.

packages/core/src/app/cart/Redeemable.tsx
bc-mackxu: if disableCoupon is true, it should throw error here and don't need apply coupon
if !disableCoupon, it will keep the same logic, clear error and apply coupon in line 282.

bc-peng: `throw error;` appears to be new.

bc-mackxu: Here is the manual test for existing behaviour:
Image

Image

Image

bc-mackxu: Yes, when disablecoupon is true, we don't need try apply coupon and throw error if GC is failed. The manual test video is covered that part 

image

bc-peng: Discussed offline.
`throw error` here doesn't change the existing shopper experience.

bc-peng (dismissed):

packages/core/src/app/coupon/NewOrderSummarySubtotals.tsx
animesh1987: @bc-peng maybe we hold this for now, and update the removal of any GC/SC first. Then we can come back to this, or we will just have to rework things.

bc-mackxu: @animesh1987 Just to clarify, do you mean we keep the removal logic in the FE for now and clean it up later once the backend handles the removal logic?

bc-mackxu: As discussed, we are going to move the removal logic to the backend when checkout page is loaded.

bc-peng (approved):

richapatel1510 (approved):

bc-launchbay:
success [20260612000841](https://launchbay.bigcommerce.net/projects/535/releases/287820) by Launchbay

bc-launchbay:
success [20260612000841](https://launchbay.bigcommerce.net/projects/535/releases/287820) by Launchbay

bc-launchbay:
success [20260612000841](https://launchbay.bigcommerce.net/projects/535/releases/287820) by Launchbay

PR #3075, opened 2026-06-10
feat(checkout): Migrate CBA MPGS to Resolver Configuration

henrikh-kovalenko:
## What/Why?

Migrate CBA MPGS payment method to the new resolver-based architecture by registering it in the HostedCreditCardPaymentMethod component resolver (packages/hosted-credit-card-integration).

The migration is gated behind the PI-4748.cba_resolver_configuration experiment:

Experiment ON — CBA MPGS routes through the new resolver flow (HostedCreditCardComponent), where createCBAMPGSPaymentStrategy is included conditionally.
Experiment OFF — CBA MPGS falls back to the legacy V1 flow (HostedCreditCardPaymentMethod in core), where createCBAMPGSPaymentStrategy is added to the integration list conditionally.
The experiment flag is checked via isExperimentEnabled in core and directly via checkoutSettings.features in the hosted-credit-card-integration package (where the utility is not available). Both conditional blocks are intentionally duplicated to make cleanup straightforward once the experiment is fully rolled out.

Added global functional for experiment 

## Rollout/Rollback
roll back this commit 

## Testing
without 3ds isHostedFormEnabled false

https://github.com/user-attachments/assets/1e6a2220-3134-4c0b-8ac8-09edb63462a1

with 3ds isHostedFormEnabled false

https://github.com/user-attachments/assets/0051827e-c690-49e0-a91b-e75a33107084

without 3ds isHostedFormEnabled true

https://github.com/user-attachments/assets/fee735a1-a32b-4c00-b516-ecc53f26f7b8

with 3ds isHostedFormEnabled true

https://github.com/user-attachments/assets/d040a04d-1863-4020-98cc-a4109ce6f82f

image

---

> [!NOTE]
> **Medium Risk**
> Changes which payment UI and SDK integration list run for CBA MPGS, but behavior is gated and the two paths are mutually exclusive for the strategy registration.
> 
> **Overview**
> **CBA MPGS** checkout routing is migrated behind **`PI-4748.cba_resolver_configuration`**, with complementary legacy vs resolver behavior so only one path owns **`createCBAMPGSPaymentStrategy`**.
> 
> Payment method resolution now supports an optional **`experiment`** on **`PaymentMethodResolveId`**. **`resolvePaymentMethod`** filters registry entries with **`isExperimentEnabled`** (underscore keys normalized to dots), and **`PaymentMethodV2`** passes **`checkoutSettings`** into that resolver. **`cba_mpgs`** is registered on the hosted-credit-card integration resolver only when the flag is on.
> 
> With the experiment **on**, checkout uses the hosted-credit-card integration and conditionally adds the CBA MPGS strategy there. With it **off**, resolution falls back to legacy **V1** and the core **`HostedCreditCardPaymentMethod`** keeps registering that strategy instead.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a7d6c3f8e45dfb917cbe076911b2fd79ab3e7e4d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/app/payment/paymentMethod/HostedCreditCardPaymentMethod.tsx
+++ b/packages/core/src/app/payment/paymentMethod/HostedCreditCardPaymentMethod.tsx
@@ -1,4 +1,3 @@
-import { createCBAMPGSPaymentStrategy } from '@bigcommerce/checkout-sdk/integrations/cba-mpgs';
 import { createCreditCardPaymentStrategy } from '@bigcommerce/checkout-sdk/integrations/credit-card';
 import {
     createCyberSourcePaymentStrategy,
@@ -46,7 +45,6 @@ const HostedCreditCardPaymentMethod: FunctionComponent<
                         createCyberSourcePaymentStrategy,
                         createCyberSourceV2PaymentStrategy,
                         createSagePayPaymentStrategy,
-                        createCBAMPGSPaymentStrategy,
                     ],
                     creditCard: getHostedFormOptions && {
                         form: await getHostedFormOptions(selectedInstrument),
--- a/packages/core/src/app/payment/paymentMethod/HostedCreditCardPaymentMethod.tsx
+++ b/packages/core/src/app/payment/paymentMethod/HostedCreditCardPaymentMethod.tsx
@@ -1,3 +1,4 @@
+import { createCBAMPGSPaymentStrategy } from '@bigcommerce/checkout-sdk/integrations/cba-mpgs';
 import { createCreditCardPaymentStrategy } from '@bigcommerce/checkout-sdk/integrations/credit-card';
 import {
     createCyberSourcePaymentStrategy,
@@ -6,6 +7,9 @@ import {
 import { createSagePayPaymentStrategy } from '@bigcommerce/checkout-sdk/integrations/sagepay';
 import React, { type FunctionComponent, useCallback } from 'react';
 
+import { useCheckout } from '@bigcommerce/checkout/contexts';
+
+import { isExperimentEnabled } from '../../common/utility';
 import {
     withHostedCreditCardFieldset,
     type WithInjectedHostedCreditCardFieldsetProps,
@@ -34,6 +38,13 @@ const HostedCreditCardPaymentMethod: FunctionComponent<
     initializePayment,
     ...rest
 }) => {
+    const { selectedState: config } = useCheckout(({ data }) => data.getConfig());
+    const isCBAMPGSResolverEnabled = isExperimentEnabled(
+        config?.checkoutSettings,
+        'PI-4748.cba_resolver_configuration',
+        false,
+    );
+
     const initializeHostedCreditCardPayment: CreditCardPaymentMethodProps['initializePayment'] =
         useCallback(
             async (options, selectedInstrument) => {
@@ -45,13 +56,14 @@ const HostedCreditCardPaymentMethod: FunctionComponent<
                         createCyberSourcePaymentStrategy,
                         createCyberSourceV2PaymentStrategy,
                         createSagePayPaymentStrategy,
+                        ...(!isCBAMPGSResolverEnabled ? [createCBAMPGSPaymentStrategy] : []),
                     ],
                     creditCard: getHostedFormOptions && {
                         form: await getHostedFormOptions(selectedInstrument),
                     },
                 });
             },
-            [getHostedFormOptions, initializePayment],
+            [getHostedFormOptions, initializePayment, isCBAMPGSResolverEnabled],
         );
 
     return (
--- a/packages/core/src/app/payment/resolvePaymentMethod.ts
+++ b/packages/core/src/app/payment/resolvePaymentMethod.ts
@@ -1,3 +1,4 @@
+import type { CheckoutSettings } from '@bigcommerce/checkout-sdk';
 import { type ComponentType } from 'react';
 
 import {
@@ -5,15 +6,36 @@ import {
     type PaymentMethodResolveId,
 } from '@bigcommerce/checkout/payment-integration-api';
 
+import { isExperimentEnabled } from '../common/utility';
 import { resolveLazyComponent } from '../common/resolver';
 import * as lazyPaymentMethods from '../generated/paymentIntegrations';
 
 export default function resolvePaymentMethod(
     query: PaymentMethodResolveId,
+    checkoutSettings?: CheckoutSettings,
 ): ComponentType<PaymentMethodProps> | undefined {
     const { ComponentRegistry, ...allExports } = lazyPaymentMethods;
+    const filteredMethods = Object.fromEntries(
+        Object.entries(ComponentRegistry).map(([key, value]) => {
+            const methods = value.filter((resolveId) => {
+                const { experiment } = resolveId as PaymentMethodResolveId;
+
+                if (!experiment) {
+                    return true;
+                }
+
+                // should be normalized after config
+                const normalizedKey = experiment.replace('_', '.');
+
+                return isExperimentEnabled(checkoutSettings, normalizedKey, false);
+            });
+
+            return [key, methods];
+        }),
+    );
+
     const components = Object.fromEntries(
-        Object.keys(ComponentRegistry).map((key) => [
+        Object.keys(filteredMethods).map((key) => [
             key,
             allExports[key as keyof typeof allExports],
         ]),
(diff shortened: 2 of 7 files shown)


animesh1987 (dismissed):

bc-ania (dismissed):

packages/core/src/app/payment/paymentMethod/HostedCreditCardPaymentMethod.tsx
```diff
@@ -35,6 +38,13 @@ const HostedCreditCardPaymentMethod: FunctionComponent<
     initializePayment,
     ...rest
 }) => {
+    const { selectedState: config } = useCheckout(({ data }) => data.getConfig());
+    const isCBAMPGSResolverEnabled = isExperimentEnabled(
```
animesh1987, line 42: 👍

packages/core/src/app/payment/resolvePaymentMethod.ts
```diff
@@ -1,19 +1,41 @@
...
+import { isExperimentEnabled } from '../common/utility';
 import { resolveLazyComponent } from '../common/resolver';
 import * as lazyPaymentMethods from '../generated/paymentIntegrations';
 
 export default function resolvePaymentMethod(
     query: PaymentMethodResolveId,
+    checkoutSettings?: CheckoutSettings,
 ): ComponentType<PaymentMethodProps> | undefined {
     const { ComponentRegistry, ...allExports } = lazyPaymentMethods;
+    const filteredMethods = Object.fromEntries(
```
animesh1987, line 18: Why are we adding this?

henrikh-kovalenko: Hi @animesh1987  we want to experiments for our PaymentMethods 
However we can't somehow change this 
image
that's why one place where we can check this experiments in this file, and we filter our payment methods if experiment is false then this method should not be in result arr

henrikh-kovalenko: for behavior of this PR we set experiment name in config then in file resolvePaymentMethod we check is this experiment true or false if it is true pr if we don't have experiment then this method will be in result arr if not than we will not have this mthod in arr

packages/core/src/app/payment/paymentMethod/HostedCreditCardPaymentMethod.tsx
```diff
@@ -35,6 +38,13 @@ const HostedCreditCardPaymentMethod: FunctionComponent<
...
+        config?.checkoutSettings,
+        'PI-4748.cba_resolver_configuration',
+        false,
```
bc-peng, line 45: The purpose of this fallback value is for the time when `PI-4748.cba_resolver_configuration` is `undefined` in the checkout settings.
If we have surfaced it, we don't need to explicitly set the value here.

packages/core/src/app/payment/resolvePaymentMethod.ts
bc-peng: Attaching an experiment name to a payment method config requires some explanations on the purpose of this experiment.

Could we use the experiment to control payment method identifiers instead of making it part of the payment method resolution logic?

packages/core/src/app/payment/paymentMethod/HostedCreditCardPaymentMethod.tsx
henrikh-kovalenko: But we have default value true and we need false
image

packages/core/src/app/payment/resolvePaymentMethod.ts
henrikh-kovalenko: This experiment is specifically needed to safely migrate the CBA MPGS and other payment providers that are using hosted fields from the legacy core flow to the new resolver-based package architecture.
Why it needs to be on the resolver level, not on the identifier level:
The resolver is the exact point where the routing decision is made — it determines whether a payment method renders via the new package component or falls back to the legacy V1 core component. Controlling the identifier itself wouldn't give us that routing control, since the same cba_mpgs ID needs to exist in both paths during the transition.
Why we're being cautious:
We previously had an incident with the hosted fields migration process, so we want to roll this out gradually and have a clean rollback path. The experiment flag gives us that safety net.
What happens after the migration is complete:
Once we've validated the rollout, we'll remove both the experiment flag and all the conditional logic around it — it's intentionally duplicated in the code to make that cleanup straightforward (each side clearly shows what gets removed).
Future migrations:
We'll need similar experiments for migrating other payment providers through the same pattern. The experiment field on PaymentMethodResolveId is designed to be generic for exactly that purpose — any future provider migration can use the same mechanism without changes to the resolver infrastructure itself.

packages/core/src/app/payment/paymentMethod/HostedCreditCardPaymentMethod.tsx
bc-peng: Could you please explain when the experiment will not appear in the `features[]`?

packages/core/src/app/payment/resolvePaymentMethod.ts
bc-peng: My question is narrower: it's about where the gate lives, not whether it exists. 

Agreed on the rollout safety, the experiment, gradual rollout, and clean revert path. All make sense given the prior incident, no objection there.

The point that the routing decision must happen at runtime in the resolver is right (the registry is generated at build time). But that doesn't require an `experiment` to be a field on `PaymentMethodResolveId`.

If we do want a generic per-package migration-gate mechanism long term, I think that's worth its own small design rather than an experiment field.

packages/core/src/app/payment/paymentMethod/HostedCreditCardPaymentMethod.tsx
henrikh-kovalenko: Before the experiment is surfaced/registered in LaunchDarkly (or whatever feature flag service is used), the key won't exist in features[] at all — features['PI-4748.cba_resolver_configuration'] will be undefined.

In that case ?? fallbackValue kicks in. Since the default fallback in isExperimentEnabled is true, without explicitly passing false here, cba_mpgs would resolve through the new flow on day one — before we've intentionally enabled it for anyone. That's exactly what we want to avoid.

packages/core/src/app/payment/resolvePaymentMethod.ts
henrikh-kovalenko: this is not long term, we need this for migration maybe 3 providers. 
 Maybe you have some suggestions how we can resolve it ?

packages/core/src/app/payment/paymentMethod/HostedCreditCardPaymentMethod.tsx
bc-peng: You are right, this is intended use case when we introduced this fallback value.

The experiment appears to be added on Jun 23, 2026 at https://app.launchdarkly.com/projects/default/flags/PI-4748.cba_resolver_configuration/targeting?env=production&selected-env=production.

Is anything blocking us from showing it to Checkout FE now?

packages/core/src/app/payment/resolvePaymentMethod.ts
bc-peng: Thanks for the context. That changes the perspective. 

> this is not long term, we need this for migration maybe 3 providers.

If possible, could you please add this context as comments? My concern is we will soon forget about it.

packages/core/src/app/payment/paymentMethod/HostedCreditCardPaymentMethod.tsx
henrikh-kovalenko: I turned on this flag if you about that. Or I miss understood

packages/core/src/app/payment/resolvePaymentMethod.ts
henrikh-kovalenko: I created task for remove this functional after migration will be completed PI-5464

animesh1987 (approved):

bc-peng (approved):
Thank you for your explanations and creating PI-5464. 🙏🏻

bc-launchbay:
success [20260706104703](https://launchbay.bigcommerce.net/projects/535/releases/290305) by Launchbay

bc-launchbay:
success [20260706104703](https://launchbay.bigcommerce.net/projects/535/releases/290305) by Henrikh Kovalenko

bc-launchbay:
success [20260706104703](https://launchbay.bigcommerce.net/projects/535/releases/290305) by Launchbay

PR #3081, opened 2026-06-11
feat(checkout): CHECKOUT-9819 implement post order request for b2b user

richapatel1510:
## What/Why?

This PR implements the post-order B2B metadata request for the normal (non-invoice) B2B checkout flow as part of CHECKOUT-9819.

Previously, on order submission only the invoice comment was persisted via `submitB2BMetadata`. This change consolidates **all** B2B data captured during checkout — PO number, reference number, invoice comment, order extra fields, and address extra fields/IDs — into a single metadata payload that is sent after a successful order, then clears the associated sessionStorage.

**Technical details:**

- **New `b2bMetadata.ts` helper** (`packages/core/src/app/payment/b2bMetadata.ts`) centralizes payload assembly and storage cleanup:
  - `buildB2BMetadataOptions()` builds the full `PersistB2BMetadataOptions` payload (invoice flag, invoice comment, PO number, reference number, order extra fields, address extra info).
  - `buildExtraFields()` maps stored values to the API shape, sending the field **label** rather than the field name/id (values are stored keyed by name, but the API expects the label).
  - `buildAddressExtraInfo()` collects billing/shipping address extra fields plus the selected billing/shipping address IDs.
  - `clearB2BMetadataStorage()` clears every B2B sessionStorage entry after a successful persist.
- **Payment** (`packages/core/src/app/payment/Payment.tsx`): `persistB2BMetadataIfNeeded` now builds and submits the consolidated payload and then clears storage. Address extra fields are wired into props via `getAddressExtraFields()`, gated on the `hasAddressExtraFields` capability.
- **Bug fix** (`packages/utility/src/storage/poNumberStorage.ts`): corrects the storage key casing from `PoNumber` → `poNumber` so the PO number is read back correctly.

## Rollout/Rollback

This change is gated behind existing B2B capabilities (`hasAddressExtraFields`, `hasCompanyAddressBook`) and the `persistB2BMetadata` flag — when these are off, behavior is unchanged, so there is no impact to non-B2B storefronts.

Rollback is code-only (revert this PR); there are no database migrations or data backfills involved. Reverting restores the prior behavior of persisting only the invoice comment.

## Testing

- CI checks
- Manual testing 

https://github.com/user-attachments/assets/6b334760-3618-4c70-89b7-2a0121d93b39

### DIFF
--- a/packages/core/src/app/address/B2BExtraFieldsSessionStorage.ts
+++ b/packages/core/src/app/address/B2BExtraFieldsSessionStorage.ts
@@ -3,6 +3,8 @@ export class B2BExtraFieldsSessionStorage {
     static readonly SHIPPING_KEY = 'b2bAddressExtraFields_shipping';
     static readonly CONSIGNMENT_KEY_PREFIX = 'b2bAddressExtraFields_consignment_';
     static readonly ORDER_KEY = 'b2bOrderExtraFields';
+    static readonly BILLING_ADDRESS_ID_KEY = 'b2bBillingAddressId';
+    static readonly SHIPPING_ADDRESS_ID_KEY = 'b2bShippingAddressId';
 
     static getConsignmentKey(consignmentId: string): string {
         return `${this.CONSIGNMENT_KEY_PREFIX}${consignmentId}`;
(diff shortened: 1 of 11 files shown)


richapatel1510:
Covered by https://github.com/bigcommerce/checkout-js/pull/3087

PR #3085, opened 2026-06-12
fix(checkout): CHECKOUT-9891 Remove B2B sessionStorage on order completion

bc-mackxu:
## Summary

- Adds `B2BExtraFieldsSessionStorage.removeAll()` to clear all B2B address extra field keys (billing, shipping, order, all consignment keys) from sessionStorage
- After a successful order submission, clears all three B2B sessionStorage stores: `B2BExtraFieldsSessionStorage`, `AdditionalPaymentFieldSessionStorage`, and `InvoicePaymentCommentSessionStorage`
- Cleanup happens after `persistB2BMetadataIfNeeded()` (so the data is submitted first) and before navigation to order confirmation

## Why

B2B extra fields, additional payment field, and invoice payment comment were saved to sessionStorage during checkout but never removed. This meant stale values from a previous order would persist into the next checkout session in the same browser tab.

## Notes for reviewer

- Cleanup only runs on successful order submission — if the order fails the user stays on the payment page and their entered values are preserved
- `removeAll()` iterates sessionStorage keys rather than hardcoding individual removes, so it handles an arbitrary number of consignment keys

### DIFF
--- a/packages/core/src/app/address/B2BExtraAddressFieldsSessionStorage.test.ts
+++ b/packages/core/src/app/address/B2BExtraAddressFieldsSessionStorage.test.ts
@@ -38,4 +38,39 @@ describe('B2BExtraFieldsSessionStorage', () => {
             ).toBeUndefined();
         });
     });
+
+    describe('removeAll', () => {
+        it('removes billing, shipping, order, and consignment keys', () => {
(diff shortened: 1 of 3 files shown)


bc-mackxu:
Closing — duplicate, already addressed in another PR.

PR #3092, opened 2026-06-15
feat(checkout): CHECKOUT-9891 refactor b2b session storage

richapatel1510:
## What/Why?

Refactor session storage for b2b use cases. 
- Consolidate PO number, additional payment field, invoice comment storage into 1 class. 
- Add a clearAll and getAll helper in `B2BSessionStorage` to clear and get all the address extra fields from storage at once.

## Rollout/Rollback

Revert this PR.

## Testing

- CI checks

---

> [!NOTE]
> **Medium Risk**
> Wide checkout refactor around B2B metadata and address-book IDs; new session key names and encoding may drop in-progress shopper data until they re-enter fields, though persist/clear behavior is covered by updated tests.
> 
> **Overview**
> **Consolidates all B2B checkout session persistence** into `B2BSessionStorage` in `@bigcommerce/checkout/utility`, replacing `poNumberStorage`, payment-field storage classes, and core `B2BExtraFieldsSessionStorage`.
> 
> Checkout **billing, shipping, payment, and address mappers** now read/write PO numbers, invoice comments, reference fields, order/address extra fields, and company address book IDs through shared keys and helpers (`get`, `getValue`, `getAll`, `clearAll`, `copyShippingToBilling`, `reassignConsignmentKey`). **`buildB2BMetadataOptions`** pulls everything via `getAll()` instead of multiple storage classes; **`clearB2BMetadataStorage`** calls **`clearAll()`**.
> 
> **Storage format changes:** values use **base64-encoded JSON** under **`b2b_*` keys** (replacing legacy key names). **Cleared text fields** are stored as **empty strings** rather than removing session keys. Address book IDs are encoded like other values instead of plain string storage.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 805b0f2667e6765756d6f425586c95ab5f6f2dbf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


richapatel1510:
cursor review

packages/utility/src/storage/B2BPaymentFieldsSessionStorage.ts
```diff
@@ -0,0 +1,29 @@
+export class B2BPaymentFieldsSessionStorage {
+    /* eslint-disable @typescript-eslint/naming-convention -- UPPER_CASE key constants, mirroring B2BExtraFieldsSessionStorage */
+    static readonly PO_NUMBER_KEY = 'poNumber';
+    static readonly ADDITIONAL_PAYMENT_FIELD_KEY = 'additionalPaymentField';
+    static readonly INVOICE_COMMENT_KEY = 'invoicePaymentComment';
```
bc-maxy: These don't have a b2b prefix (compared to the ones is B2BExtraFieldsSessionStorage.ts), is this because they can be used outside of b2b?

packages/utility/src/storage/B2BPaymentFieldsSessionStorage.ts
```diff
@@ -0,0 +1,29 @@
...
+            sessionStorage.removeItem(key);
+        }
+    }
+
+    static remove(key: string): void {
+        sessionStorage.removeItem(key);
+    }
+
+    static clearAll(): void {
+        [this.PO_NUMBER_KEY, this.ADDITIONAL_PAYMENT_FIELD_KEY, this.INVOICE_COMMENT_KEY].forEach(
```
bc-maxy: Can we make this array dynamic so that in case we add a new payment field, we don't need to update this method?

Maybe 
```
static readonly KEYS = {
        PO_NUMBER: 'poNumber',
        ADDITIONAL_PAYMENT_FIELD: 'additionalPaymentField',
        INVOICE_COMMENT: 'invoicePaymentComment',
    } as const;
```
instead of declaring them individually? What do you think?

packages/utility/src/storage/B2BPaymentFieldsSessionStorage.ts
richapatel1510, line 1: Does it make sense to have a single B2BSessionStorage class and save all the session storage from there? The only difference this `B2BPaymentFieldsSessionStorage.ts` and `B2BExtraFieldsSessionStorage.ts` is the `atob` parsing logic we use for extra fields.

bc-maxy (commented):
This is great 👏

packages/utility/src/storage/B2BPaymentFieldsSessionStorage.ts
richapatel1510: I was thinking the same, but then I thought when using those keys to add values to session storage it will be a huge line just to save some work while clearing.
For example while using it will be `B2BPaymentFieldsSessionStorage.get(B2BPaymentFieldsSessionStorage.KEYS.PO_NUMBER)`

packages/utility/src/storage/B2BPaymentFieldsSessionStorage.ts
richapatel1510: No, they cannot be outside of b2b, just left it as is since when we started the project we didn't want any b2b branching in checkout-js, but now that we already have b2b prefix for extra fields it makes sense to add it here as well or remove from the other keys. Something we can decide.

packages/utility/src/storage/B2BPaymentFieldsSessionStorage.ts
bc-peng: Yes, besides B2B, we only use session storage to save newsletter subscription checkbox, so it makes sense to only use one class to make them all.

richapatel1510: So we use the same parsing logic for payment fields as well? That should be okay, right?

bc-maxy: Would be great to have one class, one prefix and one clear method for all b2b storage keys 😇

packages/utility/src/storage/B2BPaymentFieldsSessionStorage.ts
richapatel1510: @bc-peng WDYT?

packages/utility/src/storage/B2BPaymentFieldsSessionStorage.ts
richapatel1510: Also there's one downside if we merge this, we might need to move the entire session storage class to utility package, jus to that chequePaymentIntegration can read from it.

packages/utility/src/storage/B2BPaymentFieldsSessionStorage.ts
bc-maxy: I think it's low impact considering we are already using B2BPaymentFieldsSessionStorage.get(B2BPaymentFieldsSessionStorage.PO_NUMBER) but definitely not a blocker :D

richapatel1510: If we add this https://github.com/bigcommerce/checkout-js/pull/3092#discussion_r3411298370 then we don't need this map, since we will map with the prefix. 👍

packages/utility/src/storage/B2BPaymentFieldsSessionStorage.ts
bc-peng: That sounds OK.

packages/utility/src/storage/B2BPaymentFieldsSessionStorage.ts
bc-peng: Since the file is named `B2BPaymentFieldsSessionStorage.ts` already, the prefix makes sense.

richapatel1510:
cursor review

bc-peng (approved):

animesh1987 (approved):

bc-launchbay:
success [20260616012341](https://launchbay.bigcommerce.net/projects/535/releases/288138) by Launchbay

bc-launchbay:
success [20260616012341](https://launchbay.bigcommerce.net/projects/535/releases/288138) by Launchbay

bc-launchbay:
success [20260616012341](https://launchbay.bigcommerce.net/projects/535/releases/288138) by Launchbay

PR #3115, opened 2026-06-22
fix(checkout): CHECKOUT-10138 Make sure checkout settings are defined

bc-maxy:
## What/Why?

I did a quick investigation on this sentry issue: https://bigcommerce.sentry.io/issues/7547058371/?project=1542560
I could not find an obvious root cause - useCheckout work seems unrelated based on Claude evaluation.

In any case, I'm adding a check for `checkoutSetting` in this PR to make sure we don't evaluate `checkoutSettings` when they are undefined.

## Rollout/Rollback

Revert the PR.

## Testing

CI.

---

> [!NOTE]
> **Low Risk**
> Single defensive guard in theme selection; behavior unchanged when checkoutSettings is present.
> 
> **Overview**
> **ThemeProvider** no longer reads theme experiment/setting flags unless **`checkoutSettings`** is present on checkout config.
> 
> The guard changes from **`if (config)`** to **`if (config?.checkoutSettings)`**, so a config object without settings skips the nested **`features`** / **`checkoutUserExperienceSettings`** access and keeps **`themeV2`** at **`false`** instead of throwing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1d48c6cb9b86e435f93568485b4d314e69cfb015. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/contexts/src/theme/ThemeProvider.tsx
+++ b/packages/contexts/src/theme/ThemeProvider.tsx
@@ -13,7 +13,7 @@ export const ThemeProvider = ({ children }: ThemeProviderProps) => {
 
     let themeV2 = false;
 
-    if (config) {
+    if (config?.checkoutSettings) {
         const newThemeExperimentEnabled = Boolean(
             config.checkoutSettings.features['CHECKOUT-7962.update_font_style_on_checkout_page'] ??
                 true,


packages/contexts/src/theme/ThemeProvider.tsx
```diff
@@ -13,7 +13,7 @@ export const ThemeProvider = ({ children }: ThemeProviderProps) => {
 
     let themeV2 = false;
 
-    if (config) {
+    if (config?.checkoutSettings) {
```
bc-peng, line 16: My initial read is that this is not a FE issue.
When `config` is defined, we should always have `checkoutSettings` and `checkoutSettings.features` for OOPC.

bc-maxy: Good point! This is the case based on our types, but looks like it is undefined in some cases in practice

animesh1987: If there is no config, should we just return early?

bc-maxy: In general I agree but in this case I think this looks cleaner than "duplicated" <ThemeContext.Provider ... /> (for early return and then for computed themeV2), or? 😄

bc-maxy:
The sentry error btw is `undefined is not an object (evaluating 'o.checkoutSettings.features')`, which means that it's the `checkoutSettings` that are undefined.
We don't need an additional type check for `features` at this point then.

animesh1987 (approved):

packages/contexts/src/theme/ThemeProvider.tsx
bc-peng: Can we use [this component](https://github.com/bigcommerce/checkout-js/blob/e85a3c096404d0dc95b1f6f1d01ce21be54ab0ad/packages/payment-integration-api/src/CaptureMessageComponent.tsx#L9) to record the case where we have `config`, but not `checkoutSettings`?

You may need to place it on L28.

bc-maxy: @bc-peng It creates a circular dependency if we use it. I can refactor it or move to contexts package if you think it's worth it?

bc-peng (approved):

packages/contexts/src/theme/ThemeProvider.tsx
bc-peng: I see. The circular dependency is formed as the component is reading checkout context.

For a serious Sentry issue, we can move the component to the `contexts` package.

Since this Sentry issue only caught 17 events over the last 30 days, we should be able to safely ignore it.

Let's merge the existing code.

bc-launchbay:
success [20260629040323](https://launchbay.bigcommerce.net/projects/535/releases/289553) by Launchbay

bc-launchbay:
success [20260629040323](https://launchbay.bigcommerce.net/projects/535/releases/289553) by Launchbay

bc-launchbay:
success [20260629040323](https://launchbay.bigcommerce.net/projects/535/releases/289553) by Launchbay

PR #3117, opened 2026-06-22
feat(checkout): BACK-679 Display backorder details for mobile modals

animesh1987:
## What/Why?
Display backorder details for mobile modals

## Rollout/Rollback
- revert this PR

## Testing
- CI
- Screencast

https://github.com/user-attachments/assets/26dd974b-ff30-4675-8b32-9c908c6d12d7

---

> [!NOTE]
> **Medium Risk**
> Checkout order-summary UX changes affect mobile modals, bundle experiment gating, and shared module-scoped backorder toggle state across breakpoints; behavior is heavily tested but still customer-facing cart UI.
> 
> **Overview**
> Enables the **Backorder details** switch in the mobile cart modal (when inventory settings allow) and wires it so shoppers can expand per-line backorder quantity and messages, including on **bundled** children when `BACK-425.update_bundle_item_ux` is on.
> 
> **`OrderSummaryItems`** now treats the heading as optional item count plus an optional toggle: the modal passes `isMobileCartModal` and hides the count while still showing the switch when appropriate. On mobile, the toggle is shown only if the bundle pick-list experiment is enabled (so a backorder on a hidden bundle child does not surface a useless switch). Toggle state is stored in a **module-scoped** flag so it survives responsive remounts; line items only expand when the toggle is actually rendered, so a desktop “on” selection does not leak expanded UI on a gated mobile modal.
> 
> **`OrderSummaryDrawer`** forwards **raw** line items into the modal instead of pre-stripped bundles; **`OrderSummaryModal`** still uses `removeBundledItems` for the header item count only. Styling adds `cart-section-heading-container--switch-only` to right-align the switch when there is no heading count.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b2ad74ac9076555b705fb14d6ca48f0d055b185c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/app/order/OrderSummaryItem.test.tsx
+++ b/packages/core/src/app/order/OrderSummaryItem.test.tsx
@@ -1,6 +1,11 @@
-import { type CurrencyService } from '@bigcommerce/checkout-sdk';
+import {
+    type CheckoutService,
+    createCheckoutService,
+    type CurrencyService,
+} from '@bigcommerce/checkout-sdk';
 import React from 'react';
 
+import { CheckoutProvider, LocaleContext } from '@bigcommerce/checkout/contexts';
 import { createLocaleContext } from '@bigcommerce/checkout/locale';
 import { render, screen } from '@bigcommerce/checkout/test-utils';
 
@@ -210,6 +215,69 @@ describe('OrderSummaryItem', () => {
             );
 
             expect(screen.queryByRole('list')).not.toBeInTheDocument();
+        });
+    });
+
+    describe('when bundled items are backordered and details are expanded', () => {
+        let checkoutService: CheckoutService;
+
+        const renderWithConfig = (shouldExpandBackorderDetails: boolean) =>
+            render(
+                <CheckoutProvider checkoutService={checkoutService}>
+                    <LocaleContext.Provider value={localeContext}>
+                        <OrderSummaryItem
+                            orderItem={{
+                                amount: 10,
+                                id: 'foo',
+                                name: 'Product',
+                                quantity: 1,
+                                bundledItems: [
+                                    {
+                                        id: 'b1',
+                                        name: 'Bundled Hat',
+                                        bundleLabel: 'Accessories',
+                                        quantityBackordered: 2,
+                                        quantityOnHand: 3,
+                                        backorderMessage: 'Ships in 5 days',
+                                    },
+                                ],
+                            }}
+                            shouldExpandBackorderDetails={shouldExpandBackorderDetails}
+                        />
+                    </LocaleContext.Provider>
+                </CheckoutProvider>,
+            );
+
+        beforeEach(() => {
+            checkoutService = createCheckoutService();
+
+            jest.spyOn(checkoutService.getState().data, 'getConfig').mockReturnValue({
+                ...getStoreConfig(),
+                inventorySettings: {
+                    showQuantityOnBackorder: true,
+                    showBackorderMessage: true,
+                    showQuantityOnHand: false,
+                    showBackorderAvailabilityPrompt: false,
+                    backorderAvailabilityPrompt: null,
+                    shouldDisplayBackorderMessagesOnStorefront: true,
+                },
+            });
+        });
+
+        it('renders the backorder quantity and message for the bundled child item', () => {
+            renderWithConfig(true);
+
+            expect(screen.getByTestId('cart-item-backorder-qty')).toBeInTheDocument();
+            expect(screen.getByTestId('cart-item-backorder-message')).toHaveTextContent(
+                'Ships in 5 days',
+            );
+        });
+
+        it('does not render the bundled child backorder details when collapsed', () => {
+            renderWithConfig(false);
+
+            expect(screen.queryByTestId('cart-item-backorder-qty')).not.toBeInTheDocument();
+            expect(screen.queryByTestId('cart-item-backorder-message')).not.toBeInTheDocument();
         });
     });
 });
--- a/packages/core/src/app/order/OrderSummaryItem.test.tsx
+++ b/packages/core/src/app/order/OrderSummaryItem.test.tsx
@@ -259,6 +259,8 @@ describe('OrderSummaryItem', () => {
                     showQuantityOnHand: false,
                     showBackorderAvailabilityPrompt: false,
                     backorderAvailabilityPrompt: null,
+                    showDefaultShippingExpectationPrompt: false,
+                    defaultShippingExpectationPrompt: null,
                     shouldDisplayBackorderMessagesOnStorefront: true,
                 },
             });
--- a/packages/core/src/app/order/OrderSummaryItems.test.tsx
+++ b/packages/core/src/app/order/OrderSummaryItems.test.tsx
@@ -20,7 +20,7 @@ import { getStoreConfig } from '../config/config.mock';
 
 import OrderSummaryItems, {
     type OrderSummaryItemsProps,
-    resetBackorderDetailsExpanded,
+    setBackorderDetailsExpanded,
 } from './OrderSummaryItems';
 
 describe('OrderSummaryItems', () => {
@@ -84,7 +84,7 @@ describe('OrderSummaryItems', () => {
     afterEach(() => {
         // The toggle selection is held in a module-scoped value so it can survive the
         // responsive remount; reset it between tests to avoid leaking state.
-        resetBackorderDetailsExpanded();
+        setBackorderDetailsExpanded(false);
     });
 
     describe('backorder quantity text', () => {
--- a/packages/core/src/app/order/OrderSummaryItems.test.tsx
+++ b/packages/core/src/app/order/OrderSummaryItems.test.tsx
@@ -37,6 +37,29 @@ describe('OrderSummaryItems', () => {
         );
     };
 
+    const enableBundleExperiment = () => {
+        const config = getStoreConfig();
+
+        jest.spyOn(checkoutService.getState().data, 'getConfig').mockReturnValue({
+            ...config,
+            checkoutSettings: {
+                ...config.checkoutSettings,
+                features: {
+                    ...config.checkoutSettings.features,
+                    'BACK-425.update_bundle_item_ux': true,
+                },
+            },
+            inventorySettings: {
+                showQuantityOnBackorder: true,
+                showBackorderMessage: true,
+                showQuantityOnHand: false,
+                showBackorderAvailabilityPrompt: false,
+                backorderAvailabilityPrompt: null,
+                shouldDisplayBackorderMessagesOnStorefront: true,
+            },
+        });
+    };
+
     beforeEach(() => {
         checkoutService = createCheckoutService();
         localeContext = createLocaleContext(getStoreConfig());
@@ -162,7 +185,7 @@ describe('OrderSummaryItems', () => {
         });
     });
 
-    describe('backorder details toggle parity and persistence', () => {
+    describe('backorder details toggle gating (mobile vs desktop) and persistence', () => {
         const backorderItems = {
             customItems: [],
             physicalItems: [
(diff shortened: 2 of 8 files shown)


agchan12:
### Code review

Found 1 issue:

1. **Desktop backorder toggle can appear as a no-op when only hidden bundle children are backordered (experiment off).** `getBackorderCount(items)` at line 216 counts raw items including bundle children. On desktop (`displayLineItemsCount=true`), `showBackorderToggle` evaluates to `true` even when only a bundle child is backordered. But `nonBundledItems` (line 237) strips those children from the rendered `ProductList`, so toggling expands to show nothing. The mobile path is correctly gated by `!isMobileCartModal || pickListExperimentEnabled` (line 233), but the desktop path has no equivalent guard. The code comment at lines 225-227 states the intent is to "stop it appearing when only hidden bundle children are backordered" — this invariant is only enforced on mobile. Consider computing `backorderCount` from `nonBundledItems` when the experiment is off, or applying the experiment gate to both paths.

https://github.com/bigcommerce/checkout-js/blob/2af72ba286858765933d61c556337faa7ae5beb3/packages/core/src/app/order/OrderSummaryItems.tsx#L215-L237

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

bc-shawnwang (dismissed):

packages/core/src/app/order/OrderSummaryItems.tsx
```diff
@@ -30,6 +31,24 @@ import mapFromPhysical from './mapFromPhysical';
...
+
+const getBackorderDetailsExpanded = (): boolean => backorderDetailsExpanded;
+
+const setBackorderDetailsExpanded = (value: boolean): void => {
+    backorderDetailsExpanded = value;
+};
+
+export const resetBackorderDetailsExpanded = (): void => {
+    backorderDetailsExpanded = false;
+};
```
richapatel1510: Instead of this can we reuse `setBackorderDetailsExpanded(false)` where we call `resetBackorderDetailsExpanded()`

packages/core/src/app/order/OrderSummaryItems.tsx
```diff
@@ -186,13 +200,38 @@ const OrderSummaryItems = ({
...
 
+    // The item-count heading is hidden only in the mobile cart modal. There, bundle children are
+    // not rendered while the bundle experiment is off, so gate the backorder toggle behind the
+    // experiment to stop it appearing when only hidden bundle children are backordered.
+    const isMobileCartModal = !displayLineItemsCount;
+
+    const showBackorderToggle =
+        shouldDisplayBackorderDetails &&
+        backorderCount > 0 &&
+        (!isMobileCartModal || pickListExperimentEnabled);
```
richapatel1510, line 224: Can we use `displayLineItemsCount` directly here?

richapatel1510: Also `showBackorderToggle` is going to be used in `SummaryHeading` only if `displayLineItemsCount` is true, so we probably don't need this?

packages/core/src/app/order/OrderSummaryItems.tsx
```diff
@@ -186,13 +200,38 @@ const OrderSummaryItems = ({
...
+
+            return next;
+        });
+    }, []);
+
+    const backorderCount = getBackorderCount(items);
+    const shouldDisplayBackorderDetails =
+        !!config?.inventorySettings?.shouldDisplayBackorderMessagesOnStorefront &&
+        (!!config?.inventorySettings?.showQuantityOnBackorder ||
+            !!config?.inventorySettings?.showBackorderMessage);
```
richapatel1510, line 213: This was already in the child component that needed this value, any reason we moved this up?

animesh1987:
@agchan12 that is by design, since I am rolling out the experiment. So I don't want the branching for backorder display to be duplicated in mobile as well. Once I am done rolling out maybe next week, I will remove the experiment and it will be good.

packages/core/src/app/order/OrderSummaryItems.tsx
```diff
@@ -186,13 +200,38 @@ const OrderSummaryItems = ({
     items,
 }: OrderSummaryItemsProps): ReactElement => {
     const [isExpanded, setIsExpanded] = useState(false);
-    const [showBackorderDetails, setShowBackorderDetails] = useState(false);
+    const [showBackorderDetails, setShowBackorderDetails] = useState(getBackorderDetailsExpanded);
```
richapatel1510, line 196: Do we need this state now? We are setting 2 values when we change it?

agchan12 (dismissed):

packages/core/src/app/order/OrderSummaryItem.test.tsx
```diff
@@ -212,4 +217,67 @@ describe('OrderSummaryItem', () => {
...
+
+            jest.spyOn(checkoutService.getState().data, 'getConfig').mockReturnValue({
+                ...getStoreConfig(),
+                inventorySettings: {
+                    showQuantityOnBackorder: true,
+                    showBackorderMessage: true,
+                    showQuantityOnHand: false,
+                    showBackorderAvailabilityPrompt: false,
+                    backorderAvailabilityPrompt: null,
+                    shouldDisplayBackorderMessagesOnStorefront: true,
```
bc-maxy, line 264: I think you also need these based on our interface?
```
showDefaultShippingExpectationPrompt: false,
defaultShippingExpectationPrompt: null,
```

packages/core/src/app/order/OrderSummaryItems.test.tsx
```diff
@@ -35,6 +37,29 @@ describe('OrderSummaryItems', () => {
...
+                    'BACK-425.update_bundle_item_ux': true,
+                },
+            },
+            inventorySettings: {
+                showQuantityOnBackorder: true,
+                showBackorderMessage: true,
+                showQuantityOnHand: false,
+                showBackorderAvailabilityPrompt: false,
+                backorderAvailabilityPrompt: null,
+                shouldDisplayBackorderMessagesOnStorefront: true,
```
bc-maxy, line 60: and similarly here unless you are returning partial settings on purpose

packages/core/src/app/order/OrderSummaryItems.tsx
```diff
@@ -30,6 +31,24 @@ import mapFromPhysical from './mapFromPhysical';
 import OrderSummaryItem from './OrderSummaryItem';
 import { removeAndBundleItemsTogether, removeBundledItems } from './removeBundledItems';
 
+// The backorder-details toggle must survive the responsive remount (desktop/mobile
+// breakpoint swap unmounts and remounts the order summary subtree) and stay consistent
+// between the desktop and mobile instances. Because <MobileView> renders mutually
+// exclusive subtrees, only one OrderSummaryItems is mounted at a time, so a module-scoped
+// value written on toggle and read on mount is sufficient. It re-initializes to OFF on a
+// full page reload, which matches the requirement of persisting across resize only.
```
bc-maxy: nit: A bit too verbose, maybe something like this 😄 :
// Module-scoped to survive the responsive remount. Safe as MobileView mounts only one instance at a time.

packages/core/src/app/order/OrderSummaryItems.tsx
```diff
@@ -186,13 +200,38 @@ const OrderSummaryItems = ({
...
+        (!!config?.inventorySettings?.showQuantityOnBackorder ||
+            !!config?.inventorySettings?.showBackorderMessage);
     const pickListExperimentEnabled = config
         ? isExperimentEnabled(config.checkoutSettings, 'BACK-425.update_bundle_item_ux', false)
         : false;
 
+    // The item-count heading is hidden only in the mobile cart modal. There, bundle children are
+    // not rendered while the bundle experiment is off, so gate the backorder toggle behind the
+    // experiment to stop it appearing when only hidden bundle children are backordered.
+    const isMobileCartModal = !displayLineItemsCount;
```
bc-maxy: nit: what do you think about adding isMobileCartModal prop explicitly? This coupling is very hidden right now

packages/core/src/app/order/OrderSummaryModal.test.tsx
```diff
@@ -81,6 +82,113 @@ describe('OrderSummaryModal', () => {
...
+                        quantityOutOfStock: 0,
+                        backorderMessage: 'Ships in 5 days',
+                    },
+                },
+            ],
+            digitalItems: [],
+            giftCertificates: [],
+        };
+
+        beforeEach(() => {
```
bc-maxy, line 119: Do we need to do

`resetBackorderDetailsExpanded()` between tests here?

agchan12:
Didn't even know Claude posted this on my behalf :laughing:. Need to update my Claude settings

packages/core/src/app/order/OrderSummaryItems.tsx
animesh1987: OrderSummaryItemBackorderDetails has its own check, this check is to gate the toggle itself appearing altogether. They serve different things. Once the experiment code is removed I can work to refactor to a common method.

packages/core/src/app/order/OrderSummaryItems.tsx
animesh1987: `displayLineItemsCount` is only true for desktop, so we need this for mobile screens. This is a reason why it didn't work straightaway as part of desktop work.

packages/core/src/app/order/OrderSummaryItems.test.tsx
animesh1987: They are different configurations, so it can be setup that way.

packages/core/src/app/order/OrderSummaryItems.tsx
animesh1987: Yes as useState drives the re-render. And there is one module level value and another value for capture the checked prop. So since one value persists the selection across the `MobileView` unmount/remount, module level value won't trigger the re-render hence we need both for now.

packages/core/src/app/order/OrderSummaryModal.test.tsx
animesh1987: Done.

packages/core/src/app/order/OrderSummaryItems.tsx
animesh1987: Good idea, added an `isMobileCartModal` prop that makes it more deterministic.

bc-maxy (dismissed):

packages/core/src/app/order/OrderSummaryItems.tsx
richapatel1510: I meant we don't keep this new function but just instead call `setBackorderDetailsExpanded(false)` there where `resetBackorderDetailsExpanded()` is called.

packages/core/src/app/order/OrderSummaryItems.tsx
richapatel1510: Yes, but is this true understanding that 
`when displayLineItemsCount is false, it is always mobile cart`? 
If so, do we need a new flag `isMobileCartModal`?

animesh1987: Added that since @bc-maxy suggested it to keep the logic more explicit and not tie he mobile modal to lineItems logic. My goal is to refactor everything as a follow up to only rely on `isMobileCartModal`.

packages/core/src/app/order/OrderSummaryItems.tsx
richapatel1510: Do we use this anywhere actually? `resetBackorderDetailsExpanded`?

richapatel1510 (approved):

bc-maxy (approved):

bc-launchbay:
success [20260623054312](https://launchbay.bigcommerce.net/projects/535/releases/288938) by Launchbay

bc-launchbay:
success [20260623054312](https://launchbay.bigcommerce.net/projects/535/releases/288938) by Launchbay

bc-launchbay:
success [20260623054312](https://launchbay.bigcommerce.net/projects/535/releases/288938) by Launchbay

PR #3124, opened 2026-06-26
feat(checkout): CHECKOUT-10140 add extra fields to billing and consignment calls

richapatel1510:
### What

- Extra fields now flow from the address form to the billing/shipping/consignment
  API calls through a single mapper (`mapAddressFromFormValues`), which attaches
  `extraFields` only when the form collected them (no-op for B2C).
- Removed the `getAddressWithCustomerExtraFields` helper (and its test). `Billing`
  and `useShipping` no longer graft `extraFields` from the customer address book —
  they use the checkout-state address (`getBillingAddress()` / `getShippingAddress()`)
  directly.
- `mapAddressExtraFieldsFromFormValues` strips the `B2B_EXTRA_FIELD_PREFIX` from the
  emitted `fieldId`, so form keys (`b2bExtraField_100`) round-trip to the raw id
  (`100`) stored in checkout state.
- `isEqualAddress` now includes `extraFields` in its comparison (removed from the
  ignore list).
- Simplified `ConsignmentAddressSelector` to build the address via the shared mapper
  and dropped the redundant conditional strip on the consignment update path (still
  strips for the customer address-book save).
- Updated tests accordingly (`useShipping`, `isEqualAddress`,
  `mapAddressFromFormValues`, `mapAddressExtraFieldsFromFormValues`).
- Added `stripCustomerAddressFields` to drop `CustomerAddress`-only metadata (`id`, `type`, `isShipping`, `isBilling`, `isDefaultShipping`, `isDefaultBilling`), and applied it on the three saved-address select paths (billing, single shipping, consignment) so a clean address is sent to the API. Added unit tests.

### Why

- B2B `extraFields` were not being carried through to the billing and consignment
  create/update calls. Routing them through one mapper makes a single, reliable path
  for all address submissions instead of reconstructing them after the fact.
- The old address-book grafting depended on an exact address match and didn't cover
  every entry point; with extra fields now round-tripping on the checkout-state
  address, it's no longer needed.
- Without prefix normalization, a round-tripped B2B address carried prefixed ids
  while checkout state stored raw ids, so `isEqualAddress` saw an unchanged address
  as different and refired `updateBillingAddress` / `updateShippingAddress` on every
  step.
- Selecting a saved address passed the raw `CustomerAddress` straight to the update calls. Those fields aren't part of `AddressRequestBody` and the SDK doesn't strip them, so we were forwarding address-book metadata that doesn't belong in the request. The form-submit paths were already clean; this closes the same gap on the select paths.

## Rollout/Rollback

Revert this PR.

## Testing

- CI checks
- E2E tests cannot be recorded since BE changes are still in progress.

https://github.com/user-attachments/assets/5d0cf842-fc11-4f39-ac5f-1d981c1859d1

---

> [!NOTE]
> **Medium Risk**
> Changes what address payloads are sent on billing and consignment update paths and how equality is detected; wrong extra-field shape could cause extra API updates or missing B2B data until backend is aligned.
> 
> **Overview**
> **B2B address `extraFields` now ride along on billing and consignment/shipping updates** through `mapAddressFromFormValues` when the form actually collected them (unchanged for B2C).
> 
> The old **address-book grafting** (`getAddressWithCustomerExtraFields`) and **stripping** (`stripExtraFieldsFromAddress`) paths are removed. **Billing** and **`useShipping`** use checkout-state addresses as-is; **consignment** select/save flows send the mapped address (including `extraFields`) instead of stripping them before `updateConsignment`.
> 
> **`mapAddressExtraFieldsFromFormValues`** normalizes form keys by stripping `B2B_EXTRA_FIELD_PREFIX` so `fieldId` matches checkout state. **`isEqualAddress`** now compares `extraFields`, which should stop redundant `updateBillingAddress` / `updateShippingAddress` calls when B2B values round-trip correctly.
> 
> **ConsignmentAddressSelector** skips saving to the BC customer address book when **`hasCompanyAddressBook`** is enabled.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 352b4d23cb9800cea7ca656e6c96fcd3bbfc3af0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


richapatel1510:
cursor review

packages/core/src/app/address/isEqualAddress.ts
```diff
@@ -50,7 +50,6 @@ function normalizeAddress(address: ComparableAddress) {
         'type',
         'email',
         'country',
-        'extraFields',
```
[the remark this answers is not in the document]
richapatel1510: @bc-peng What do you think about now getting rid of that prefix `b2bExtraField_` since we don't want to now store them in session storage?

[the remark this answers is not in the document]
bc-peng: ~~For address extra fields, we will rely on the checkout API as the single source of truth.~~
~~Therefore, it should be the best practice to avoid local prefix now.~~

We need to keep the prefix to identify an extra field in a flat form.

richapatel1510:
cursor review

packages/core/src/app/shipping/hooks/useShipping.ts
```diff
@@ -148,19 +147,9 @@ export const useShipping = () => {
     const shippableItemsCount = getShippableItemsCount(cart);
     const shouldShowMultiShipping = hasMultiShippingEnabled && !methodId && shippableItemsCount > 1;
 
-    const rawShippingAddress =
```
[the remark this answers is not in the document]
richapatel1510: checkout addresses will now have extraFields when saved customer have them.

richapatel1510:
cursor review

richapatel1510:
cursor review

packages/core/src/app/address/isEqualAddress.ts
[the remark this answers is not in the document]
richapatel1510: Not applicable. This assumes the form emits empty extra-field placeholders that
checkout state lacks, but an empty extra field carries its default value and is
stored in checkout state the same way the form emits it. Both sides carry the
entry, so `isEqualAddress` already matches an unchanged address — there's no
false "changed" result. Empty is a real value here, not an absent field, so the
`customFields`-style empty filter doesn't apply to `extraFields`.

packages/core/src/app/address/mapAddressFromFormValues.ts
```diff
@@ -25,5 +28,10 @@ export default function mapAddressFromFormValues(
         ...address,
         shouldSaveAddress,
         customFields: mapCustomFormFieldsFromFormValues(customFields),
+        // Only carries extra-field values when the form collected them, which only
+        // happens when the `hasAddressExtraFields` capability renders the inputs.
+        // This keeps both billing and shipping/consignment calls sending extra
+        // fields from a single place, and is a no-op for B2C.
+        ...(extraFields ? { extraFields: mapAddressExtraFieldsFromFormValues(extraFields) } : {}),
```
bc-peng, line 35: Is there any issue if we do `extraFields: mapAddressExtraFieldsFromFormValues(extraFields)` just like how we handle `customFields`? I guess, in that case, an empty extraFields will look like `extraFields: []` in the payload.

packages/core/src/app/address/stripCustomerAddressFields.ts
```diff
@@ -0,0 +1,18 @@
...
+const CUSTOMER_ADDRESS_ONLY_FIELDS = [
+    'id',
+    'type',
+    'isShipping',
+    'isBilling',
+    'isDefaultShipping',
+    'isDefaultBilling',
+];
+
+export default function stripCustomerAddressFields<T extends Partial<Address>>(address: T): T {
```
bc-peng: I understand we are removing some B2B company address fields.
Maybe we can rename it to `stripCompanyAddressFields`?

packages/core/src/app/shipping/ConsignmentAddressSelector.tsx
```diff
@@ -117,14 +122,14 @@ const ConsignmentAddressSelector = ({
...
+        await handleSelectAddress(address);
 
-        if (!isGuest) {
+        // Skip the BC customer address-book save when the B2B company address book
+        // is in use: that book is managed by B2B and the API won't accept the
+        // extra fields anyway. Otherwise save without extra fields, as before.
+        if (!isGuest && !hasCompanyAddressBook) {
             try {
-                await createCustomerAddress(address);
+                await createCustomerAddress(stripExtraFieldsFromAddress(address));
```
bc-peng: I see this is a safe approach.
Just want to confirm, in theory, a B2C flow's customer address cannot have extra fields that need to be stripped, right?

richapatel1510: Yes, that's correct. A B2C flow's customer address cannot have extra fields.

richapatel1510: Removed it as well, since not needed. https://github.com/bigcommerce/checkout-js/pull/3124/commits/1e3026782fc4b4815c79ede5a19ed0abb5c2e46e

packages/core/src/app/address/stripCustomerAddressFields.ts
richapatel1510: https://github.com/bigcommerce/checkout-js/pull/3124/commits/1e3026782fc4b4815c79ede5a19ed0abb5c2e46e

packages/core/src/app/address/mapAddressFromFormValues.ts
richapatel1510: Tested it, `extraFields` will be undefined in case of B2C flow, so normal mapping should be okay.
https://github.com/bigcommerce/checkout-js/pull/3124/commits/1e3026782fc4b4815c79ede5a19ed0abb5c2e46e

packages/core/src/app/address/stripCompanyAddressFields.ts
```diff
@@ -0,0 +1,18 @@
+import { type Address } from '@bigcommerce/checkout-sdk';
+import { omit } from 'lodash';
+
+// `CustomerAddress` extends `Address` with address-book metadata that is not part
+// of `AddressRequestBody`. When a saved address is selected it's passed straight to
+// the update calls, so strip these so we don't forward them to the API.
+const CUSTOMER_ADDRESS_ONLY_FIELDS = [
```
animesh1987: Should we keep this logic in SDK?

packages/core/src/app/address/mapAddressFromFormValues.ts
richapatel1510: Actually will not do that because the unconditional version similar to customFields puts `extraFields: undefined` on the object, while checkout state omits the key entirely in case of B2C — and `isEqual({extraFields: undefined}, {})` is `false` which we don't want.

packages/core/src/app/address/stripCompanyAddressFields.ts
bc-peng: What happens if our API sees those fields? Will they be simply ignored?

Note for the order sync discussion:
- we need to add `isShipping` and `isBilling` back when doing the post order sync for saving a new address. BE should be able to derive those values from the placement of an address.

richapatel1510: For now they are ignored. No effect. Do we want to keep sending them for now?

richapatel1510: Removed from here, will raise a separate PR to add this to SDK https://github.com/bigcommerce/checkout-js/pull/3124/commits/352b4d23cb9800cea7ca656e6c96fcd3bbfc3af0

bc-peng (approved):
LGTM. 👏👏

packages/core/src/app/address/stripCompanyAddressFields.ts
richapatel1510: https://github.com/bigcommerce/checkout-sdk-js/pull/3303

animesh1987 (approved):

bc-launchbay:
success [20260701002813](https://launchbay.bigcommerce.net/projects/535/releases/289877) by Launchbay

bc-launchbay:
success [20260701002813](https://launchbay.bigcommerce.net/projects/535/releases/289877) by Launchbay

bc-launchbay:
success [20260701002813](https://launchbay.bigcommerce.net/projects/535/releases/289877) promoted by https://launchbay.bigcommerce.net/projects/535/releases/289892 being deployed. Deployed by Launchbay

PR #3130, opened 2026-06-30
chore(checkout): BACK-714 Display backorder info for picklist items order confirmation page

animesh1987:
## What/Why?
Display backorder info for picklist items order confirmation page

## Rollout/Rollback
- revert this PR
## Testing
- CI
- Screencast 

https://github.com/user-attachments/assets/7fd406ac-4d4a-4187-93b9-0598227b6d37

---

> [!NOTE]
> **Medium Risk**
> Touches order-summary bundling and backorder display for picklist bundles on confirmation; wrong attribute linking could hide or mis-nest line items, but behavior is gated behind the bundle experiment with a cart fallback.
> 
> **Overview**
> When the picklist bundle experiment is on, **order summary** now prefers **`order.bundledItems`** from checkout state (via SDK **1.935.1**) to attach picklist children to parents using **`addedByAttributeId`** / option **`attributeId`**, instead of relying only on **`parentId`** in cart line items. That path is used on surfaces like **order confirmation**, where bundled children may not appear as normal line items but still need to render under the parent for **backorder details**.
> 
> Adds **`buildBundleItemsMapFromOrder`** in `removeBundledItems.ts` (filters bundled ids from top-level items, builds the same **`bundleItemsMap`** shape as **`removeAndBundleItemsTogether`**). **`OrderSummaryItems`** loads **`order`** from **`useCheckout`** and falls back to the existing cart bundling logic when **`bundledItems`** is absent.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 41933576bd5df4d4967b9bfaa824f198a740d3f2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/app/order/OrderSummaryItems.tsx
+++ b/packages/core/src/app/order/OrderSummaryItems.tsx
@@ -29,7 +29,11 @@ import mapFromDigital from './mapFromDigital';
 import mapFromGiftCertificate from './mapFromGiftCertificate';
 import mapFromPhysical from './mapFromPhysical';
 import OrderSummaryItem from './OrderSummaryItem';
-import { removeAndBundleItemsTogether, removeBundledItems } from './removeBundledItems';
+import {
+    buildBundleItemsMapFromOrder,
+    removeAndBundleItemsTogether,
+    removeBundledItems,
+} from './removeBundledItems';
 
 // Module-scoped to survive the responsive remount. Safe as MobileView mounts only one instance at a time.
 let backorderDetailsExpanded = false;
@@ -194,7 +198,12 @@ const OrderSummaryItems = ({
 }: OrderSummaryItemsProps): ReactElement => {
     const [isExpanded, setIsExpanded] = useState(false);
     const [showBackorderDetails, setShowBackorderDetails] = useState(getBackorderDetailsExpanded);
-    const { selectedState: config } = useCheckout(({ data }) => data.getConfig());
+    const {
+        selectedState: { config, order },
+    } = useCheckout(({ data }) => ({
+        config: data.getConfig(),
+        order: data.getOrder(),
+    }));
 
     const toggleBackorderDetails = useCallback(() => {
         setShowBackorderDetails((prev) => {
@@ -229,7 +238,9 @@ const OrderSummaryItems = ({
     const expandBackorderDetails = showBackorderSwitch && showBackorderDetails;
 
     const { nonBundledItems, bundleItemsMap } = pickListExperimentEnabled
-        ? removeAndBundleItemsTogether(items)
+        ? order?.bundledItems
+            ? buildBundleItemsMapFromOrder(items, order.bundledItems)
+            : removeAndBundleItemsTogether(items)
         : { nonBundledItems: removeBundledItems(items), bundleItemsMap: undefined };
 
     const collapsedLimit = isSmallScreen()
--- a/packages/core/src/app/order/removeBundledItems.ts
+++ b/packages/core/src/app/order/removeBundledItems.ts
@@ -7,42 +7,48 @@ export function buildBundleItemsMapFromOrder(
     nonBundledItems: LineItemMap;
     bundleItemsMap: Map<string | number, Array<PhysicalItem | DigitalItem>>;
 } {
-    const attrToChild = new Map<string, PhysicalItem | DigitalItem>();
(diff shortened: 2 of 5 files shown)


animesh1987:
cursor review

packages/core/src/app/order/OrderSummaryItems.tsx
```diff
@@ -229,7 +238,9 @@ const OrderSummaryItems = ({
     const expandBackorderDetails = showBackorderSwitch && showBackorderDetails;
 
     const { nonBundledItems, bundleItemsMap } = pickListExperimentEnabled
-        ? removeAndBundleItemsTogether(items)
+        ? order?.bundledItems
+            ? buildBundleItemsMapFromOrder(items, order.bundledItems)
+            : removeAndBundleItemsTogether(items)
```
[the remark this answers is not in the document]
animesh1987: That is fine, since they are supposed to be part of a bundle.

packages/core/src/app/order/removeBundledItems.ts
```diff
@@ -1,5 +1,59 @@
 import { type DigitalItem, type LineItemMap, type PhysicalItem } from '@bigcommerce/checkout-sdk';
 
+export function buildBundleItemsMapFromOrder(
```
bc-shawnwang, line 3: since it's attributeId -> bundledItem mapping, I feel it doesn't cover the case when:

There are items split to multiple items because of multi shipping, their picklist options' attributeId is same. bundled items will be split as well, so addedByAttributeId is same but parent item id is different. 

in sum, bundled item should be linked based on both parentItemId and addByAttributeId (checkout page seems correct because it's first map by parent id in first layer, then pickup by attributeId)

So can you check and test on this case:
set product B's stock status as: 3 onhand, 100 backorder.
set B as the picklist option of product A, and check "adjust inventory"

Then in storefront add 10 product A with picklist of product B to cart
checkout page add two consignments, one is 5 x productA, another is 5 x product A

see checkout page and order confirmation page, if each bundled item's stock position is correct

agchan12, line 3: Is it possible to add test coverage for this?

packages/core/src/app/order/OrderSummaryItems.tsx
```diff
@@ -229,7 +238,9 @@ const OrderSummaryItems = ({
     const expandBackorderDetails = showBackorderSwitch && showBackorderDetails;
 
     const { nonBundledItems, bundleItemsMap } = pickListExperimentEnabled
-        ? removeAndBundleItemsTogether(items)
+        ? order?.bundledItems
```
agchan12, line 241: What happens here if `order.bundledItems` is present but empty? I think the truthiness check would pass and we'd skip the parent ID based fallback? Would we want to fall through to `removeAndBundleItemsTogether` in that case?

animesh1987: This is an implicit logic to decide where the shopper is, on checkout page or order confirmation page. `removeAndBundleItemsTogether ` is the method used for checkout page and other one for order confirmation page since they have different shape of data returned via the API.

packages/core/src/app/order/removeBundledItems.ts
animesh1987: This is covered by the component tests for order confirmation page. But I add test for the method itself as well.

bc-shawnwang: for this case, found the api storefront orders api for bundled item miss the parent order product id,  need the api to add it so FE can fix in this case. So ignore now.

bc-shawnwang (approved):

packages/core/src/app/order/removeBundledItems.ts
animesh1987: We don't have parentId, will have to expose it in API and update implementation as follow up.

bc-launchbay:
success [20260701071259](https://launchbay.bigcommerce.net/projects/535/releases/289913) by Launchbay

bc-launchbay:
success [20260701071259](https://launchbay.bigcommerce.net/projects/535/releases/289913) by Launchbay

bc-launchbay:
success [20260701071259](https://launchbay.bigcommerce.net/projects/535/releases/289913) by Launchbay

PR #3135, opened 2026-07-02
feat(checkout): CHECKOUT-10026 Add new Google Places API services

bc-maxy:
## What/Why?

Adds a new, self-contained `newGooglePlacesApi/` module (script loader, session-token service, mapping utils) implementing the new version of Google's Places API (Autocomplete). This is Part 1 of the work and in Part 2 I actually will wire it up with our existing service and with our new experiment flag.

This PR also includes a small type-only refactor so `GoogleAutocompleteService` depends on a shared `GoogleMapsPlacesScriptLoader` interface instead of the legacy loader class, enabling loader sharing in a later PR.

As mentioned, nothing is wired up yet so no existing code in production calls the new module, and GoogleAutocompleteService's behavior is unchanged. This should have zero behavior change.

## Rollout/Rollback

Revert the PR, but in practice the revert of this PR will do nothing - it purely adds the service without connecting it to any components.

## Testing

New CI tests.

---

> [!NOTE]
> **Low Risk**
> Production paths are untouched; changes are additive module code plus a type-only constructor dependency on GoogleAutocompleteService with the same default loader.
> 
> **Overview**
> Introduces a self-contained **`newGooglePlacesApi/`** stack for Google’s **Places (New)** autocomplete flow—script loading via `importLibrary`, **`NewGooglePlacesApiService`** for suggestions and place details, and mapping helpers so responses still look like legacy **`PlaceResult`** / **`AutocompleteItem`** shapes. Session tokens, prediction caching, and **`toPlace()`** on selection are handled in the service; **`isNewPlacesApiPermissionDenied`** is included for future gRPC error handling.
> 
> **`GoogleAutocompleteService`** now depends on a shared **`GoogleMapsPlacesScriptLoader`** interface instead of the legacy loader class, so a follow-up can inject **`NewGooglePlacesApiScriptLoader`** (which also implements **`loadMapsSdk`** for legacy **`AutocompleteService`/`PlacesService`**) and avoid loading Maps JS twice.
> 
> **No checkout wiring or feature flag usage in this PR**—existing autocomplete behavior is unchanged until Part 2.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 325fa175b5a9328c9b233c4fedf7303abf4de00f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #3142, opened 2026-07-06
feat(checkout): CHECKOUT-10026 Connect new Google Places API behind feature flag

bc-maxy:
## What/Why?

This is part II of this change: https://github.com/bigcommerce/checkout-js/pull/3135, where I actually wire up this new Places API service to production, but hide new functionality behind a feature flag.

When the [CHECKOUT-10026.new_google_places_api](https://app.launchdarkly.com/projects/default/flags/CHECKOUT-10026.new_google_places_api) experiment is on, address autocomplete uses Google's new Places API (session-token based, AutocompleteSuggestion/Place classes) instead of the legacy API service. If the new API is denied due to permissions error (`gRPC PERMISSION_DENIED`), we assume that the service is not enabled for the store and silently fall back to the legacy API for the rest of the session. No visible errors for users and no visible change between old and new service - this is an infrastructure swap behind a feature flag.

NOTE: when the flag is off, there will be no change for end users, However, the code won't be byte-to-byte identical: we initialize `newGooglePlacesApiServiceRef.current` with `new NewGooglePlacesApiService(apiKey)` unconditionally on mount, even when the flag is off. It has no side effects though (no network call, no script load - the script loader inside of it is lazy), so this is just a trivial extra allocation without any functional change for end users.

[CHECKOUT-10026.new_google_places_api](https://app.launchdarkly.com/projects/default/flags/CHECKOUT-10026.new_google_places_api) is currently set to `true` in integration and staging and `false` for everyone in production.

## Rollout/Rollback

Set [CHECKOUT-10026.new_google_places_api](https://app.launchdarkly.com/projects/default/flags/CHECKOUT-10026.new_google_places_api) experiment to false if the store where we rolled out the feature is affected. For issues with other stores revert the PR.

## Testing

If you want to test yourself you can use these API keys in https://console.cloud.google.com/apis/credentials?project=bc-hackathon-env:
1. **Google Maps API Key test** - our old API key that still has access to **old Places API** and has no access to new Places API. Represents old users.
2. **Maxs Places API Key Test** - our new API key that has access to **new Places API** and no access to old Places API. Represents new users.

My tests are below:

### With experiment set to false (same as current production)

Existing customers with old API keys:

https://github.com/user-attachments/assets/47e3a4c0-b98c-43b0-b35c-37bafd7816c8

New customers with new API keys (doesn't work):

https://github.com/user-attachments/assets/d4a50153-7a1a-4e94-8046-307d195ff895

### With experiment set to true

Existing customers with old API keys (Initial 403, then fallback to the old service. Billing step will remember that we fallback and skips new api):

https://github.com/user-attachments/assets/6215bf66-7765-40be-81de-5b93e4ae9530

New customers with new API keys (just works):

https://github.com/user-attachments/assets/bc6aea36-d54d-426b-a606-73cd271ad79b

---

> [!NOTE]
> **Medium Risk**
> Touches core checkout address capture and external Google API behavior; rollout is flag-gated with legacy fallback, but wrong fallback rules could break autocomplete for some API key configurations.
> 
> **Overview**
> Checkout address line 1 autocomplete can use Google's **new Places API** when experiment `CHECKOUT-10026.new_google_places_api` is on, passed from `AddressForm` through `GoogleAutocompleteFormField` as `isNewPlacesApiEnabled`.
> 
> Autocomplete logic moves into **`useGoogleAutocomplete`**, which prefers the new API for suggestions and place details and **falls back to the legacy Maps Places service** only on permission-denied (`gRPC` code 7). That denial **latches for the session** (module state) so later keystrokes skip the new API; transient errors clear suggestions or skip `onSelect` without latching. With the flag off, only legacy runs. Legacy shares the new API's script loader when the flag is on to avoid double-loading Maps JS.
> 
> `GoogleAutocomplete` is thinned to the hook plus UI; legacy prediction mapping lives in `utils.ts`. A broad **`GoogleAutocomplete.test.tsx`** suite covers happy path, fallback, latch, and flag-off behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 25a61c18b576c3b1a579c85fd06bc3959c26fca0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/app/address/AddressForm.tsx
+++ b/packages/core/src/app/address/AddressForm.tsx
@@ -61,6 +61,11 @@ const AddressForm: React.FC<AddressFormProps> = ({
         'CHECKOUT-9019.use_new_phone_number_validation',
         false,
     );
+    const isNewGooglePlacesApiEnabled = isExperimentEnabled(
+        config?.checkoutSettings,
+        'CHECKOUT-10026.new_google_places_api',
+        false,
+    );
     const countriesWithAutocomplete = ['US', 'CA', 'AU', 'NZ', 'GB'];
 
     const containerRef = useRef<HTMLDivElement>(null);
@@ -174,6 +179,7 @@ const AddressForm: React.FC<AddressFormProps> = ({
                                     countryCode={countryCode}
                                     field={field}
                                     isFloatingLabelEnabled={isFloatingLabelEnabledValue}
+                                    isNewPlacesApiEnabled={isNewGooglePlacesApiEnabled}
                                     key={field.id}
                                     nextElement={nextElementRef.current || undefined}
                                     onChange={handleAutocompleteChange}
--- a/packages/core/src/app/address/googleAutocomplete/GoogleAutocomplete.tsx
+++ b/packages/core/src/app/address/googleAutocomplete/GoogleAutocomplete.tsx
@@ -1,14 +1,13 @@
 import { noop } from 'lodash';
-import React, { useRef, useState } from 'react';
+import React from 'react';
 
 import { Autocomplete, type AutocompleteItem } from '@bigcommerce/checkout/ui';
 
-import GoogleAutocompleteService from './GoogleAutocompleteService';
 import { type GoogleAutocompleteOptionTypes } from './googleAutocompleteTypes';
-import { NewGooglePlacesApiService } from './newGooglePlacesApi';
-import { getNewGooglePlacesApiScriptLoader } from './newGooglePlacesApi/getNewGooglePlacesApiScriptLoader';
-import { isNewPlacesApiPermissionDenied } from './newGooglePlacesApi/utils';
+import { useGoogleAutocomplete } from './useGoogleAutocomplete';
 import './GoogleAutocomplete.scss';
+
+export { resetNewGooglePlacesApiState } from './useGoogleAutocomplete';
 
 export interface GoogleAutocompleteProps {
     initialValue?: string;
@@ -25,24 +24,6 @@ export interface GoogleAutocompleteProps {
     onChange?(value: string, isOpen: boolean): void;
 }
 
-const toAutocompleteItems = (
-    results?: google.maps.places.AutocompletePrediction[],
-): AutocompleteItem[] => {
-    return (results || []).map((result) => ({
-        label: result.description,
-        value: result.structured_formatting.main_text,
-        highlightedSlices: result.matched_substrings,
-        id: result.place_id,
-    }));
-};
-
-// module-scoped because we want to persist across all checkout steps
-const newGooglePlacesApiState = { isUnavailable: false };
-
-export function resetNewGooglePlacesApiState(): void {
-    newGooglePlacesApiState.isUnavailable = false;
-}
-
 const GoogleAutocomplete: React.FC<GoogleAutocompleteProps> = ({
     initialValue,
     onToggleOpen = noop,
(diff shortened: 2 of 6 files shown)


bc-maxy:
cursor review

packages/core/src/app/address/googleAutocomplete/GoogleAutocomplete.tsx
```diff
@@ -40,41 +51,115 @@ const GoogleAutocomplete: React.FC<GoogleAutocompleteProps> = ({
...
+
+        service
+            .getSuggestions(input, types, componentRestrictions)
+            .then(setItems)
+            .catch((error) => {
+                if (isNewPlacesApiPermissionDenied(error)) {
+                    newGooglePlacesApiState.isUnavailable = true;
+                    fetchSuggestionsWithLegacyApi(input);
+                }
+            });
```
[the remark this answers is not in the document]
bc-maxy: This is only theoretical and does not apply here. There's no scenario where an older request gets PERMISSION_DENIED while a newer, concurrent request on the same key legitimately succeeds. If the key lacks permission to the new service, it can't randomly succeed on a third try - all of the requests will fail. The latch flipping true is just a matter of which rejection happens to land first, not a race between success and failure. If permission is genuinely denied, there's no successful response left in flight to leak through.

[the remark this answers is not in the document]
bc-maxy: Having said that we shouldn't send a new request on every keystroke 😅  I'll add some debounce in a new PR, should cut down autocomplete bills for older api users as well 😅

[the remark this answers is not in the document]
bc-maxy: Resolved in https://github.com/bigcommerce/checkout-js/pull/3142/commits/ee5e99265445cc0563f0380db6d6034665fac993

bc-maxy:
cursor review

packages/core/src/app/address/googleAutocomplete/GoogleAutocomplete.tsx
```diff
@@ -84,42 +182,32 @@ const GoogleAutocomplete: React.FC<GoogleAutocompleteProps> = ({
...
-                    componentRestrictions,
-                },
-                (results) => {
-                    const autocompleteItems = toAutocompleteItems(results ?? undefined);
-
-                    setItems(autocompleteItems);
-                },
-            );
-        });
+        fetchSuggestionsWithNewApi(input);
```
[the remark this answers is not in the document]
bc-maxy: Fixed in https://github.com/bigcommerce/checkout-js/pull/3142/commits/c05b09fa1f2494e798fb5879183a35873b413fa7

bc-maxy:
cursor review

packages/core/src/app/address/googleAutocomplete/GoogleAutocomplete.tsx
```diff
@@ -40,41 +51,128 @@ const GoogleAutocomplete: React.FC<GoogleAutocompleteProps> = ({
...
+        googleAutocompleteServiceRef.current = new GoogleAutocompleteService(
+            apiKey,
+            isNewPlacesApiEnabled ? getNewGooglePlacesApiScriptLoader() : undefined,
+        );
     }
 
-    const onSelectHandler = (item: AutocompleteItem) => {
+    const isUsingLegacyApi = () => !isNewPlacesApiEnabled || newGooglePlacesApiState.isUnavailable;
+
+    const finalizeSelection = (
```
bc-peng: It looks like many lines are added to this file.
Is it possible to move them out as separate file(s) to keep this file small?

bc-peng: The rest of the PR looks good.

packages/core/src/app/address/googleAutocomplete/useGoogleAutocomplete.ts
```diff
@@ -0,0 +1,205 @@
...
+            .getSuggestions(input, types, componentRestrictions)
+            .then((results) => {
+                if (latestNewApiInputRef.current === input) {
+                    setItems(results);
+                }
+            })
+            .catch((error) => {
+                if (isNewPlacesApiPermissionDenied(error)) {
+                    newGooglePlacesApiState.isUnavailable = true;
+                    fetchSuggestionsWithLegacyApi(input);
```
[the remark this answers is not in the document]
bc-maxy: [d23c96d](https://github.com/bigcommerce/checkout-js/pull/3142/commits/d23c96d1bdee9ca8f2165d00c42f795cd97b0733)

packages/core/src/app/address/googleAutocomplete/GoogleAutocomplete.tsx
bc-maxy: Thanks for checking it @bc-peng 🙏 I made changes here: [33ca5c2](https://github.com/bigcommerce/checkout-js/pull/3142/commits/33ca5c299b41e7a6c1bd747cfb97f4b1c5cd51c4). Is this what you had in mind?

packages/core/src/app/address/googleAutocomplete/useGoogleAutocomplete.ts
```diff
@@ -0,0 +1,208 @@
...
+                { input, types: types || ['geocode'], componentRestrictions },
+                (results) => {
+                    setItems(toAutocompleteItems(results ?? undefined));
+                },
+            );
+        });
+    };
+
+    const fetchSuggestionsWithNewApi = (input: string): void => {
+        const service = newGooglePlacesApiServiceRef.current;
```
richapatel1510: Not blocking, just a thought. Does it makes sense to have 2 classes or similar for old and new aPI and then based on the `isUsingLegacyApi` we use the correct class in the hook to avoid naming such as  `fetchSuggestionsWithNewApi/fetchSuggestionsWithLegacyApi`

bc-maxy: Thanks for checking @richapatel1510 🙏 I like this idea. However, I tried it as two classes and initialization and updates ended up more verbose (constructors, and passing render-scoped values like types/fields in on every call) without a real readability gain. Instead, I grouped the existing functions into two objects (legacyApi/newApi), which in my mind gets us the similar naming clarity you were after, without the class overhead. What do you think about that?

richapatel1510 (dismissed):

packages/core/src/app/address/googleAutocomplete/useGoogleAutocomplete.ts
richapatel1510: Makes sense. The naming and readability is better now. Thanks!

packages/core/src/app/address/AddressForm.tsx
```diff
@@ -61,6 +61,11 @@ const AddressForm: React.FC<AddressFormProps> = ({
         'CHECKOUT-9019.use_new_phone_number_validation',
         false,
     );
+    const isNewGooglePlacesApiEnabled = isExperimentEnabled(
+        config?.checkoutSettings,
+        'CHECKOUT-10026.new_google_places_api',
+        false,
```
bc-peng, line 67: If we have surfaced this experiment in checkout settings, we don't have to set this fallback value.
It's intended for that missing case only.

packages/core/src/app/address/googleAutocomplete/useGoogleAutocomplete.ts
```diff
@@ -0,0 +1,210 @@
...
+        }
+
+        onSelect(place, item);
+    };
+
+    const legacyApi = {
+        fetchSuggestions: (input: string): void => {
+            const service = googleAutocompleteServiceRef.current;
+
+            if (!service) return;
```
animesh1987:
```suggestion
            if (!service) {
             return;
             }
```

packages/core/src/app/address/googleAutocomplete/useGoogleAutocomplete.ts
```diff
@@ -0,0 +1,210 @@
...
+                    (results) => {
+                        setItems(toAutocompleteItems(results ?? undefined));
+                    },
+                );
+            });
+        },
+        select: (item: AutocompleteItem): void => {
+            const service = googleAutocompleteServiceRef.current;
+
+            if (!service) return;
```
animesh1987: Same as above.

packages/core/src/app/address/googleAutocomplete/useGoogleAutocomplete.ts
```diff
@@ -0,0 +1,210 @@
...
+    types,
+    componentRestrictions,
+    onSelect,
+    onChange,
+}: UseGoogleAutocompleteProps): UseGoogleAutocompleteResult {
+    const [items, setItems] = useState<AutocompleteItem[]>([]);
+    const [autoComplete, setAutoComplete] = useState<string>('off');
+    const newGooglePlacesApiServiceRef = useRef<NewGooglePlacesApiService>();
+    const googleAutocompleteServiceRef = useRef<GoogleAutocompleteService>();
+    const latestNewApiInputRef = useRef<string>();
```
animesh1987: We can just call this `apiInputRef`, as there is no corresponding old input ref variable here.

packages/core/src/app/address/AddressForm.tsx
bc-maxy: Thank you for bringing this up - looks I forgot to propagate it to settings. Merged this PR today: https://github.com/bigcommerce/bigcommerce/pull/69182 and since it's still going through the pipeline, maybe we keep this false fallback? 😅

bc-peng (approved):
LGTM. 👏
I don't see we change any UX in this PR. Just a bit context if we suddenly block BCapp pipeline.
> two Playwright e2e tests in `functional-tests` repo (GoogleAddressAutoComplete.spec.ts) cover Google autocomplete on checkout for both shipping and billing addresses, using a real Google Maps API key.

animesh1987 (approved):

bc-launchbay:
success [20260715000317](https://launchbay.bigcommerce.net/projects/535/releases/291347) by Launchbay

bc-launchbay:
success [20260715000317](https://launchbay.bigcommerce.net/projects/535/releases/291347) by Launchbay

bc-launchbay:
success [20260715000317](https://launchbay.bigcommerce.net/projects/535/releases/291347) by Launchbay

PR #3170, opened 2026-07-20
refactor(checkout): CHECKOUT-10241 Add CannotCreatePersonalAccountSessionStorage

bc-peng:
## What/Why?

Use session storage to:
- pass `cannotCreatePersonalAccount` from the checkout page to the order confirmation page.

Because the capability object is currently unavailable on the order confirmation page.

## Rollout/Rollback

Revert.

### When `cannotCreatePersonalAccount` is `true`, sign-up form should be hidden.

https://github.com/user-attachments/assets/c07b64a6-5dfd-490a-80c9-fcc4f78d4e7e

### When `cannotCreatePersonalAccount` is `false`, sign-up form should be on display.

https://github.com/user-attachments/assets/eb5a039c-c8f2-4dcf-8d37-6f0535bcecf8

---

> [!NOTE]
> **Low Risk**
> Scoped to guest account UI on order confirmation; uses short-lived session storage with no auth or payment changes.
> 
> **Overview**
> **Order confirmation** no longer reads `cannotCreatePersonalAccount` from capabilities (unavailable on that page). The flag is written to **session storage** on checkout when redirecting after place order, then read once on order confirmation mount and cleared so the guest create-account / set-password UI still respects the capability.
> 
> Adds `CannotCreatePersonalAccountSessionStorage` in `@bigcommerce/checkout/utility` and wires **CheckoutPage** to persist the value before `navigateToOrderConfirmation`. **OrderConfirmation** drops `useCapabilities` for this flag in favor of the storage helper. Tests cover storage round-trip, checkout persistence on navigation, and one-time consumption on the confirmation page.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 29b71b8f7e609a60e28e70522027fb8c52440f09. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/app/order/OrderConfirmation/OrderConfirmation.tsx
+++ b/packages/core/src/app/order/OrderConfirmation/OrderConfirmation.tsx
@@ -5,10 +5,13 @@ import {
 import { createRequestSender } from '@bigcommerce/request-sender';
 import React, { type ReactElement, useEffect, useRef, useState } from 'react';
 
-import { useAnalytics, useCapabilities, useCheckout } from '@bigcommerce/checkout/contexts';
+import { useAnalytics, useCheckout } from '@bigcommerce/checkout/contexts';
 import { type ErrorLogger } from '@bigcommerce/checkout/error-handling-utils';
 import { OrderConfirmationPageSkeleton } from '@bigcommerce/checkout/ui';
-import { isExperimentEnabled } from '@bigcommerce/checkout/utility';
+import {
+    CannotCreatePersonalAccountSessionStorage,
+    isExperimentEnabled,
+} from '@bigcommerce/checkout/utility';
 
 import { type EmbeddedCheckoutStylesheet } from '../../embeddedCheckout';
 import { type CreatedCustomer, type SignUpFormValues } from '../../guestSignup';
@@ -64,9 +67,13 @@ export const OrderConfirmation = ({
         isLoadingOrder: statuses.isLoadingOrder(),
     }));
     const { analyticsTracker } = useAnalytics();
-    const {
-        orderConfirmation: { cannotCreatePersonalAccount },
-    } = useCapabilities();
+    const [cannotCreatePersonalAccount] = useState(() => {
+        const value = CannotCreatePersonalAccountSessionStorage.getCannotCreatePersonalAccount();
+
+        CannotCreatePersonalAccountSessionStorage.removeCannotCreatePersonalAccount();
+
+        return value;
+    });
 
     const handleUnhandledError = (e: Error) => {
         setError(e);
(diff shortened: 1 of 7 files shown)


packages/core/src/app/order/OrderConfirmation/OrderConfirmation.tsx
```diff
@@ -64,9 +67,13 @@ export const OrderConfirmation = ({
         isLoadingOrder: statuses.isLoadingOrder(),
     }));
     const { analyticsTracker } = useAnalytics();
-    const {
-        orderConfirmation: { cannotCreatePersonalAccount },
-    } = useCapabilities();
+    const [cannotCreatePersonalAccount] = useState(() => {
+        const value = CannotCreatePersonalAccountSessionStorage.getCannotCreatePersonalAccount();
+
+        CannotCreatePersonalAccountSessionStorage.removeCannotCreatePersonalAccount();
```
bc-maxy, line 76: What do you think about keeping the useState initializer pure and moving the removeCannotCreatePersonalAccount() call into a mount effect?

```
useEffect(() => {
    CannotCreatePersonalAccountSessionStorage.removeCannotCreatePersonalAccount();
}, []);
```

Or, since we write to session storage on every purchase (including false), do we need to remove it at all? Right now, if the shopper refreshes the confirmation page, the flag is already consumed, so the create-account form would reappear even though the capability says it shouldn't.

bc-peng: > the create-account form would reappear even though the capability says it shouldn't.
You are right on this case.

This is intended as a short-term solution.
- We still want the API to remain the single source of truth, and clearing session storage promptly helps avoid unnecessary exposure.

That is why the deletion happens here.



> What do you think about keeping the useState initializer pure and moving the removeCannotCreatePersonalAccount() call into a mount effect?

I considered using `useEffect` to ensure the deletion happens only once. However, since we also need to read session storage only once during the app lifecycle, `useState` seems necessary, while `useEffect` may be optional.

I'm happy to apply `useEffect` if that would help improve the code readability? What do you think?

animesh1987: I believe we will need to have the useState + useEffect for this to work properly? State to capture the value, and effect to set and remove value from session storage?

bc-peng: The single `useState` does work properly, but it's hard to read.

I have updated it as `useState + useEffect`.

[29b71b8](https://github.com/bigcommerce/checkout-js/pull/3170/commits/29b71b8f7e609a60e28e70522027fb8c52440f09)

animesh1987 (approved):

bc-launchbay:
success [20260721014847](https://launchbay.bigcommerce.net/projects/535/releases/291892) by Launchbay

bc-launchbay:
success [20260721014847](https://launchbay.bigcommerce.net/projects/535/releases/291892) by Launchbay

bc-launchbay:
success [20260721014847](https://launchbay.bigcommerce.net/projects/535/releases/291892) promoted by https://launchbay.bigcommerce.net/projects/535/releases/291904 being deployed. Deployed by Launchbay

PR #3189, opened 2026-07-23
feat(checkout): CHECKOUT-10208 Accessibility - payment radio buttons keyboard operable

bc-charlesho:
### What
Allows the payment method radio buttons on checkout to be accessible using the keyboard's up/down arrow keys.

- Wraps the `Checklist` in a `LoadingOverlay` to communicate the loading state visually, instead of disabling the controls.
- Keeps the payment method radios **enabled and keyboard-operable** during initialisation, so keyboard/assistive technology users retain access.
- Adds a new `isReadOnly` prop to the shared `ChecklistItem` component (`packages/ui`): it overrides the input's `onChange` with a no-op and sets `aria-disabled`, keeping the radio focusable while preventing it from being toggled — unlike `isDisabled`, which uses the native `disabled` attribute and removes it from the tab order.
- `PaymentMethodList` threads `isInitializingPayment` down to each item as `isReadOnly`, so selection is blocked while a method is initializing, without affecting keyboard reachability.

### Why?
Restores keyboard and screen-reader access to the payment method selector during initialisation, while still signalling the section is loading and preventing a method from being changed mid-initialization.

**Problem:** the original disabled-radio approach removed items from the tab order, breaking keyboard/AT access. An earlier attempt at fixing this only blocked the `onMethodSelect` callback, but testing showed Formik's own field `onChange` (wired directly to the radio input) still changed `paymentProviderRadio` and the visually selected item — silently desyncing the selected method from what was actually being initialized/validated/submitted. The `isReadOnly` prop closes that gap at the source, by intercepting the input's `onChange` directly rather than relying on a downstream guard.
## Rollout/Rollback
Revert the PR.

## Testing
- Added `ChecklistItem.test.tsx` coverage: with `isReadOnly`, the radio stays focusable/enabled but doesn't get checked when clicked.
- Added `PaymentForm.test.tsx` coverage: clicking a different payment method radio during `isInitializingPayment` does not change which radio is checked, and does not call `onMethodSelect`.

### Before
Shoppers can't tab into the payment method radio button group, and can't select an option using the up/down arrow keys.

**Theme v2**
https://github.com/user-attachments/assets/73763df5-52a1-48b5-a1df-b857dcfb5d8b

**Theme v1**
https://github.com/user-attachments/assets/84f743f4-8eab-4d28-92b7-f9a5f79ce474

### After
Shoppers can tab into the payment method radio button group, and select an option using the up/down arrow keys.

**Theme v2**
https://github.com/user-attachments/assets/4b146e15-ad55-4fbf-b4d0-1b172442c207

**Theme v1**
https://github.com/user-attachments/assets/a37e5b5a-bc10-4d34-bce0-4a473113bc20

---

> [!NOTE]
> **Medium Risk**
> Touches the checkout payment selection path and Formik wiring; behavior is constrained by tests and only applies during payment initialization, but regressions could affect which method is submitted.
> 
> **Overview**
> Fixes checkout payment method selection during initialization so keyboard and assistive-tech users can still reach the radio group, while blocking actual method changes until init finishes.
> 
> **Payment method list:** The checklist is no longer disabled during `isInitializingPayment`. A `LoadingOverlay` shows loading instead. Each `ChecklistItem` gets `isReadOnly` while initializing (cheque/PO disable reasons still use `isDisabled` only).
> 
> **Shared `ChecklistItem`:** New optional `isReadOnly` sets `aria-disabled` and swaps the radio’s `onChange` for a no-op so Formik’s field value does not update on click—unlike `disabled`, which removes radios from the tab order.
> 
> Tests cover read-only checklist behavior and that `PaymentForm` does not call `onMethodSelect` or change the checked radio when initializing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a7354d561a1b0cecad9de0da38fe759491da8113. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/app/payment/paymentMethod/PaymentMethodList.tsx
+++ b/packages/core/src/app/payment/paymentMethod/PaymentMethodList.tsx
@@ -3,7 +3,7 @@ import { find, get, noop } from 'lodash';
 import React, { type FunctionComponent, memo, useCallback, useMemo } from 'react';
 
 import { useCheckout, useLocale } from '@bigcommerce/checkout/contexts';
-import { Checklist, ChecklistItem } from '@bigcommerce/checkout/ui';
+import { Checklist, ChecklistItem, LoadingOverlay } from '@bigcommerce/checkout/ui';
 
 import { connectFormik, type ConnectFormikProps } from '../../common/form';
 import { isMobile } from '../../common/utility';
@@ -87,47 +87,46 @@ const PaymentMethodList: FunctionComponent<
             <div aria-live="assertive" className="is-srOnly" role="status">
                 {titleText}
             </div>
-            <Checklist
-                defaultSelectedItemId={values.paymentProviderRadio}
-                isDisabled={isInitializingPayment}
-                name="paymentProviderRadio"
-                onSelect={handleSelect}
-            >
-                {methods.map((method) => {
-                    const value = getUniquePaymentMethodId(method.id, method.gateway);
-                    const showOnlyOnMobileDevices = get(
-                        method,
-                        'initializationData.showOnlyOnMobileDevices',
-                        false,
-                    );
+            <LoadingOverlay isLoading={Boolean(isInitializingPayment)}>
+                <Checklist
+                    defaultSelectedItemId={values.paymentProviderRadio}
+                    name="paymentProviderRadio"
+                    onSelect={handleSelect}
+                >
+                    {methods.map((method) => {
+                        const value = getUniquePaymentMethodId(method.id, method.gateway);
+                        const showOnlyOnMobileDevices = get(
+                            method,
+                            'initializationData.showOnlyOnMobileDevices',
+                            false,
+                        );
 
-                    if (showOnlyOnMobileDevices && !isMobile()) {
-                        return;
-                    }
+                        if (showOnlyOnMobileDevices && !isMobile()) {
+                            return;
+                        }
 
-                    return (
-                        <PaymentMethodListItem
-                            disabledReason={
-                                method === chequeMethod ? chequeDisabledReason : undefined
-                            }
-                            isDisabled={isInitializingPayment}
-                            isEmbedded={isEmbedded}
-                            isUsingMultiShipping={isUsingMultiShipping}
-                            key={value}
-                            method={method}
-                            onUnhandledError={onUnhandledError}
-                            value={value}
-                        />
-                    );
-                })}
-            </Checklist>
+                        return (
+                            <PaymentMethodListItem
+                                disabledReason={
+                                    method === chequeMethod ? chequeDisabledReason : undefined
+                                }
+                                isEmbedded={isEmbedded}
+                                isUsingMultiShipping={isUsingMultiShipping}
+                                key={value}
+                                method={method}
+                                onUnhandledError={onUnhandledError}
+                                value={value}
+                            />
+                        );
+                    })}
+                </Checklist>
+            </LoadingOverlay>
         </>
     );
 };
 
 interface PaymentMethodListItemProps {
     disabledReason?: PoDisabledReason;
-    isDisabled?: boolean;
     isEmbedded?: boolean;
     isUsingMultiShipping?: boolean;
     method: PaymentMethod;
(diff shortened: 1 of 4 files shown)


packages/core/src/app/payment/paymentMethod/PaymentMethodList.tsx
```diff
@@ -87,48 +87,49 @@ const PaymentMethodList: FunctionComponent<
             <div aria-live="assertive" className="is-srOnly" role="status">
                 {titleText}
             </div>
-            <Checklist
-                defaultSelectedItemId={values.paymentProviderRadio}
-                isDisabled={isInitializingPayment}
```
bc-maxy, line 92: 💅  Can we remove this isDisabled prop from Checklist component now that we don't pass it anymore or is it still in use in some other place? 👀

bc-charlesho: Checklist's isDisabled={isInitializingPayment} doesn't exist anywhere in the code now.

However, the ChecklistItem's isDisabled={Boolean(disabledReason)} still exists. But its a different prop, on a different component, tied to a different value (disabledReason).

packages/ui/src/form/ChecklistItem/ChecklistItem.tsx
```diff
@@ -18,12 +18,15 @@ export interface ChecklistItemProps {
     content?: ReactNode;
     htmlId?: string;
     isDisabled?: boolean;
+    // Unlike isDisabled, keeps the input focusable while blocking selection.
```
bc-peng:
```suggestion
```

packages/core/src/app/payment/paymentMethod/PaymentMethodList.tsx
```diff
@@ -87,48 +87,49 @@ const PaymentMethodList: FunctionComponent<
...
-                            isUsingMultiShipping={isUsingMultiShipping}
-                            key={value}
-                            method={method}
-                            onUnhandledError={onUnhandledError}
-                            value={value}
-                        />
-                    );
-                })}
-            </Checklist>
+            <LoadingOverlay isLoading={Boolean(isInitializingPayment)}>
```
bc-peng, line 90: There is another LoadingOverlay right above `<PaymentMethodList />` in `packages/core/src/app/payment/PaymentForm.tsx`.

Could we confirm double layer of loading overlays work fine in V1 as well?

packages/core/src/app/payment/paymentMethod/PaymentMethodList.tsx
bc-maxy: I'm reading this as.. safe to remove the prop then? 😄

packages/core/src/app/payment/paymentMethod/PaymentMethodList.tsx
bc-charlesho: Peng, yes I confirm it is working in V1.
https://github.com/user-attachments/assets/b92391e8-e5ce-4d7b-9847-4b386e1cc1cc

packages/ui/src/form/ChecklistItem/ChecklistItem.tsx
bc-charlesho: I'll remove comment and commit the change.

packages/core/src/app/payment/paymentMethod/PaymentMethodList.tsx
bc-charlesho: Its already been removed from the Checklist component.

bc-maxy: Had a look at this and this is fine - we can move the cleanup to another PR

bc-maxy (dismissed):

bc-maxy (approved):
LGTM - great work 👏

bc-launchbay:
success [20260810030200](https://launchbay.bigcommerce.net/projects/535/releases/294372) by Launchbay

bc-launchbay:
success [20260810030200](https://launchbay.bigcommerce.net/projects/535/releases/294372) by Launchbay

bc-launchbay:
success [20260810030200](https://launchbay.bigcommerce.net/projects/535/releases/294372) promoted by https://launchbay.bigcommerce.net/projects/535/releases/294582 being deployed. Deployed by animesh1987

PR #3253, opened 2026-08-17
fix(checkout): CHECKOUT-10283 Apply rounded borders to modals in enhanced theme

bc-maxy:
## What/Why?

For enhancedThemeV1, apply the same 12px rounded borders to modals like for all other elements.

## Rollout/Rollback

Revert the PR.

## Testing

Before:
Screenshot 2026-08-17 at 12 33 35 PM

After:
Screenshot 2026-08-17 at 12 20 08 PM

---

> [!NOTE]
> **Low Risk**
> Visual-only theme styling on modals with no auth, data, or checkout logic changes.
> 
> **Overview**
> **enhancedThemeV1** checkout modals now get the same **12px** corner radius as buttons, alert boxes, and other themed UI.
> 
> `Modal` reads `ThemeContext` and adds the `enhancedThemeV1` class on the dialog when the flag is on, which activates the existing SCSS rule (`&.modal` in `_common.scss`). Tests cover class presence when the theme flag is true vs false.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dd6a816b684222f668ace6080788a99ca9d920ec. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/scss/enhancedThemeV1/common/_common.scss
+++ b/packages/core/src/scss/enhancedThemeV1/common/_common.scss
@@ -8,6 +8,10 @@ $element-border-radius: remCalc(12px);
         border-radius: $element-border-radius;
     }
 
+    &.modal {
+        border-radius: $element-border-radius;
+    }
+
     // Optimized Checkout fallback styles
     background-color: $color-background;
 
--- a/packages/core/src/scss/enhancedThemeV1/common/_common.scss
+++ b/packages/core/src/scss/enhancedThemeV1/common/_common.scss
@@ -4,10 +4,7 @@ $element-border-radius: remCalc(12px);
     .button,
     .shippingOptions-panel,
     .alertBox,
-    .applied-coupons-list ul {
-        border-radius: $element-border-radius;
-    }
-
+    .applied-coupons-list ul,
     &.modal {
         border-radius: $element-border-radius;
     }
--- a/packages/core/src/scss/enhancedThemeV1/components/checkout/_modal.scss
+++ b/packages/core/src/scss/enhancedThemeV1/components/checkout/_modal.scss
@@ -1,0 +1,5 @@
+@include enhancedThemeV1 {
+    &.modal {
+        border-radius: $element-border-radius;
+    }
+}
--- a/packages/core/src/scss/enhancedThemeV1/components/checkout/_modal.scss
+++ b/packages/core/src/scss/enhancedThemeV1/components/checkout/_modal.scss
@@ -1,5 +1,0 @@
-@include enhancedThemeV1 {
-    &.modal {
-        border-radius: $element-border-radius;
-    }
-}
--- a/packages/ui/src/modal/Modal.tsx
+++ b/packages/ui/src/modal/Modal.tsx
@@ -42,9 +42,7 @@ const Modal: FunctionComponent<ModalProps> = ({
     shouldShowCloseButton = false,
     ...rest
 }) => {
-    const themeContext = useContext(ThemeContext);
-
-    const isEnhancedThemeV1 = themeContext?.enhancedThemeV1;
+    const { enhancedThemeV1 } = useContext(ThemeContext) ?? {};
 
     const handleClose = useCallback(
         (event: MouseEvent | KeyboardEvent) => {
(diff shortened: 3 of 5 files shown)


bc-maxy:
cursor review

packages/core/src/scss/enhancedThemeV1/components/checkout/_modal.scss
```diff
@@ -0,0 +1,5 @@
+@include enhancedThemeV1 {
+    &.modal {
+        border-radius: $element-border-radius;
```
richapatel1510: since its only the border radius for modal, can we move this to common.scss as well? Are we expecting any more styles that might be added for modal?

bc-maxy: Good point, thank you! I moved it to common

packages/core/src/scss/enhancedThemeV1/common/_common.scss
```diff
@@ -8,6 +8,10 @@ $element-border-radius: remCalc(12px);
         border-radius: $element-border-radius;
     }
 
+    &.modal {
```
richapatel1510:
```suggestion
.applied-coupons-list ul,
.modal {
```

richapatel1510: Can include it along with the above list?

richapatel1510 (dismissed):

packages/ui/src/modal/Modal.tsx
```diff
@@ -40,6 +42,10 @@ const Modal: FunctionComponent<ModalProps> = ({
     shouldShowCloseButton = false,
     ...rest
 }) => {
+    const themeContext = useContext(ThemeContext);
+
+    const isEnhancedThemeV1 = themeContext?.enhancedThemeV1;
```
bc-peng: Thought we may avoid `isEnhancedThemeV1` by doing `const { enhancedThemeV1 } = useContext(ThemeContext) ?? {};`, then use it as 
```
classNames(
    'modal optimizedCheckout-contentPrimary',
    { enhancedThemeV1 },
    additionalModalClassName,
),
```
What do you think?

bc-maxy: Yeah I guess it does reduce number of lines, thanks, adjusted:)

bc-peng (approved):

animesh1987 (approved):

packages/ui/src/modal/Modal.tsx
richapatel1510: @bc-maxy Is there a circular dependency in using `const { enhancedThemeV1 } = useThemeContext();`?

bc-peng: Thought it should be a one-way reading from `useThemeContext()`. 

Could you tell us more?

richapatel1510: I mean why are we not using `useThemeContext()` directly here? Why do we use `useContext(ThemeContext)`? Is there a circular dependency issue in using that?

bc-maxy: No, I don't think there is a circular dependency in using useThemeContext. The reason I'm going for useContext(ThemeContext) instead is because this way it doesn't throw when the modal is used outside of ThemeContextProvider which has this:

```
throw new Error('useThemeContext must be used within an <ThemeContextProvider>');
```

If we use useContext(ThemeContext) ?? {}, then when we don't have this context provider it just resolves to no enhancedThemeV1 class instead of breaking. I think this is a better approach since this is a shared ui component and theme context is app specific. Let me know if you agree

richapatel1510: Makes sense. I just brought it up because we use `useThemeContext` everywhere else hence a bit different approach here.

bc-maxy: Definitely, thank you for checking and flagging this 🙌

bc-launchbay:
success [20260818041744](https://launchbay.bigcommerce.net/projects/535/releases/295361) by Launchbay

bc-launchbay:
success [20260818041744](https://launchbay.bigcommerce.net/projects/535/releases/295361) by Launchbay

bc-launchbay:
success [20260818041744](https://launchbay.bigcommerce.net/projects/535/releases/295361) promoted by https://launchbay.bigcommerce.net/projects/535/releases/295383 being deployed. Deployed by Launchbay

PR #3254, opened 2026-08-17
feat(checkout): CHECKOUT-10303 Fix `Same as shipping address` checkbox to be accurate

richapatel1510:
## What/Why?

- Fix `Same as shipping address` or `My billing address is same as shipping` checkbox state to honour the latest stae and not only the default checkout setting.
- If the addresses are not same on the stored checkout object the checkbox should not be enabled even if the default setting is set to `true`.

## Rollout/Rollback

Revert this PR.

## Testing

- CI checks
- Manual testing

**BEFORE**
- Existing theme 

https://github.com/user-attachments/assets/72e7bd49-cced-4a7c-979f-206499cd22ef

- New theme 

https://github.com/user-attachments/assets/a4625e13-a987-40a8-a2b5-1886697a5d93

**AFTER**
- Existing theme 

https://github.com/user-attachments/assets/516900fb-0593-408a-8cd6-e75280bcddc6

- New theme 

https://github.com/user-attachments/assets/94ecea0c-6b49-44da-8727-35a1eee927c4

- B2B - QuoteToCheckout

https://github.com/user-attachments/assets/60db9ace-14ac-4d1b-ad10-f34c86224653

---

> [!NOTE]
> **Medium Risk**
> Changes checkout billing/shipping state derivation on load and after billing save; incorrect logic could mislead shoppers about billing address, but scope is limited and covered by new tests.
> 
> **Overview**
> Fixes the **“My billing address is the same as my shipping address”** checkbox so it reflects persisted billing/shipping data, not only the store default `checkoutBillingSameAsShippingEnabled`.
> 
> On checkout load, `CheckoutPage` now seeds `isBillingSameAsShipping` via **`getInitialBillingSameAsShipping`**, which falls back to the store setting when shipping is missing or billing is “empty” (new **`isEmptyAddress`**, e.g. guest billing with only email), otherwise uses **`isEqualAddress`**.
> 
> After the billing step saves (no checkbox there), **`handleBillingNextStep`** re-reads addresses from the checkout service and updates the flag before advancing, so returning to shipping shows the correct checked state.
> 
> Exports **`normalizeAddress`** / **`ComparableAddress`** from `isEqualAddress` for shared comparison logic. Adds unit and Checkout integration tests; restores jest-dom matchers shadowed by Playwright in `Checkout.test.tsx`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit da9b8a7bc64866cb108fd87ae9c022ab305228ae. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/app/address/isEmptyAddress.ts
+++ b/packages/core/src/app/address/isEmptyAddress.ts
@@ -1,0 +1,23 @@
+import { type ComparableAddress, normalizeAddress } from './isEqualAddress';
+
+// Empty = no shopper-entered content, e.g. a guest billing address holding only an email.
+export default function isEmptyAddress(address?: ComparableAddress): boolean {
+    if (!address) {
+        return true;
+    }
+
+    const {
+        customFields = [],
+        extraFields = [],
+        countryCode: _countryCode,
+        ...fields
+    } = normalizeAddress(address);
+
+    return (
+        Object.values(fields).every((value) => !value) &&
+        customFields.length === 0 &&
+        extraFields.length === 0 &&
+        !address.stateOrProvince &&
+        !address.stateOrProvinceCode
+    );
+}
--- a/packages/core/src/app/address/isEmptyAddress.ts
+++ b/packages/core/src/app/address/isEmptyAddress.ts
@@ -1,7 +1,7 @@
 import { type ComparableAddress, normalizeAddress } from './isEqualAddress';
 
 // Empty = no shopper-entered content, e.g. a guest billing address holding only an email.
-export default function isEmptyAddress(address?: ComparableAddress): boolean {
+export function isEmptyAddress(address?: ComparableAddress): boolean {
     if (!address) {
         return true;
     }
--- a/packages/core/src/app/checkout/Checkout.test.tsx
+++ b/packages/core/src/app/checkout/Checkout.test.tsx
@@ -28,6 +28,7 @@ import {
     CheckoutPreset,
     checkoutSettings,
     checkoutWithBillingEmail,
+    checkoutWithShipping,
     checkoutWithShippingAndBilling,
     checkoutWithShippingDiscount,
     consignmentAutomaticDiscount,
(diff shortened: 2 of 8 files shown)


richapatel1510:
cursor review

packages/core/src/app/address/isEmptyAddress.ts
```diff
@@ -0,0 +1,23 @@
+import { type ComparableAddress, normalizeAddress } from './isEqualAddress';
+
+// Empty = no shopper-entered content, e.g. a guest billing address holding only an email.
+export default function isEmptyAddress(address?: ComparableAddress): boolean {
```
bc-maxy: nit

```suggestion
export function isEmptyAddress(address?: ComparableAddress): boolean {
```

packages/core/src/app/checkout/getInitialBillingSameAsShipping.ts
```diff
@@ -0,0 +1,22 @@
...
+import { isEmptyAddress, isEqualAddress } from '../address';
+
+export interface GetInitialBillingSameAsShippingOptions {
+    billingAddress?: BillingAddress;
+    shippingAddress?: Address;
+    defaultValue: boolean;
+}
+
+// Seed from the persisted addresses when they exist; fresh checkouts fall back to the store setting.
+export default function getInitialBillingSameAsShipping({
```
bc-maxy:
```suggestion
export function getInitialBillingSameAsShipping({
```

packages/core/src/app/checkout/getInitialBillingSameAsShipping.ts
```diff
@@ -0,0 +1,22 @@
+import { type Address, type BillingAddress } from '@bigcommerce/checkout-sdk/essential';
+
+import { isEmptyAddress, isEqualAddress } from '../address';
+
+export interface GetInitialBillingSameAsShippingOptions {
```
bc-maxy:
```suggestion
interface GetInitialBillingSameAsShippingOptions {
```

packages/core/src/app/checkout/Checkout.test.tsx
```diff
@@ -719,6 +723,146 @@ describe('Checkout', () => {
...
+            await checkout.waitForPaymentStep();
+
+            await userEvent.click(screen.getAllByRole('button', { name: 'Edit' })[1]);
+            await checkout.waitForShippingStep();
+
+            expect(
+                screen
+                    .queryByLabelText('My billing address is the same as my shipping address.')
+                    ?.hasAttribute('checked'),
+            ).toBeFalsy();
```
bc-maxy: Would something like this work instead? (also for all other instances)

```suggestion
            expect(
                screen
                    .getByLabelText('My billing address is the same as my shipping address.')
                    .not.toBeChecked();
```

bc-maxy (commented):
minor comments, looks good overall to me!

packages/core/src/app/address/isEmptyAddress.ts
animesh1987: Maybe we need to add a skill that enforces this? cc @bc-maxy

packages/core/src/app/checkout/CheckoutPage.tsx
```diff
@@ -444,6 +445,24 @@ const Checkout = ({
         [],
     );
 
+    // The billing step has no same-as-shipping checkbox, so re-derive the flag
+    // from the just-saved addresses; read them at call time as the props
+    // captured before the billing update are stale.
+    const handleBillingNextStep = useCallback((): void => {
```
animesh1987, line 451: Is this in the existing flow?

packages/core/src/app/address/isEmptyAddress.ts
bc-maxy: Yep I'll investigate 😄

bc-maxy: also for useCheckout

packages/core/src/app/checkout/CheckoutPage.tsx
richapatel1510: yes, its for existing flow where the checkbox value is derived incorrectly after billing step is edited separately.

packages/core/src/app/checkout/Checkout.test.tsx
richapatel1510: Done ✅ 
One catch: it didn't work out of the box — the `@bigcommerce/checkout/test-framework` import loads Playwright's `expect`, which shadows jest-dom's `toBeChecked` in the shared matcher registry (that's why the old code checked `hasAttribute('checked')` with an eslint-disable). Added a one-line `expect.extend(jestDomMatchers)` after the imports to restore them, then switched all instances to `toBeChecked() / not.toBeChecked()`.

bc-maxy: Oh ok I understand now why it was done this way. That sounds like something we should fix - like moving Playwright fixtures to a separate test-framework export? 🤔 But definitely not in this PR

bc-maxy (approved):

packages/core/src/app/checkout/Checkout.test.tsx
```diff
@@ -43,6 +48,10 @@ import { getCountries } from '../geography/countries.mock';
 
 import Checkout, { type CheckoutProps } from './Checkout';
 
+// The test-framework import above loads Playwright's expect, whose Locator-only matchers
+// (e.g. toBeChecked) shadow jest-dom's in the shared matcher registry; restore jest-dom's.
+expect.extend(jestDomMatchers);
```
bc-peng, line 53: Would it be more beneficial to put it in `jest-setup.ts`?

bc-peng (approved):

packages/core/src/app/checkout/Checkout.test.tsx
richapatel1510: Yes, it would be beneficial. But the ideal fix is to fix the root cause: the `@bigcommerce/checkout/test-framework` barrel re-exports Playwright's expect, which leaks its Locator-only matchers into Jest's registry. I can create a separate PR to split the Jest-safe exports from the Playwright ones, after which this `expect.extend` line can be deleted altogether — keeping it file-local here so it's easy to find and remove.

bc-launchbay:
success [20260820031121](https://launchbay.bigcommerce.net/projects/535/releases/295745) by Launchbay

bc-launchbay:
success [20260820031121](https://launchbay.bigcommerce.net/projects/535/releases/295745) by Launchbay

PR #3270, opened 2026-08-21
fix(checkout): CHECKOUT-10193 Add retry to checkout-button-resolver

bc-maxy:
## What/Why?

Shoppers on flaky networks who fail to fetch the wallet-button-v1-resolver chunk now get up to 5 automatic retries, and if all fail, instead of throwing and making the whole checkout page into white-screen failure, wrap wallet buttons into a LazyContainer and stop it there so that the rest of checkout is unaffected. 

No visual change on the happy path.

Worst case adds ~4s of retry delay, which is the intended trade-off.

## Rollout/Rollback

Revert the PR.

## Testing

New CI tests + manual.

Before:

https://github.com/user-attachments/assets/a007304b-8443-4612-935e-9f1f5d666c70

After:

https://github.com/user-attachments/assets/1742dba1-ea6e-4d6d-a3f9-7685d570ba2c

---

> [!NOTE]
> **Low Risk**
> Resilience-only changes around lazy-loaded wallet UI; payment and auth paths are untouched, with graceful degradation on chunk failure.
> 
> **Overview**
> Wallet checkout buttons now **retry** loading the `wallet-button-v1-resolver` code-split chunk (via shared `retry`, up to 5 attempts with backoff) in both top-of-checkout **`CheckoutButtonContainer`** and **`CheckoutButtonList`**, instead of failing on the first network blip.
> 
> **`CheckoutButtonContainer`** also wraps the remote/wallet button block in **`LazyContainer`**, so a **`ChunkLoadError`** after retries surfaces the existing unstable-network copy instead of breaking the whole checkout. Happy-path rendering is unchanged.
> 
> CI adds coverage that simulates a failed chunk load and asserts **`retry`** runs and the network error message appears.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a409646fea172a970686b99c020b5eb869c837b4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


bc-maxy:
cursor review

bc-maxy:
cursor review

bc-maxy:
cursor review

packages/error-handling-utils/src/ErrorBoundary.tsx
```diff
@@ -22,12 +23,16 @@ class ErrorBoundary extends Component<ErrorBoundaryProps, ErrorBoundaryState> {
...
-        const { filter = () => true, errorLogger } = this.props;
+        const { filter = () => true, errorLogger, onError } = this.props;
 
         if (!filter(error)) {
             throw error;
         }
 
+        if (onError) {
+            onError(error);
+        }
```
bc-peng: Do we have other options to retry the resolver besides touching `ErrorBoundary`?

bc-maxy: The error boundary part is I guess a separate issue from retries. What bugbot mentioned above is that without onError callback isLoading never clears and the skeleton keeps all buttons hidden from the screen (cause we apply top: -300%).

Do you think I should move this change to another PR and just keep retries in this one?

bc-peng: I didn't read Cursor review😂 Just tried to understand the reasoning of changing `ErrorBoundary` because `ErrorBoundary` is bit legacy.

bc-maxy: Ah I see 😄  Well that's the only error boundary I saw used in the repo, do we have an alternative? I thought React requires it to be a class, but maybe that's changed now 😄

bc-peng: I see we need to update ErrorBoundary to handle a loading error with:
```
   const handleChunkLoadError = () => {
        setHasChunkLoadError(true);
    };
```
Is it possible that we do `isLoading={isLoading && !hasChunkLoadError}` in the other way, such as only hiding it when all wallet buttons are properly loaded?

bc-maxy: Good point! What do you think about this change: [9ecac59](https://github.com/bigcommerce/checkout-js/pull/3270/commits/9ecac59ba0b01bce8916f5032e1cee78505d1c82) ?

bc-peng: `methodId` could be a dangerous prop. Let’s discuss this offline.

packages/core/src/app/customer/CheckoutButtonContainer.tsx
```diff
@@ -110,22 +114,24 @@ const CheckoutButtonContainer: FunctionComponent<
...
+                    >
+                        <div className="checkoutRemote">{renderButtons()}</div>
+                    </WalletButtonsContainerSkeleton>
+                </div>
+                <div className="checkout-separator">
+                    <span className="optimizedCheckout-headingSecondary sub-header">
+                        <TranslatedString id="remote.or_text" />
+                    </span>
+                </div>
+            </LazyContainer>
```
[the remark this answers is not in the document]
bc-maxy: Is this an acceptable tradeoff for now?

We don't track isLoading per wallet button. Wrapping each button in its own LazyContainer prevents the crash from the "Before" video, but because the failed button never resolves, the skeleton stays in a permanent loading state:

https://github.com/user-attachments/assets/f49790f9-5fa5-4adc-a676-85b90d61352a

Fixing that properly means reworking the component's loading logic, which is a much bigger change. Since this failure is an edge case we only found through error logs (no customer reports from what I know), I'd rather ship this as-is and handle the loading-state rework in a follow-up PR if needed. Curious to hear your thoughts though, let me know.

packages/error-handling-utils/src/ErrorBoundary.tsx
bc-maxy: Reverted to a simpler implementation after our discussion

bc-peng (approved):

animesh1987 (approved):

bc-launchbay:
success [20260827051608](https://launchbay.bigcommerce.net/projects/535/releases/296580) by Launchbay

bc-launchbay:
success [20260827051608](https://launchbay.bigcommerce.net/projects/535/releases/296580) by Launchbay

bc-launchbay:
success [20260827051608](https://launchbay.bigcommerce.net/projects/535/releases/296580) promoted by https://launchbay.bigcommerce.net/projects/535/releases/296746 being deployed. Deployed by Launchbay

PR #3280, opened 2026-08-26
CHECKOUT-9630 Remove no countries error on checkout experiment

bc-mackxu:
## What/Why?

Removes the rolled-out `CHECKOUT-9630.no_countries_error_on_checkout` LaunchDarkly experiment and its dead code path. The experiment's fallback value was already `true` (enabled), so the "on" behavior — logging an `no_countries_available` unhandled error and rendering nothing when the shipping step has no countries available — becomes the permanent, unconditional behavior in [ShippingForm.tsx](packages/core/src/app/shipping/ShippingForm.tsx).

Changes:
- [useShipping.ts](packages/core/src/app/shipping/hooks/useShipping.ts): removed `isNoCountriesErrorOnCheckoutEnabled` (and the now-unused `isExperimentEnabled` import).
- [ShippingForm.tsx](packages/core/src/app/shipping/ShippingForm.tsx): removed the experiment guard from both the error-reporting effect and the early `return null`.
- [useShipping.mock.ts](packages/core/src/app/shipping/hooks/useShipping.mock.ts): removed the corresponding mock field.
- [Shipping.test.tsx](packages/core/src/app/shipping/Shipping.test.tsx): dropped the "experiment disabled" test and the experiment-flag setup from the "experiment enabled" test, since the behavior is now unconditional.

## Rollout/Rollback

No rollout needed — the experiment was already defaulting to enabled, so this only removes the flag check; runtime behavior is unchanged. Rollback is a straight revert of this PR.

Note: the LaunchDarkly Terraform module (`bigcommerce/terraform-launchdarkly`) and the backend constant (`bigcommerce/bigcommerce: config/Experiments/CheckoutExperiments.php`) still reference this flag and are intentionally left in place — those should be removed in a follow-up PR after this one merges, per team convention (keep the flag alive in LD while the code cleanup is in flight).

## Testing

Ran the updated `Shipping.test.tsx` suite locally — all 20 tests pass, including the simplified "No countries available error handling" test. `tsc --noEmit` shows no new type errors versus the pre-existing baseline.

### DIFF
--- a/packages/core/src/app/shipping/Shipping.test.tsx
+++ b/packages/core/src/app/shipping/Shipping.test.tsx
@@ -980,21 +980,8 @@ describe('Shipping step', () => {
     });
 
     describe('No countries available error handling', () => {
-        it('calls onUnhandledError when no countries are available and experiment is enabled', async () => {
-            const config = {
-                ...checkoutSettings,
-                storeConfig: {
-                    ...checkoutSettings.storeConfig,
-                    checkoutSettings: {
-                        ...checkoutSettings.storeConfig.checkoutSettings,
-                        features: {
-                            'CHECKOUT-9630.no_countries_error_on_checkout': true,
-                        },
-                    },
-                },
-            };
-
-            checkoutService = checkout.use(CheckoutPreset.CheckoutWithBillingEmail, { config });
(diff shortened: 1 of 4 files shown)


bc-mackxu:
Closed PR, one store still off

PR #3295, opened 2026-09-01
feat(checkout): CHECKOUT-10366 allow address fields to be reordered for enhancedTheme

richapatel1510:
## What/Why?

- We had fixed the order of built in fields for enhancedThemeV1 in this [PR](https://github.com/bigcommerce/checkout-js/pull/2813/) where we lost the ability that we previously allowed fields to be reordered from the CP settings. 
- Bring back that capability to allow reordering fields on CP settings and mantaining that order on checkout UI with just one change, which is keeping the Country field at top for enhancedThemeV1.
- Also the previous change of fixed order was done only for SingleShipping form, but we need to keep it consistent across all the address forms so changes are directly done to AddressForm.

## Rollout/Rollback

Revert this PR.

## Testing

- CI checks
- Manual testing
**BEFORE**

https://github.com/user-attachments/assets/94fa948a-37f4-411d-b391-58a978312e9b

**AFTER**

https://github.com/user-attachments/assets/579203db-5c65-455b-ac2b-f5949acfc013

---

> [!NOTE]
> **Low Risk**
> Checkout address layout only; no auth or payment logic, with behavior gated to enhanced theme and covered by new tests.
> 
> **Overview**
> **Restores Control Panel–configured address field order** for enhanced theme v1, replacing the prior full fixed-field reorder that blocked merchant customization.
> 
> When `enhancedThemeV1` is on, **`AddressForm`** now only runs `moveCountryFieldToTop` so **`countryCode` stays first** and every other field keeps the order from `formFields`. Without enhanced theme, fields render unchanged. **`reorderAddressFormFields`** and its export are removed; **`ShippingAddressForm`** no longer applies theme-specific sorting and passes `formFields` through to **`AddressForm`**.
> 
> Tests wrap renders in **`ThemeContext`** / **`ThemeProvider`** and add coverage for enhanced vs default field order plus unit tests for **`moveCountryFieldToTop`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7b8d01ba165757edd8d0e8a119681a32d279ec91. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


richapatel1510:
cursor review

packages/core/src/app/address/moveCountryFieldToTop.ts
```diff
@@ -0,0 +1,16 @@
+import { type FormField } from '@bigcommerce/checkout-sdk';
+
+const COUNTRY_FIELD_NAME = 'countryCode';
+
+export function moveCountryFieldToTop(formFields: FormField[]): FormField[] {
+    const countryFieldIndex = formFields.findIndex((field) => field.name === COUNTRY_FIELD_NAME);
+
+    if (countryFieldIndex <= 0) {
```
animesh1987, line 8: Do we just need to check for `countryFieldIndex === 0` here? Since not sure it can be less than 0?

packages/core/src/app/shipping/ShippingAddressForm.tsx
```diff
@@ -117,7 +113,7 @@ const ShippingAddressForm = ({
                     <AddressForm
                         countryCode={formAddress && formAddress.countryCode}
                         fieldName={addressFieldName}
-                        formFields={sortedFormFields}
+                        formFields={formFields}
```
animesh1987, line 116: We don't need to reorder here?

packages/core/src/app/address/moveCountryFieldToTop.ts
richapatel1510: `findIndex` can return `-1` if the country field is not present in the `formFields` array, which will not ideally be the case but we have to keep this condition to make TS happy. Hence, if no country field present we return `formFields` as is, otherwise move country field to top.

packages/core/src/app/shipping/ShippingAddressForm.tsx
richapatel1510: No, all the reorder happen in AddressForm now, no need of individual reorders.

packages/core/src/app/address/AddressForm.tsx
```diff
@@ -77,6 +84,8 @@ const AddressForm: React.FC<AddressFormProps> = ({
         false,
     );
     const countriesWithAutocomplete = ['US', 'CA', 'AU', 'NZ', 'GB'];
+    // Field order comes from the API; the enhanced theme only pins the country field first
```
bc-maxy:
```suggestion
```
nit but I think the code is quite clear already

bc-maxy (dismissed):
Looks good to me!

bc-maxy (approved):

animesh1987 (approved):

bc-peng (approved):

bc-launchbay:
success [20260903015158](https://launchbay.bigcommerce.net/projects/535/releases/297446) by Launchbay

bc-launchbay:
success [20260903015158](https://launchbay.bigcommerce.net/projects/535/releases/297446) by Launchbay

bc-launchbay:
success [20260903015158](https://launchbay.bigcommerce.net/projects/535/releases/297446) promoted by https://launchbay.bigcommerce.net/projects/535/releases/297450 being deployed. Deployed by Launchbay

PR #3324, opened 2026-09-08
feat(checkout): CHECKOUT-10399 Redesign payment loaders

bc-maxy:
## What/Why?

A small re-design of payment methods loaders only impacting enhancedThemeV1.

Right now we render a skeleton when loading payment methods and then show a spinner over all payment methods while they initialize. 

With this PR we remove the spinner and instead keep the same skeleton (for consistency) and show it inside of payment methods. This makes more sense because we are actually waiting for a method to initialize so it makes sense to scope the loader to a payment method.

## Rollout/Rollback

Revert the PR.

## Testing

### Before

https://github.com/user-attachments/assets/ed137e66-a687-45b5-8920-05685a02700b

### After

https://github.com/user-attachments/assets/a4f4422c-b585-4faa-a0be-801e6e1f7b6d

---

> [!NOTE]
> **Low Risk**
> Checkout payment UI only, gated behind enhancedThemeV1 with prior loader behavior preserved for the default theme; no auth or payment-processing logic changes.
> 
> **Overview**
> For **enhancedThemeV1**, payment initialization no longer shows a **full payment-method list** `LoadingOverlay` spinner. The list stays visible; only the **selected** method shows a scoped loader.
> 
> While `isInitializingPayment` is true, the selected (non–custom-checklist) method wraps its content in `LoadingOverlay` with `hideContentWhenLoading` and a new **`PaymentMethodSkeleton`** (checklist-style pulse animation). Other methods remain visible but **read-only** (selection blocked; checklist radios styled as non-interactive). **Legacy theme** behavior is unchanged—the entire checklist still uses the list-level overlay.
> 
> Adds **`PaymentMethodSkeleton`** in `@bigcommerce/checkout/ui`, enhanced-theme SCSS for the skeleton and disabled payment radios, and refactors skeleton timing into **`$loading-skeleton-grey-box-timing`**. **`PaymentForm`** tests cover the enhanced-theme loader behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e400e2136c232a8600da5ce0388629ae59a6e522. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/app/payment/paymentMethod/PaymentMethodList.tsx
+++ b/packages/core/src/app/payment/paymentMethod/PaymentMethodList.tsx
@@ -7,7 +7,6 @@ import {
     Checklist,
     ChecklistItem,
     LoadingOverlay,
-    LoadingSkeletonContext,
     PaymentMethodSkeleton,
 } from '@bigcommerce/checkout/ui';
 
@@ -121,13 +120,9 @@ const PaymentMethodList: FunctionComponent<
                 {titleText}
             </div>
             {enhancedThemeV1 ? (
-                // enhancedThemeV1 keeps the method list visible while a method
-                // initializes: no spinner veil, and the initializing method's
-                // fields render as a skeleton (items stay read-only via
-                // isReadOnly until initialization finishes).
-                <LoadingSkeletonContext.Provider value={paymentMethodSkeleton}>
-                    {checklist}
-                </LoadingSkeletonContext.Provider>
+                // enhancedThemeV1 keeps the method list visible while a method initializes:
+                // no spinner veil. The skeleton is applied per-item in PaymentMethodListItem.
+                checklist
             ) : (
                 <LoadingOverlay isLoading={Boolean(isInitializingPayment)}>
                     {checklist}
@@ -168,13 +163,14 @@ const PaymentMethodListItem: FunctionComponent<PaymentMethodListItemProps> = ({
             />
         );
 
-        // enhancedThemeV1 has no spinner veil over the list, so stand in a
-        // skeleton (from LoadingSkeletonContext) for the selected method's
-        // fields while it initializes. Doing it here works for every
-        // integration, whether or not it forwards isInitializing to its own
-        // LoadingOverlay.
+        // enhancedThemeV1 has no spinner veil over the list, so stand in a skeleton for the
+        // selected method's fields while it initializes.
         return enhancedThemeV1 ? (
-            <LoadingOverlay hideContentWhenLoading isLoading={Boolean(isInitializingPayment)}>
+            <LoadingOverlay
+                hideContentWhenLoading
+                isLoading={Boolean(isInitializingPayment)}
+                loadingSkeleton={paymentMethodSkeleton}
+            >
                 {paymentMethod}
             </LoadingOverlay>
         ) : (
--- a/packages/core/src/app/payment/paymentMethod/PaymentMethodList.tsx
+++ b/packages/core/src/app/payment/paymentMethod/PaymentMethodList.tsx
@@ -101,6 +101,7 @@ const PaymentMethodList: FunctionComponent<
                         disabledReason={method === chequeMethod ? chequeDisabledReason : undefined}
                         isEmbedded={isEmbedded}
                         isInitializingPayment={isInitializingPayment}
+                        isSelected={value === values.paymentProviderRadio}
                         isUsingMultiShipping={isUsingMultiShipping}
                         key={value}
                         method={method}
@@ -132,6 +133,7 @@ interface PaymentMethodListItemProps {
     disabledReason?: PoDisabledReason;
     isEmbedded?: boolean;
     isInitializingPayment?: boolean;
+    isSelected: boolean;
     isUsingMultiShipping?: boolean;
     method: PaymentMethod;
     value: string;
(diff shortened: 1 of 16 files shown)


bc-maxy:
Originally I had something like this which matches the card payment method, but then I realized it doesn't match non-card methods and we need something generic instead...

https://github.com/user-attachments/assets/5e0bac8f-e437-427a-b102-55de6d575ba3

bc-maxy:
cursor review

bc-maxy:
cursor review

packages/core/src/app/payment/paymentMethod/PaymentMethodList.tsx
```diff
@@ -130,21 +144,47 @@ const PaymentMethodListItem: FunctionComponent<PaymentMethodListItemProps> = ({
...
                 isEmbedded={isEmbedded}
                 isUsingMultiShipping={isUsingMultiShipping}
                 method={method}
                 onUnhandledError={onUnhandledError || noop}
             />
         );
-    }, [isEmbedded, isUsingMultiShipping, method, onUnhandledError]);
+
+        // Custom checklist items manage their own loading UI
+        return enhancedThemeV1 && !isCustomChecklistItem ? (
```
richapatel1510, line 167: Custom checklist items were previously covered by the list-wide overlay and now get nothing. The comment says they manage their own loading UI — is that true for all of them, or just the wallet buttons you checked? If any of them don't, they'll just sit there looking like a dead radio during init.

packages/core/src/scss/enhancedThemeV1/components/checkout/_payment-loading-skeleton.scss
```diff
@@ -0,0 +1,20 @@
...
+        background-color: mix(color("greys", "light"), color("greys", "medium"), 50%);
+    }
+}
+
+.payment-method-skeleton {
+    @include enhancedThemeV1 {
+        padding: spacing("half") 0;
+
+        div {
+            animation: payment-method-skeleton-pulse 4000ms infinite cubic-bezier(0.13, 0.615, 0.315, 0.915);
```
richapatel1510, line 16: 💅 
`4000ms infinite cubic-bezier(0.13, 0.615, 0.315, 0.915)` is character-for-character `$loading-skeleton-grey-box-animation` from `settings/foundation/forms/_settings.scss`. Worth pulling the timing from the variable so the two don't drift apart later — only the colour stops actually differ here.

packages/core/src/app/payment/paymentMethod/PaymentMethodList.tsx
```diff
@@ -130,21 +144,47 @@ const PaymentMethodListItem: FunctionComponent<PaymentMethodListItemProps> = ({
...
+            <LoadingOverlay
+                hideContentWhenLoading
```
richapatel1510, line 169: Heads up that `hideContentWhenLoading` sets `display: none` on the content rather than overlaying it, so the method initializes with no layout box.

No change for `credit-card / hosted-payment / Adyen / Braintree / PayPal Commerce` — they already self-wrap in this with the same flag. But the integrations that don't (`Stripe, CheckoutCom, Mollie, Moneris, Worldpay`) used to render visible under the overlay and now go fully hidden during init. `Stripe UPE` mounts Elements into `#stripe-${method.id}-component-field` right in that window.

Can you try to test Stripe manally? That one plus another hosted-field provider would be good to check.

richapatel1510 (commented):
Couple of questions, rest LGTM! 👏

packages/core/src/app/payment/paymentMethod/PaymentMethodList.tsx
bc-maxy: From my understanding it's only Stripe that uses this CustomChecklistItem. And it does have a loader so it should be okay. Wallets go down the normal ChecklistItem path and do get the new skeleton

packages/core/src/app/payment/paymentMethod/PaymentMethodList.tsx
bc-maxy: I'm not sure how to test Stripe UPE as I think this is a legacy system. Do you know? My understanding is that we can only set up Stripe OCS for new stores.

Tested with Stripe OCS (the first payment result) and CheckoutCom though:

https://github.com/user-attachments/assets/ff2301c5-a7a7-4bcf-9d26-5e37691e3ab1

bc-maxy: I managed to set up UPE and test this 😄 (big thanks to @PavlenkoM )

Seems to work as expected (see below).

### Before

https://github.com/user-attachments/assets/44665491-a169-4757-a043-648e4229a595

### After

https://github.com/user-attachments/assets/c381f8a3-b021-4d67-9995-be3c538c9c96

richapatel1510: Looks good. Thanks for checking 🙇

richapatel1510 (approved):

animesh1987 (approved):

bc-launchbay:
success [20260924001458](https://launchbay.bigcommerce.net/projects/535/releases/300234) by Launchbay

bc-launchbay:
success [20260924001458](https://launchbay.bigcommerce.net/projects/535/releases/300234) by Launchbay

bc-launchbay:
success [20260924001458](https://launchbay.bigcommerce.net/projects/535/releases/300234) by Launchbay

PR #3343, opened 2026-09-15
feat(payment): read storefront shipping options filter for wallet buttons

PavlenkoM:
## What/Why?

Wallet payment sheets are handed every available shipping option. Once the shopper picks
an address inside the sheet, that list is never narrowed — the wallet owns the modal from
that point on. Stores whose relevant options depend on the address, or whose list is long
enough that the wallet truncates it (Stripe caps at 9 rates), show shoppers irrelevant
rates or silently lose relevant ones.

The SDK already accepts an optional `filterAvailableShippingOptions` callback on the Apple
Pay, Google Pay and Stripe Link V2 initialize options (`@bigcommerce/checkout-sdk@1.978.0`,
already pinned here). Nothing in checkout-js was handing it one, so a store with a
customised storefront had to fork checkout-js to filter at the checkout customer step.

A storefront now registers:

```js
window.BigCommerce = window.BigCommerce || {};
window.BigCommerce.walletButtons = {
    filterAvailableShippingOptions: async (shippingOptions, context) =>
        shippingOptions.filter((option) => option.type !== 'shipping_pickupinstore'),
};
```

`CheckoutButton` is the single component every wallet wrapper renders through
(`ApplePayButton`, `GooglePayButton`, `StripeLinkV2Button`, `AmazonPayV2Button`, PayPal,
Bolt), so resolving the callback there covers every wallet at once.

Design points worth reviewing:

- **Resolved lazily, at call time.** The lookup happens when the shopper changes an address
  inside a sheet, not when the button initializes, so registration order cannot race. This
  matters because a storefront may register the callback from a script that runs after the
  buttons have initialized.
- **Absent callback returns the options untouched.** Because resolution is lazy, the wrapper
  is attached on every store, so this is the common path. Returning `undefined` instead
  would throw a `TypeError` out of the Google Pay and Link V2 strategies (both `.map()`
  without optional chaining, outside the surrounding try/catch) and would silently yield
  zero shipping options in Apple Pay. This keeps the change inert for stores that do not
  register a callback.
- **Context is normalized.** `CheckoutButton` only knows a single `methodId`, which for
  Google Pay is the gateway variant (`googlepaystripeupe`). It is normalized to
  `{ methodId: 'googlepay', gatewayId: 'googlepaystripeupe' }` so a callback branching on
  `methodId` behaves identically here and on the other surfaces that pass this callback.
  Without this, `context.methodId === 'googlepay'` would silently stop matching.
- **Passed to every wallet button, with no allow list.** This matches how `CheckoutButton`
  already works: `container`, `onClick` and `onUnhandledError` go to every wallet, and only
  10 of the 52 customer strategies in the SDK read `onClick` at all. Strategies that
  implement `filterAvailableShippingOptions` call it; the rest ignore it. No customer
  strategy spreads its options bag into a third-party call, so an unread property is inert.
  A hardcoded list of supporting wallets would have to be kept in step with the SDK, and
  forgetting to update it would fail silently.
- **Always passed, so registration timing never matters.** The function handed to the
  strategy re-reads the global at the moment the shopper changes an address inside a sheet.
  A storefront can register its callback at any point — including after an `await fetch()`
  for its own configuration — and filtering still works. The alternative considered was
  checking for the callback when the button mounts and omitting the option otherwise; that
  would have kept the diff inside this package, but a callback registered a moment too late
  would then silently never filter, which is the worst failure mode for a merchant to debug.
- **Inert for stores that do not use it.** With no callback registered the function returns
  the shipping options unchanged, so the rate list is identical to today.
- **The callback's result is validated.** Anything other than an array — `undefined`, `null`,
  an object, a string — plus a throw or a rejection, falls back to the unfiltered options and
  logs. This is not defensive padding: the strategies call `.map()` on the result **outside**
  their own try/catch (`google-pay-gateway.ts:345`,
  `stripe-link-v2-button-strategy.ts:532`), so a callback that forgets to `return` would
  hard-break the wallet sheet rather than degrade. Validating at this boundary keeps the
  contract of the global independent of each strategy's internals.
- **Explicit options still win.** The resolved filter is spread *before*
  `additionalInitializationOptions`, so a wallet wrapper that passes its own
  `filterAvailableShippingOptions` overrides the storefront one.

A storefront-side change passes the same callback to the product and cart page wallet
buttons, so one registration covers all three surfaces.

## Rollout/Rollback

No experiment or flag. The change is inert unless a storefront defines the global, and the
absent-callback path returns the list untouched.

Rollback is a straight revert. No migrations, no persisted state, no configuration.

## Testing

`npx nx run-many --target=test` across the affected packages — 149 passing:

```
checkout-button-integration   26 passed
stripe-integration            49 passed
apple-pay-integration          3 passed
google-pay-integration        65 passed
amazon-pay-v2-integration      6 passed
```

`core` also passes in full (229 suites, 1566 passed / 2 skipped).

Lint is clean — 0 errors (one pre-existing `no-explicit-any` warning in
`StripeV3PaymentMethod.tsx`, untouched by this PR).

`npm run typecheck` reports **167 errors, byte-identical to the set on a clean master**, and
none in any file this PR touches. I measured the baseline by stashing this branch and
re-running, then diffing the sorted error sets.

New unit coverage in `resolveWalletShippingOptionsFilter.test.ts`:

- the options are returned unchanged when the global is absent, when the namespace has no
  callback, and when the callback is not a function
- the filtered options are returned when a callback is registered, including an async one
- the normalized context is passed to the callback
  (`googlepaystripeupe` → `methodId: 'googlepay'`)
- the callback is resolved at call time, not when the filter is created — a callback
  registered *after* the filter is built is still honoured
- every `googlepay*` variant normalizes to `methodId: 'googlepay'`

- the options are returned unchanged when the global is absent, when the namespace has no
  callback, and when the callback is not a function
- the filtered options are returned when a callback is registered, including an async one
- the normalized context is passed to the callback, and every `googlepay*` variant maps to
  `methodId: 'googlepay'`
- **the callback is resolved at call time, not when the filter is created** — a callback
  registered *after* the filter was built is still honoured, which is the guarantee that
  makes late registration safe

In `CheckoutButton.test.tsx`: the filter reaches the options for supporting wallets, it is
also passed to a wallet that ignores it (`amazonpay`, asserted against the exact options
object), and an explicitly provided filter takes precedence.

Five existing tests assert the exact `initializeCustomer` options and now receive the extra
property, so each gained one expectation: `StripeLinkV2Button`, `BigCommercePaymentsButton`,
`BigcommercePaymentsPayLaterButton`, `PayPalCommerceButton` and `PayPalCommerceCreditButton`.
The Apple Pay and Google Pay button tests use `expect.objectContaining`, so they were
unaffected.

Two unrelated failures showed up while verifying and are **pre-existing, not from this PR**:
`test-framework` fails on clean master too, and
`core/.../ManageInstrumentsModal.test.tsx` is a known flaky test — it appears in CircleCI's
flaky-test report for this repo with the same React-modal/`userEvent` timing signature, and
it passes in isolation on this branch.

### Verified on a real store

Filtering applied inside the wallet payment sheets on the checkout customer step, with a
callback registered on `window.BigCommerce.walletButtons.filterAvailableShippingOptions`:

#### Apple Pay

https://github.com/user-attachments/assets/f2aa059d-9885-4f40-8c7e-ef22a3f4d310

#### GPay

https://github.com/user-attachments/assets/3fa28cc8-e875-4152-84ad-cf3bf22d2b4a

#### Apple Pay

https://github.com/user-attachments/assets/969b9c35-c29b-4f04-8d24-097a4bd6641a

#### GPay

https://github.com/user-attachments/assets/e366aab7-65c2-4c2b-9b4c-d800136a9a4d

#### Apple Pay

https://github.com/user-attachments/assets/2e48727e-06d7-422e-a024-186870fdfa04

#### GPay

https://github.com/user-attachments/assets/c5fd5de0-6e32-4201-8da0-3b51fec6ce12

The product and cart surfaces are covered by the same storefront-side change and verified
separately; this PR covers the checkout customer step only.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Touches checkout customer initialization for all wallet buttons; incorrect filtering or validation could affect shipping options in payment sheets, though the no-callback path is unchanged and errors degrade to the full list.
> 
> **Overview**
> **Checkout wallet buttons** now pass `filterAvailableShippingOptions` into `initializeCustomer`, so storefronts can narrow shipping rates shown inside Apple Pay, Google Pay, Stripe Link, and other wallet sheets without forking checkout-js.
> 
> A new `resolveWalletShippingOptionsFilter` helper looks up `window.BigCommerce.walletButtons.filterAvailableShippingOptions` **when the shopper changes address** (not at button mount), normalizes Google Pay gateway IDs to `methodId: 'googlepay'`, supports async callbacks, and **falls back to the original list** if the callback is missing, invalid, throws, or returns a non-array. `CheckoutButton` applies this for every wallet method; `additionalInitializationOptions` can still override with an explicit filter.
> 
> Tests cover the resolver behavior and updated expectations on `CheckoutButton` and several wallet button integration tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2662beb6072c851f0bf2357df1c008fd24daaa1c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/checkout-button-integration/src/resolveWalletShippingOptionsFilter.ts
```diff
@@ -0,0 +1,57 @@
...
+): WalletShippingOptionsFilterContext => ({
+    methodId: methodId.startsWith(googlePayMethodIdPrefix) ? googlePayMethodIdPrefix : methodId,
+    gatewayId: methodId,
+    page: 'checkout',
+});
+
+export const resolveWalletShippingOptionsFilter =
+    (methodId: string) =>
+    async (shippingOptions: ShippingOption[]): Promise<ShippingOption[]> => {
+        const filter = window.BigCommerce?.walletButtons?.filterAvailableShippingOptions;
```
bc-donfran, line 38: Do we need the window.Bigcommerce here. If implemented wrong it lets any javascript be rendered on the checkout page which may lead to magecart attacks.

packages/checkout-button-integration/src/resolveWalletShippingOptionsFilter.ts
```diff
@@ -0,0 +1,57 @@
...
+
+        if (typeof filter !== 'function') {
+            return shippingOptions;
+        }
+
+        try {
+            const filteredOptions = await filter(
```
bc-donfran, line 45: This line can lead to dos if the await doesn't resolve. It might be worth adding a timeout just in case

packages/checkout-button-integration/src/resolveWalletShippingOptionsFilter.ts
```diff
@@ -0,0 +1,57 @@
...
+                shippingOptions,
+                getWalletShippingOptionsFilterContext(methodId),
+            );
+
+            return Array.isArray(filteredOptions) ? filteredOptions : shippingOptions;
```
bc-donfran, line 50: This allows for price spoofing. A merchant is able to display a $0 shipping for something that isn't 0 which is sort of price deception.

bc-donfran: This code also allows forced selection where someone can return a shipping option value forcing the shopper to only select a shipping option which could be more pricey

packages/checkout-button-integration/src/resolveWalletShippingOptionsFilter.ts
PavlenkoM: make sense

packages/checkout-button-integration/src/resolveWalletShippingOptionsFilter.ts
PavlenkoM: There are several ways to customize checkout — using custom checkout options, adding code via Script Manager, or editing theme files in CP — and these already give merchants the ability to display $0 shipping for something that isn't actually $0, which could be seen as a form of price deception. They also give merchants a lot more power to change anything on the checkout page and display whatever they'd like.
The same applies to "forced selection" — merchants can already do this for shipping options in the shipping step today, and that's considered acceptable. So I'd gently push back on the idea that doing the same thing for other placements would suddenly become a problem.
I think this concern may not be specific to these particular changes — there are already plenty of ways to customize checkout beyond our default CP options. This change is really just a generic solution to make the existing logic consistent everywhere, giving merchants the same customization for wallet buttons that they already have for the shipping step.
That said, we're happy to close this discussion, since the PR isn't relevant anymore — we'll be moving these changes to the shipping team, who will add the options on the BE side. That should give us one more entry point for shipping customization :)

